PR #14, opened 2017-08-20
added date spine

jthandy:
handles the most basic case for https://github.com/fishtown-analytics/dbt-utils/issues/11


jthandy:
@drewbanin significantly updated. this works for me in a local repo. take a look, love to hear if you have any thoughts!

macros/date_spine.sql
```diff
@@ -0,0 +1,46 @@
...
+
+    select (
+        {{
+            dateadd(
+                datepart,
+                "row_number() over () - 1",
+                start_date
+            )
+        }}
+    ) as date_{{datepart}}
```
drewbanin: I think this is clever, but I'm not sure how I feel about returning different fields based on the supplied arguments. On the one hand, the resulting field names will be explicit. On the other, this could break downstream models if the datepart is changed from `day` to `week`, for instance. Maybe that's a good thing?

macros/cross_db_utils/dateadd.sql
```diff
@@ -0,0 +1,32 @@
...
+        {{ interval }},
+        "{{ datepart }}"
+        )
+
+{% endmacro %}
+
+
+{% macro postgres__dateadd(datepart, interval, from_date_or_timestamp) %}
+
+    from_date_or_timestamp + interval '{{ interval }} {{ datepart }}'
```
drewbanin: `from_date_or_timestamp` needs to be wrapped in curly brackets :)

PR #17, opened 2017-08-24
added get url parameter syntax plus cross-database version of split_part

jthandy:
resolves https://github.com/fishtown-analytics/dbt-utils/issues/15


macros/get_url_parameter.sql
```diff
@@ -0,0 +1,9 @@
+{% macro get_url_parameter(field, url_parameter) %}
+
+{% set formatted_url_parameter = "'" + url_parameter + "='" %}
```
drewbanin: we should make an `adapter.quote_string` function but this is cool for now

PR #22, opened 2017-10-04
fix for star macro with no `from` arg

drewbanin:
this can happen either in the dbt parsing step, or intentionally. If no `from` is provided, then `*` is used.


drewbanin:
this isn't an issue anymore due to changes in 090. Ref will never return null, even in parsing

PR #23, opened 2017-10-12
compatibility with 0.9.0

cmcarthur:
Changes:

- `ref()` now returns a `dbt.utils.Relation`, which has `name` and `schema` as properties

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,9 @@
 ----
 
 # dbt-utils
+
+Current version: 0.1.0
+
 This package contains macros that can be (re)used across dbt projects.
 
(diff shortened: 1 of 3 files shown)


cmcarthur:
@jthandy merging this to master will break compatibility with 0.8.x. is that going to break any of your projects?

we should make sure there's a tag for the prior version, and a new one when this gets merged.

jthandy:
I always @ tag all deps, so we just need to cut a new release.

cmcarthur:
kk, i'll release this as 0.1.0.

PR #27, opened 2017-10-20
nullcheck macro

eogilvy12:
null check string and text column types from input list of column names


macros/sql/nullcheck
```diff
@@ -0,0 +1,17 @@
+{% macro nullcheck(cols) %}
+{%- for col in cols %}
+
+    {% if col.is_string() -%}
```
drewbanin: this is awesome

drewbanin:
thanks @eogilvy12 -- let's write some docs for this soon!

PR #28, opened 2017-10-20
nullcheck_table

eogilvy12:
using nullcheck macro to null check string and text data types from base table


drewbanin:
this file needs a `.sql` file extension

PR #31, opened 2017-11-03
Create lower_alpha.sql

elyse-weiss:
I will add docs.


drewbanin:
Closing this -- it's a little too specific to be useful in this repo, but we could do lots like this in something like `dbt-string-utils` in the future.

PR #32, opened 2017-11-06
Test for columns with same value in all rows or no value at all

nave91:
test_atleast_one:

Check if there is atleast one value in a column. 

test_not_constant:

Check if column has same value across all rows.


nave91:
@drewbanin opened it here as requested

https://github.com/fishtown-analytics/dbt/pull/578

drewbanin:
Thanks @nave91! These are great! Do you mind updating the [README.md](https://github.com/fishtown-analytics/dbt-utils#schema-tests) file accordingly? Just a sentence about what the tests do, and how to call them would be perfect.

We recently went through the process of rewriting dbt's schema tests to not use CTEs where possible. This makes tests run significantly faster on Postgres. That's not critical, but if you're interested, it would be cool to rewrite these sql statements similarly. You can check out an example [here](https://github.com/fishtown-analytics/dbt/blob/development/dbt/include/global_project/macros/schema_tests/relationships.sql)

drewbanin:
Thanks @nave91! These are great! Do you mind updating the [README.md](https://github.com/fishtown-analytics/dbt-utils#schema-tests) file accordingly? Just a sentence about what the tests do, and how to call them would be perfect.

We recently went through the process of rewriting dbt's schema tests to not use CTEs where possible. This makes tests run significantly faster on Postgres. That's not critical, but if you're interested, it would be cool to rewrite these sql statements similarly. You can check out an example [here](https://github.com/fishtown-analytics/dbt/blob/development/dbt/include/global_project/macros/schema_tests/relationships.sql)

nave91:
@drewbanin just added the changes you requested. Let me know if SQL is tricky to read, I'm happy to change it.

macros/schema_tests/atleast_one.sql
```diff
@@ -0,0 +1,19 @@
...
+    from (
+       select
+        count({{ arg }}) count_of_rows
+
+        from {{ model }}
+    ) rows_with_values
+
+    where count_of_rows = 0 ) validation_errors
+
+{% endmacro %}
```
drewbanin: @nave91 I think we can rewrite this as follows:

```sql
select
    count({{ arg }})
from {{ model }}
having count({{ arg }}) = 0
```

I _think_ these are equivalent, but lmk if you disagree!

macros/schema_tests/not_constant.sql
```diff
@@ -0,0 +1,21 @@
...
+          from {{ model }}
+
+          where {{ arg }} is not null
+
+         ) not_null_rows
+  ) discount_rows_count
+
+  where count_of_distinct_rows = 1
+
+{% endmacro %}
```
drewbanin: I think the approach listed above will work here too! This macro is interesting in that a column full of `null` values will "pass" this test. Is that the desired behavior? I suppose it makes sense to use it in conjunction with `atleast_one` above

nave91: Yes that was intended, I didnt want to generate false positives when we already have a separate test for it.

nave91:
This should be good now. I tried using `having` earlier but was running into some error, I think I didnt count validation errors and was running into some error. Since it was already working, I didnt bother on digging deep but using having makes more sense in this context.

drewbanin:
@nave91 this looks great! This is super nitpicky, but do you mind also changing `atleast` to `at_least` in the test name and in the docs? It's not a huge deal, but we can't really change test names once they're merged + used in the wild, so it makes sense to do it now.

Pending that, this looks good to me! Thanks for your contribution!

nave91:
done!

drewbanin:
:shipit: approved!

PR #35, opened 2018-01-04
Add cardinality equality schema test

dwallace0723:
This custom schema test macro checks to see if the cardinality of one field in a model exactly matches the cardinality of a chosen field in a different model.

### DIFF
--- a/macros/schema_tests/cardinality_equality.sql
+++ b/macros/schema_tests/cardinality_equality.sql
@@ -2,18 +2,18 @@
 
 with table_a as (
 select
-  count(1) as num_rows,
-  {{ from }}
+  {{ from }},
+  count(1) as num_rows
 from {{ model }}
-group by 2
+group by 1
 ),
 
 table_b as (
 select
-  count(1) as num_rows,
-  {{ field }}
+  { field }},
+  count(1) as num_rows
 from {{ to }}
-group by 2
+group by 1
 )
 
 select count(1)
--- a/macros/schema_tests/cardinality_equality.sql
+++ b/macros/schema_tests/cardinality_equality.sql
@@ -14,15 +14,31 @@ select
   count(1) as num_rows
 from {{ to }}
 group by 1
+),
+
+except_a as (
+  select *
+  from table_a
+  except
+  select *
+  from table_b
+),
+
+except_b as (
+  select *
+  from table_b
+  except
+  select *
+  from table_a
 )
 
 select count(1)
 from (
-select *
-from table_a
-except
-select *
-from table_b
+  select *
+  from except_a
+  union all
+  select *
+  from except_b
 )
 
 {% endmacro %}
--- a/macros/schema_tests/cardinality_equality.sql
+++ b/macros/schema_tests/cardinality_equality.sql
@@ -3,15 +3,15 @@
 with table_a as (
 select
   {{ from }},
-  count(1) as num_rows
+  count(*) as num_rows
 from {{ model }}
 group by 1
 ),
 
 table_b as (
 select
-  { field }},
-  count(1) as num_rows
+  {{ field }},
+  count(*) as num_rows
 from {{ to }}
 group by 1
 ),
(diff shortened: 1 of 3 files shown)


macros/schema_tests/cardinality_equality.sql
```diff
@@ -0,0 +1,28 @@
+{% macro test_cardinality_equality(model, from, to, field) %}
```
drewbanin: @dwallace0723 the `from`, `to`, and `field` args can be named arbitrarily -- we used these for the referential integrity test, but you can pick other names if you'd like!

macros/schema_tests/cardinality_equality.sql
```diff
@@ -0,0 +1,28 @@
...
+
+with table_a as (
+select
+  count(1) as num_rows,
```
drewbanin: Can you put the dimensions before the aggregates here (eg. `group by 1`, also in the `table_b` cte)

Any reason to use `count(1)` instead of `count(*)` here?

macros/schema_tests/cardinality_equality.sql
```diff
@@ -0,0 +1,28 @@
...
+  {{ field }}
+from {{ to }}
+group by 2
+)
+
+select count(1)
+from (
+select *
+from table_a
+except
```
drewbanin: This is groovy! I think the `except` set function is asymmetrical -- `table_a - table_b` is different than `table_b - table_a`. As a result, this test would "pass" if `table_b` contained records which were not present in `table_a`, ie. if `table_b` is a superset of `table_a`.

There may be a simpler way to do this, but I've done something like this before:

```sql
-- table_a and table_b are equal if this union results in 0 rows
(
  select * from table_a
  except
  select * from table_b
)

union all

(
  select * from table_b
  except
  select * from table_a
)
```

This is cool (for some value of cool) b/c it's essentially the definition of set equality:
A = B if A is a subset of B, and B is a subset of A

dwallace0723: Good catch! It does feel a little bit strange that there isn't a simpler way to do this than a `union all`

macros/schema_tests/cardinality_equality.sql
dwallace0723: I'll probably just leave them as is. The current naming makes logical sense to me. LMK if you think there are more suitable arg names for this specific test.

macros/schema_tests/cardinality_equality.sql
dwallace0723: re: `count(1)` instead of `count(*)`, this is strictly habitual to keep myself out of the practice of selecting *. I'm relatively positive it makes no performance difference whatsoever. Will gladly change to `count(*)` if it better fits the dbt style guidelines.

dwallace0723:
back to you @drewbanin

macros/schema_tests/cardinality_equality.sql
```diff
@@ -0,0 +1,44 @@
...
+select
+  {{ from }},
+  count(1) as num_rows
+from {{ model }}
+group by 1
+),
+
+table_b as (
+select
+  { field }},
```
drewbanin: There's a typo here

macros/schema_tests/cardinality_equality.sql
```diff
@@ -0,0 +1,44 @@
...
+)
+
+select count(1)
+from (
+  select *
+  from except_a
+  union all
+  select *
+  from except_b
+)
```
drewbanin: If you alias this subquery, then this schema test will also work on Postgres. At present, it errors out with:
```
subquery in FROM must have an alias
  LINE 36: from (
                ^
  HINT:  For example, FROM (SELECT ...) [AS] foo.
```

So just change it to
```
) as sbq
```
and you should be good! Or, you could make another CTE for the subquery

dwallace0723: Ah, good catch. I just created an additional CTE for the subquery.

dwallace0723:
Cool, changes made @drewbanin 👍

drewbanin:
Groovy! Thanks for contributing @dwallace0723 -- merging this now :)

PR #36, opened 2018-01-07
get_tables_by_prefix, ideal for union_tables jobs

jdub:
Solution suggested by @drewbanin, enhanced with "exclude" parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,34 @@ model_name:
 
 ---
 ### SQL helpers
+#### get_column_values ([source](macros/sql/get_column_values.sql))
+This macro returns the unique values for a column in a given table.
+
+Usage:
+```
+-- Returns a list of the top 50 states in the `users` table
+{% set states = fromjson(get_column_values(table=ref('users'), column='state', max_records=50)) %}
+
+{% for state in states %}
+    ...
+{% endfor %}
+
+...
+```
+
+#### get_tables_by_prefix ([source](macros/sql/get_tables_by_prefix.sql))
+This macro returns a list of tables that match a given prefix, with an optional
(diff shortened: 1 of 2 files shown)


jdub:
This works _great_ without the `adapter_macro` wrapper, but when I add it, I get this error:
```
Encountered an error:
not enough values to unpack (expected 2, got 1)
```

drewbanin:
@jdub this is great! I believe the `not enough values to unpack (expected 2, got 1)` error you're seeing was fixed in `0.9.1` (released 1/5). You could try upgrading to see if that fixes this.

Separately, I noticed a [bug](https://github.com/fishtown-analytics/dbt/issues/635) in how dbt implements `adapter_macro` which I think will prevent this particular macro from working correctly (for now).

I think the best solution in the short term would be to nix the `adapter_macro` approach, and instead just place the `get_tables_by_prefix` macro somewhere like `macros/sql/get_tables_by_prefix.sql`.

If it makes sense to extend this macro to other adapters in the future, we can definitely do that without breaking backwards compatibility.

jdub:
Not fixed in 0.9.1 sadly:

```
$ dbt --version
installed version: 0.9.1
   latest version: 0.9.1
Up to date!
$ dbt run --models awsbilling








  []

Encountered an error:
not enough values to unpack (expected 2, got 1)
```

Thanks -- I'll update with your suggested change.

drewbanin:
@jdub let me know when this is ready for another look. Can you also add documentation to the `README.md` file?

jdub:
@drewbanin README updated, with a bit of alphabetical re-ordering for good measure. This is good to go.

drewbanin:
@jdub groovy - this LGTM! Merging

PR #38, opened 2018-01-09
added new generate_series macro

jthandy:
addresses https://github.com/fishtown-analytics/dbt-utils/issues/33

have tested this on redshift, bigquery, postgres, and snowflake. the sql is pretty generic so fortunately no cross-db rewrites to worry about.


drewbanin:
@jthandy nice! This needs some docs (`README.md`) and then it's good to go!

PR #41, opened 2018-01-10
Qualify the get url parameter util macro

dwallace0723:
This fixes the `get_url_parameter` macro so that it works with 091


drewbanin:
Thanks @dwallace0723! I gave this an optical integration test and it looks good to me :)

I can give it a spin later, or just confirm that you've tested w/ 091 here and I'll gladly merge

dwallace0723:
@drewbanin tested with 091 and it works 👍

PR #42, opened 2018-01-10
Add pivot macro

mplovepop:


macros/sql/pivot.sql
```diff
@@ -0,0 +1,61 @@
...
+    +======+=======+
+    | S    | red   |
+    | S    | blue  |
+    | S    | red   |
+    | M    | red   |
+
+    select
+      size,
+      {{ pivot('size', ['red', 'blue']) }}
+    from <table>
```
mplovepop: Should add group by size

macros/sql/pivot.sql
```diff
@@ -0,0 +1,62 @@
...
+    | size | color |
+    |------+-------|
+    | S    | red   |
+    | S    | blue  |
+    | S    | red   |
+    | M    | red   |
+
+    select
+      size,
+      {{ pivot('size', ['red', 'blue']) }}
```
drewbanin: we don't do a great job of documenting this, but macros in this package need to be namespaced to `dbt_utils`, ie. 

```
{{ dbt_utils.pivot(....) }}
```

macros/sql/pivot.sql
```diff
@@ -0,0 +1,62 @@
...
+  {% for v in values %}
+    {{ op }}(
+      case
+      when {{ column }} {{ cmp }} '{{ v }}'
+        then {{ then_value }}
+      else {{ else_value }}
+      end
+    )
+    {% if alias %}
+      as {{ prefix + v + postfix }}
```
drewbanin: If you use the jinja concatenation operator (`~`) instead of the addition operator (`+`), then this will work when `v` is numeric too. Otherwise, Jinja throws:

```
2018-01-10 19:25:18,553: Compilation Error in macro pivot (pivot.sql)
  must be str, not int
```

macros/sql/pivot.sql
```diff
@@ -0,0 +1,62 @@
...
+    else_value: Value to use if comparison fails, default is 0
+#}
+
+{% macro pivot(column,
+               values,
+               alias=True,
+               op='sum',
+               cmp='=',
+               prefix='',
+               postfix='',
```
drewbanin: I think `suffix` might be more appropriate than `postfix` here, but I don't feel too strongly about it

macros/sql/pivot.sql
```diff
@@ -0,0 +1,62 @@
...
+    prefix: Column alias prefix, default is blank
+    postfix: Column alias postfix, default is blank
+    then_value: Value to use if comparison succeeds, default is 1
+    else_value: Value to use if comparison fails, default is 0
+#}
+
+{% macro pivot(column,
+               values,
+               alias=True,
+               op='sum',
```
drewbanin: I think `agg` might be more clear than `op` here (indicating that the group-by is required), but I'm ok with either

drewbanin:
@mplovepop have you seen the [get_column_values](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/get_column_values.sql) macro in here? It will issue a query to the db to find all of the distinct values in a column, then returns those values as a list. I was playing around with that + this, and it's really an amazing experience.

Check this out:

```sql
{% set values = dbt_utils.get_column_values('public.test', 'color') %}

select
  size,
  {{ dbt_utils.pivot('color', values, prefix='color_') }}
from public.test
group by 1
```

screen shot 2018-01-10 at 7 50 14 pm


I didn't have to tell this macro that my colors were red/green/blue!

macros/sql/pivot.sql
drewbanin: It's also a good idea to quote `{{ prefix ~ v ~ suffix }}` here. Otherwise, if there's no prefix and `v` is not a string, this will break.

To quote in a cross-platform way, you can do:

```
      as {{ adapter.quote(prefix ~ v ~ suffix) }}
```

So this will work for Redshift, Snowflake, and BigQuery

drewbanin:
@mplovepop LGTM -- Thanks for contributing :)

Merging!

PR #43, opened 2018-01-11
added recency schema test

jthandy:
this is super-cool--uses two cross-db-utils that already existed :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,17 @@ model_name:
 
 ```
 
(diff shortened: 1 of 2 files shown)


jthandy:
@drewbanin I previously hadn't merged this because we weren't totally sure about namespacing stuff. I think we've decided in the intervening weeks that macros called within packages should be called with their own package name. that's what I did here and it worked when calling this from another package.

PR #45, opened 2018-01-11
cleans up date spine call signature by building table using generate_…

jthandy:
…series macro

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,6 @@ This macro returns the sql required to build a date spine.
 Usage:
 ```
 {{ dbt_utils.date_spine(
-    table=ref('organizations'),
     datepart="minute",
     start_date="to_date('01/01/2016', 'mm/dd/yyyy')",
     end_date="dateadd(week, 1, current_date)"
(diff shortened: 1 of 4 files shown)


jthandy:
the point of this was to remove the need to pass in a random table and instead use the generate_series macro to perform this function. in the process, i had to calculate the number of intervals required for generate series, which required that i build a datediff version for cross_db_utils. note that in that code there is no implementation for postgres; after reading about that for a bit it really does not seem that there is a standard syntax to translate from redshift-sql to postgres-sql unless the postgres user creates a UDF. that might be the way to go but for now I'm not tackling that; I just raised an exception in that case for postgres users.

drewbanin:
@jthandy I think dropping postgres support is fine. For posterity, we have a UDF for datediff on postgres [here](https://github.com/fishtown-analytics/snowplow/blob/fea3800408781f5a869ca946448ace405333d39d/README.md#database-support)

jthandy:
yeah i saw that. the real question i have about that is how do we make that as seamless as possible from a UI perspective. we'd need a postgres package that included a macro to install that, then somehow we'd have to figure out a way to throw a meaningful error if it weren't installed so that the user would know how to address it...? the latter is what I don't know how to do at the moment.

anyway, figured we could return to it later.

jthandy:
agree. the more macros i've written recently the more i've thought about:

- doing more advanced calculations in the jinja layer, and 
- caching in the jinja layer to avoid repeated db calls.

these are not new topics, but it's hard to write this code without them becoming apparent.

merging.

PR #46, opened 2018-01-11
Have get url paramater macro return null when blank

dwallace0723:
This alters the `get_url_paramater` macro to return NULL instead of an empty string when no relevant parameters are found in the URL being parsed.


dwallace0723:
cc @drewbanin relevant to our convo yesterday

PR #53, opened 2018-02-01
Feature/surrogate key

jthandy:
resolves https://github.com/fishtown-analytics/dbt-utils/issues/51

### DIFF
--- a/macros/cross_db_utils/concat.sql
+++ b/macros/cross_db_utils/concat.sql
@@ -1,0 +1,27 @@
+{% macro concat(fields) %}
+  {{ adapter_macro('dbt_utils.concat', fields) }}
+{% endmacro %}
+
+
+{% macro default__concat(fields) -%}
+    concat({{ fields|join(', ') }})
+{%- endmacro %}
+
+
+{% macro alternative_concat(fields) %}
+    {{ fields|join(' || ') }}
+{% endmacro %}
+
+
+{% macro redshift__concat(fields) %}
+
+    {{dbt_utils.alternative_concat(fields)}}
+
+{% endmacro %}
+
+
+{% macro snowflake__dateadd(fields) %}
+
+    {{dbt_utils.alternative_concat(fields)}}
+
+{% endmacro %}
--- a/macros/cross_db_utils/concat.sql
+++ b/macros/cross_db_utils/concat.sql
@@ -14,14 +14,10 @@
 
 
 {% macro redshift__concat(fields) %}
-
     {{dbt_utils.alternative_concat(fields)}}
-
 {% endmacro %}
 
 
 {% macro snowflake__dateadd(fields) %}
-
     {{dbt_utils.alternative_concat(fields)}}
-
 {% endmacro %}
--- a/macros/cross_db_utils/concat.sql
+++ b/macros/cross_db_utils/concat.sql
@@ -18,6 +18,6 @@
 {% endmacro %}
 
 
-{% macro snowflake__dateadd(fields) %}
+{% macro snowflake__concat(fields) %}
     {{dbt_utils.alternative_concat(fields)}}
 {% endmacro %}
--- a/macros/sql/surrogate_key.sql
+++ b/macros/sql/surrogate_key.sql
@@ -4,7 +4,9 @@ md5(concat(
 
     {%- for field in fields %}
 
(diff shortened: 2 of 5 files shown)


jthandy:
@drewbanin every time I attempt to implement some particular piece of functionality it always ends up being trickier here than I realized. Type conversion was the tricky thing in this particular instance. while all four databases we care about have the same syntax for `cast(field as type)`, snowflake is more argumentative about what types it will allow you to cast as varchar. weirdly, you cannot cast a boolean to varchar using any mechanism I was able to identify.

In this process I went down the rabbit hole of learning about `safe_cast` (bigquery) and `try_cast` (snowflake) and figured they'd be good to implement as defaults when casting. redshift and postgres do not have these functions, so we default to just using standard casting, which just means that if the user supplies values that cause an error, they will receive an error.

Overall, from what testing I'm able to do, this works as intended.

macros/sql/surrogate_key.sql
```diff
@@ -0,0 +1,16 @@
+{% macro surrogate_key(fields) -%}
+
+md5(concat(
+
+    {%- for field in fields %}
+
+        coalesce(
+            {{dbt_utils.safe_cast(field, dbt_utils.type_string())}}
```
drewbanin: I was initially opposed to using `safe_cast` here because it's possible that two distinct lists of `fields` values would both evaluate to `md5('')` if none of the arguments are cast-able. It looks to me though like pretty much everything can be casted as a string, and `null` is only returned for more restrictive types like `int64` on bigquery.

macros/sql/surrogate_key.sql
```diff
@@ -0,0 +1,16 @@
+{% macro surrogate_key(fields) -%}
```
drewbanin: You can leave out this `fields` argument and then use `varargs` instead of `fields` down below. The `varargs` magic variable becomes a list of arguments passed to this macro, so you can drop the `fields=[...]` boilerplate and write

```
{{ dbt_utils.surrogate_key('field_a', 'field_b') }}
```

Kind of personal preference here, so totally fair if you prefer the list syntax. I like that the varargs syntax feels similar to `coalesce`.

macros/sql/surrogate_key.sql
```diff
@@ -0,0 +1,16 @@
...
+
+md5(concat(
+
+    {%- for field in fields %}
+
+        coalesce(
+            {{dbt_utils.safe_cast(field, dbt_utils.type_string())}}
+            , '')
```
drewbanin: I think we should concatenate each element in the list with `-`. Then, the number of items in the list is significant.

For instance, these two things should evaluate to different values:
```
{{ dbt_utils.surrogate_key(fields=['null']) }}

{{ dbt_utils.surrogate_key(fields=['null', 'null']) }}
```

In practice, this might happen with weirdness around utm parameters. Imagine:

| row | utm_medium  | utm_source |
| ---- | ------------- | ------------- |
| 1 | null | email  |
| 2 | email  | null  |

If we used the `surrogate_key` macro as it's currently implemented, both of these would evaluate to 

```
{{ dbt_utils.surrogate_key(fields=['utm_medium', 'utm_source']) }}

-->

md5('email')

```

whereas with dashes between each field, it would be:

```
-- Row 1
md5('-email')

-- Row 2
md5('email-')
```

Even the dash is not infallible here, but I think it's better than nothing!

macros/sql/surrogate_key.sql
jthandy: i agree with you. I didn't know about `varargs` but just updated. very easy.

jthandy:
@drewbanin should be good to go here. Had to address concatenation differences between database platforms.

macros/cross_db_utils/concat.sql
```diff
@@ -0,0 +1,23 @@
...
+    {{ fields|join(' || ') }}
+{% endmacro %}
+
+
+{% macro redshift__concat(fields) %}
+    {{dbt_utils.alternative_concat(fields)}}
+{% endmacro %}
+
+
+{% macro snowflake__dateadd(fields) %}
```
drewbanin: s/dateadd/concat

jthandy: blerg. fixed.

PR #54, opened 2018-02-11
Feature/last day

jthandy:
@drewbanin couple of things:

1. it turns out that `last_day` isn't widely supported across adapters (postgres and bigquery don't have it at all, and it's inconsistently implemented across redshift and snowflake).

2. I only particularly care about this for months and quarters, and so have only tested it for those. my guess is that there will be inconsistencies in how it works across adapters for weeks and so i just didn't want to get into that because (at least for my use case) this is not valuable. I left a comment so that this is documented in the function.

### DIFF
--- a/macros/cross_db_utils/last_day.sql
+++ b/macros/cross_db_utils/last_day.sql
@@ -8,30 +8,27 @@ testing is required to validate that it will work on other dateparts.
 {% endmacro %}
 
 
-{% macro default__last_day(date, datepart) -%}
+{%- macro default_last_day(date, datepart) -%}
     cast(
         {{dbt_utils.dateadd('day', '-1',
         dbt_utils.dateadd(datepart, '1', dbt_utils.date_trunc(datepart, date))
         )}}
         as date)
+{%- endmacro -%}
+
+
+{% macro default__last_day(date, datepart) -%}
+    {{dbt_utils.default_last_day(date, datepart)}}
 {%- endmacro %}
 
 
 {% macro postgres__last_day(date, datepart) -%}
 
     {%- if datepart == 'quarter' -%}
-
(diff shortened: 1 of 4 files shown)


jthandy:
also -- during testing, i realized that our bigquery implementation of `dateadd` was not (and somehow seems to have never been!?) working. had to fix in the process.

I've tested this across all 4 adapters.

macros/cross_db_utils/last_day.sql
```diff
@@ -0,0 +1,37 @@
...
+        {{dbt_utils.dateadd('day', '-1',
+        dbt_utils.dateadd(datepart, '1', dbt_utils.date_trunc(datepart, date))
+        )}}
+        as date)
+{%- endmacro %}
+
+
+{% macro postgres__last_day(date, datepart) -%}
+
+    {%- if datepart == 'quarter' -%}
```
drewbanin: just curious -- what's the problem with using `quarter` on Postgres? Does that datepart not exist?

macros/cross_db_utils/last_day.sql
```diff
@@ -0,0 +1,37 @@
...
+
+    {{ exceptions.raise_compiler_error(
+        "dbt_utils.last_day is not supported for datepart 'quarter' on this adapter") }}
+
+    {%- else -%}
+
+    cast(
```
drewbanin: Is this code the same as the code in the default `last_day` macro? If so: we should probably just call that macro here instead of duplicating the logic, right?

jthandy: yep! that's exactly it.

macros/cross_db_utils/last_day.sql
jthandy: yep! that's exactly it.

PR #55, opened 2018-03-01
Add macro that grants select permissions on all schemas in a database 

clausherther:
This is the **Snowflake** version for now. I'm not sure how to use the various adapters to make this work across platforms, but don't mind working on that if you could point me in the right direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,19 @@ Usage:
 ```
 {{ dbt_utils.haversine_distance(lat1=<float>,lon1=<float>,lat2=<float>,lon2=<float>) }}
 ```
+
+---
+### Permissions
+#### grant_select_on_schemas ([source](macros/permissions/grant_select_on_schemas.sql))
+This macro gives SELECT access to all tables in all schemata in the specified database to the specified role.
+This can useful, for example, to give the user/role used by dbt access to upstream source tables.
+In that use case, it's best to run this macro in a separate dbt project configured with security admin rights to the database in question.
+
+Usage:
+```
+{{ dbt_utils.grant_select_on_schemas(database_name='MyDB', role_name='MyRole') }}
+```
+
 ---
 ### Schema Tests
 #### equality ([source](macros/schema_tests/equality.sql))
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Usage:
 ### Permissions
 #### grant_select_on_schemas ([source](macros/permissions/grant_select_on_schemas.sql))
 This macro gives SELECT access to all tables in all schemata in the specified database to the specified role.
-This can useful, for example, to give the user/role used by dbt access to upstream source tables.
+This can be useful to give the user/role used by dbt access to upstream source tables.
 In that use case, it's best to run this macro in a separate dbt project configured with security admin rights to the database in question.
 
 Usage:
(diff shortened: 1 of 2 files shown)


drewbanin:
@clausherther this is awesome!!

In my experience, it's also useful to grant `usage` to users as well, but that might be a Redshift/Postgres-specific thing.

You can see an example of how we build cross-database compatible macros here: https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/cross_db_utils/datediff.sql

I I think if you want to do this for postgres/redshift/snowflake, you'd want some macros like:

```
{% macro grant_usage_on(schema, user) %}
  {{ adapter_macro('dbt_utils.grant_usage_on', schema, user) }}
{% endmacro %}

{% macro grant_select_on_all_tables_in_schema(schema, user) %}
  {{ adapter_macro('dbt_utils.grant_select_on_all_tables_in_schema', schema, user) }}
{% endmacro %}
```

Then below, you can set up a macro for each database type:
```
{% macro default__grant_usage_on(schema, user) %}
  grant usage on schema {{ schema }} to {{ user }};
{% endmacro %}

{% macro default__grant_select_on_all_tables_in_schema(schema, user) %}
  grant select on all tables in schema {{ schema }} to {{ user }};
{% endmacro %}

{% macro snowflake__grant_usage_on(schema, role) %}
  grant usage on schema {{ schema }} to role {{ role }};
{% endmacro %}

{% macro snowflake__grant_select_on_all_tables_in_schema(schema, role) %}
  grant select on all tables in schema {{ schema }} to {{ role }};
{% endmacro %}

{% macro bigquery__grant_usage_on(schema, role) %}
    {{ exceptions.raise_compiler_error("grants are not implemented for this adapter") }}
{% endmacro %}

{% macro bigquery__grant_select_on_all_tables_in_schema(schema, role) %}
    {{ exceptions.raise_compiler_error("grants are not implemented for this adapter") }}
{% endmacro %}
```

You'd probably need to do something similar for that query to grab schema names.

Obviously this stuff is hard to test without access to postgres/redshift/snowflake... let me know how much of this you're interested in doing! I'd ideally like to make this work with all of our adapters before merging, so, definitely happy to help out here :)

clausherther:
Ah, that's great, thanks @drewbanin !
I don't have access to BigQuery, but let me take a stab at the cross-platform setup for this for Redshift and Snowflake. Agreed on adding `usage` as well.

drewbanin:
@clausherther - just gave this another look... I didn't recognize that this runs a grant on all tables in _all_ schemas in the database. I'd feel much better about this if we broke it up in the following way:

1. macro to get all schemas in a database
2. macro to grant usage on a schema
3. macro to grant select on all tables in schema

If you want to wire these macros up to run grants on all tables in all schemas (as shown above), then that's certainly your prerogative! In the more general case though, that's probably something I'd advise against. I can totally imagine sensitive data being added to a warehouse at some point in the future. If this macro blindly grants access to every table, that could likely be undesirable.

You buy all that?

clausherther:
That makes sense @drewbanin. I had modeled this after my specific use case where I want my DBT_USER to have access to all schemas in our FIVETRAN database on Snowflake. 
Splitting this up makes this a lot more reusable for sure. Thx!

drewbanin:
@clausherther going to close this, but lmk if you want to revisit!

PR #57, opened 2018-03-02
add integration test project

drewbanin:
This PR adds integration tests for every macro defined in the dbt-utils project. These macros were tested with Postgres and Redshift. Where possible, tests were also run for BigQuery and Snowflake.

This project serves two purposes:
1. It will help build present and future confidence in the macros defined in this package
2. It will help avoid regressions in future releases of dbt

We should set up some sort of CI suite for this package. Ideally, it would run on pull requests.


jthandy:
holy shit...this is incredible. what do we need to set this up in sinter? just the ability to run a project that's not at the root directory? we don't have the ability to have public sinter accounts either I guess...in the same way you can look at the code climate score for any open source repo without having access to it.

would be an interesting conversation. it's really super-exciting to see this.

drewbanin:
@jthandy this package calls a couple of dbt context functions which are unsupported on BigQuery. This unfortunately causes runs against BQ to fail during _parsing_ which means that this project won't run at all on BQ.

I also found a bug last night that prevents statements from working on Snowflake if the the target schema does not exist: https://github.com/fishtown-analytics/dbt/issues/674. Same thing here -- this fails during parsing, so this project won't run at all on Snowflake :/

drewbanin:
I think step 1 is to get this running in Sinter even if it's still private. At the very least, we'll be able to know when we make changes in dbt that break dbt-utils. Plus, I think PR hooks _might_ work here, so it would be a nice additional check for contributions. We should definitely expose Sinter publicly in read-only mode for this repo when we can!

jthandy:
Totally agree--I'm interested in having the other conversation though.

PR #59, opened 2018-03-22
Adding index macro

nave91:


jthandy:
@nave91 whoah. this is very useful. thanks for contributing!

@drewbanin I think this should be in a postgres-specific package--agree?

nave91:
@drewbanin yes sir! should I close this one?

PR #60, opened 2018-03-26
star macro readme

jtcohen6:
## Motivation

Updated description + usage for star macro to clarify that one needs to write
```select {{dbt_utils.star(from=table,except=[]) from table```
The star macro only replaces the star, not the entire select statement. Previous wording implied that the macro would generate the entire table expression, cf. `dbt_utils.union_tables`.

## Approach

An alternative approach would be to change the underlying code to actually generate an entire select statement, including the words `select` and `from {{from}}` in the compiled output, but I think there are compelling use cases where we want a star macro coupled with some basic calculation / casting / field population in the same CTE, especially if we're using `dbt_utils.star` to exclude a field that requires slight modification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,11 +166,13 @@ Usage:
 ```
 
 #### star ([source](macros/sql/star.sql))
-This macro generates a `select` statement for each field that exists in the `from` relation. Fields listed in the `except` argument will be excluded from this list.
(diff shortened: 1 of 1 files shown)


jthandy:
Hah. I saw this conversation in dbt slack. I had never even considered that this aspect of usage might not be clear, but it does make sense. Thanks for doing this.

PR #62, opened 2018-04-17
pivot example bug

peterfine:
The example code doesn't match the example output, because it's looking for colours in the size column. This update fixes it as per the example.


drewbanin:
Thanks @peterfine :)

PR #63, opened 2018-04-30
fix snowflake bugs with 0.10.1 functionality

drewbanin:
In the pre-0.10.1 world, dbt's quoting of relations was inconsistent, leading to bugs in the below macros when run against Snowflake. This PR uses Relation objects to fix those bugs.

This build is failing because 0.10.1 is unreleased. We should merge this code only after the release is published.


clrcrl:
Most of these changes have been superseded by other changes. Except for the cleanup to the tests, which I implemented over [here](https://github.com/fishtown-analytics/dbt-utils/pull/178/commits/97d9f21e2927b165adbe89a752425e160775560f).

As a result, closing this out!

PR #64, opened 2018-05-14
Add macro to get subdomain from URL

dwallace0723:
This macro will attempt to grab a subdomain when provided a URL. If no subdomain exists, it will return `NULL`.

### DIFF
--- a/macros/web/get_subdomain.sql
+++ b/macros/web/get_subdomain.sql
@@ -36,7 +36,7 @@ Arguments:
 
 {%- set formatted_domain = "'." + domain + "'" -%}
 
-{%- set split_one = dbt_utils.split_part(string_text=url, delimiter_text="'https://'", part_number=2) -%}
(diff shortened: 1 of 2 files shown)


macros/web/get_subdomain.sql
```diff
@@ -0,0 +1,48 @@
...
+Arguments:
+    url: The name of the column containing the URL.
+    domain: The domain name
+#}
+
+{% macro get_subdomain(url, domain) -%}
+
+{%- set formatted_domain = "'." + domain + "'" -%}
+
+{%- set split_one = dbt_utils.split_part(string_text=url, delimiter_text="'https://'", part_number=2) -%}
```
drewbanin: I think this won't work for `http` urls, right? I've seen approaches to this problem use `://` in the past -- do you think that would work here?

dwallace0723: ah yeah, good catch. I think using `://` should be fine.

dwallace0723:
@drewbanin updated!

PR #66, opened 2018-05-22
Feature/current timestamp in utc

jthandy:
@drewbanin fairly self-explanatory. I actually hadn't realized that it was so difficult to get a timestamp-no-tz UTC time in each database!! spent a while with this one and tested on all 4 adapters.

I also made changes to the way that `current_timestamp()` works because there have been syntax updates in snowflake and bigquery that allow access to that function in ANSI-standard format (no parens).

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -5,5 +5,3 @@ target-path: "target"
 clean-targets: ["target", "dbt_modules"]
 macro-paths: ["macros"]
 log-path: "logs"
-
-profile: snowflake-testing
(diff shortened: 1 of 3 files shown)


drewbanin:
This all LGTM if you've tested it @jthandy. Might be worth adding integration tests too? Just 

```
select {{ dbt_utils.current_timestamp() }} as now
```

would go a long way to making me feel really good about this!

drewbanin:
@jthandy nice looking test ;)

If this were a python app, we'd have some way of mocking the time and setting it to eg. 12:00:00, then asserting that `current_timestamp()` returned the correct value. Since this is a database.... I think just making sure the function executes correctly on each adapter is going to be sufficient!

PR #67, opened 2018-06-05
fix wrong description for split_part

fred-shaykis:


drewbanin:
ooh, nice catch, thanks @fred-shaykis! Can you just change `base 1` to `default=1`? Then LGTM!

fred-shaykis:
sure thing. I think what I meant was index 1 - is that correct?

drewbanin:
ooooh, roger. Sure, that makes sense too. I'd say "1-indexed" - I think that's the prevailing nomenclature. Any combination of "index" and "1" works for me, i defer to your judgement :)

PR #70, opened 2018-06-27
Extending Pivot

ltchikindas:
Adding in:
- excluded values
- coalesce
- total column
- distinct

Changed
- else changed default to null
- aliasing renames "-" to "_"

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,8 +190,8 @@ This macro implements an "outer union." The list of tables provided to this macr
 Usage:
 ```
 {{ dbt_utils.union_tables(
-    tables=[ref('table_1'), ref('table_2')],
-    column_override={"some_field": "varchar(100)"},
+    tables=[ref('table_1'), ref('table_2')], 
+    column_override={"some_field": "varchar(100)"}, 
     exclude=["some_other_field"]
 ) }}
 ```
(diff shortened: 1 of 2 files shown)


drewbanin:
@ltchikindas I think this test is failing because the default value was changed from `0` to `null`:

screen shot 2018-06-27 at 7 17 15 pm

I'd like to maintain backwards compatibility here if possible. Do you think you can make that change?

ltchikindas:
Thanks for pointing that out! The change was intentional: 

    "then 0" would produce unexpected results on avg() because missing values would move the mean
    "then 0" would produce unexpected results on count() because missing values would be counted

I introduced coalesce=True to allow sum() to produce 0 instead of null for missing values, however, I hesitate to make coalesce=True the default, since differentiating between null (no values) and 0 (which could be sum(-1,1,0) ) seems useful.

To this end, previous values can be achieved by setting coalesce to true. The only potentially destructive change was replacing "-" in values to "_" in column names, but that feels like relatively good practice.
 
L

drewbanin:
@ltchikindas ok, I gave this some more thought and I think you're right - defaulting to `null` is probably the appropriate behavior here.

Are you able to update the integration tests to reflect the new `null` default?

The model that tests this macro is [here](https://github.com/fishtown-analytics/dbt-utils/blob/master/integration_tests/models/sql/test_pivot.sql)

It transforms [this](https://github.com/fishtown-analytics/dbt-utils/blob/master/integration_tests/data/sql/data_pivot.csv) source CSV file and confirms that it matches [this](https://github.com/fishtown-analytics/dbt-utils/blob/master/integration_tests/data/sql/data_pivot_expected.csv) expected table.

Additionally, it would be awesome if you could test some of the other new parameters! Lmk if you have any questions/issues writing tests, as we're a little light on documentation at the monent

drewbanin:
@ltchikindas any chance you're able to take another look at this sometime soon?  Sounds like this PR could benefit from your changes! https://github.com/fishtown-analytics/dbt-utils/pull/87

clrcrl:
Closing since this is stale

PR #73, opened 2018-07-20
Add coalesce_0 macro

clrcrl:
Adding a `coalesce_0` macro - I use this pattern often in my own project when I've `left join`ed a table and want to map `null`s to `0`s without changing the column name.


clrcrl:
Just realised I forgot to add a readme. brb.

drewbanin:
I'm going to close this, as it's a kimball-specific thing that warrants other kimball-specific things. I'd _love_ to create a package for all types of dimensional modeling macros! But I don't think they all belong in dbt-utils.

@clrcrl let me know if you want to discuss how to make that happen! I actually don't think it's something that fishtown analytics should own :)

PR #74, opened 2018-07-24
hash macro + update surrogate_key

jtcohen6:
Create a hash macro that performs `md5` by default. In BigQuery, convert the hash results (bytes) to hex (string). Employ this macro in `surrogate_key`.


drewbanin:
looks great! :shipit:

PR #77, opened 2018-08-16
Add support for schema.yml version 2

beckjake:
Add backwards-compatible support for v2 of schema.yml.

There are no tests that ensure schema v2 works yet, because `pip install dbt` doesn't get you a version that can parse it.


cmcarthur:
@drewbanin is definitely better positioned to both review this code and help test it

macros/schema_tests/equality.sql
```diff
@@ -1,4 +1,6 @@
-{% macro test_equality(model, arg) %}
+{% macro test_equality(model) %}
+
+{% set compare_model = kwargs.get('compare_model', kwargs.get('arg')) %}
```
drewbanin: where does `compare_model` come from here?

drewbanin:
One question, but otherwise LGTM!

macros/schema_tests/equality.sql
beckjake: It's a pretty much arbitrary key I invented to make this stuff work (in v1 it's a bit magical, in v2 I had to pick a name). It's only used in one place, this test: https://github.com/fishtown-analytics/dbt/blob/development/test/integration/008_schema_tests_test/models-v2/custom/schema.yml#L19-L20

We can change it, but it will break tests - I'd probably open a new branch off this and merge that, instead.

In hindsight, tests should reference _tags_ of dbt-utils rather than branches.

drewbanin: oh - ok, makes sense to me.

Yeah, we should use a tag. I think we should leave this branch open after merging this PR, otherwise the dbt tests will fail

beckjake:
Do not delete this branch!

PR #79, opened 2018-08-21
Star macro optional alias argument

dwallace0723:
In the current implementation of the `star` macro, if you call the macro two distinct times referencing different tables that have overlapping field names, the model will error out since the `star` macro doesn't qualify the field names with a table alias (e.g. `ERROR:  column reference "column_name" is ambiguous`). This PR adds in an optional `relation_alias` argument to prefix all generated field names using `star` with an arbitrary table alias.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ Usage:
 ```
 
 #### star ([source](macros/sql/star.sql))
-This macro generates a list of all fields that exist in the `from` relation, excluding any fields listed in the `except` argument. The construction is identical to `select * from {{ref('my_model')}}`, replacing star (`*`) with the star macro.
+This macro generates a list of all fields that exist in the `from` relation, excluding any fields listed in the `except` argument. The construction is identical to `select * from {{ref('my_model')}}`, replacing star (`*`) with the star macro. This macro also has an optional `alias` argument that will prefix all generated fields with a table alias.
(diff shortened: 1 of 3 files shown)


drewbanin:
@dwallace0723 thanks for opening this PR! Can you shoot over a quick example of how you call this macro in practice?

dwallace0723:
@drewbanin yeah, for sure. Here is an example:

```
select
  {{ dbt_utils.star(from=ref('model_a'), relation_alias='a', except=exclude_field_list_a) }},
  {{ dbt_utils.star(from=ref('model_b'), relation_alias='b', except=exclude_field_list_b) }}
from {{ ref('model_a') }} a
left join {{ ref('model_b') }} b
on a.id = b.id
```

In cases where `model_a` and `model_b` have overlapping field names, the current implementation of star will cause an error (e.g. `ERROR: column reference "column_name" is ambiguous`) since it doesn't qualify the generated field names with the tables alias in the query.

drewbanin:
ah, got it. I tend to handle this by putting each star into it's own cte. Is that something you've tried already?

dwallace0723:
@drewbanin yeah, totally. This is really just to extend the possible uses of the existing macro. At least with this optional argument, you could use it _both_ ways.

PR #80, opened 2018-09-08
Add sql_delimited_list macro

clrcrl:
Wrote this macro for a query I was writing. Thought it might be useful to share. Feedback welcome 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,11 +167,27 @@ Usage:
 ```
 
 #### group_by ([source](macros/sql/groupby.sql))
-This macro build a group by statement for fields 1...N
+This macro builds a group by statement for fields 1...N
 
 Usage:
 ```
 {{ dbt_utils.group_by(n=3) }} --> group by 1,2,3
+```
+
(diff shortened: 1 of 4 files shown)


clrcrl:
@drewbanin – those integration tests all failed locally, but I haven't touched them. Any ideas?

jthandy:
Sometimes it's useful to use other delimeters -- for example, `|` in redshift to coalesce things! i'd recommend adding a `delimeter` argument.

clrcrl:
@jthandy  –  I did consider that as well!
In that case, would you remove the `(` and `)` from the macro? I wrote it assuming it would be used for `in` statements, so put the brackets in there, but they won't be useful for coalescing.

clrcrl:
Wait a second, just realised I can pretty easily accomplish this with:
`{{ listx|join(', ') }}`
In which case I probably don't need to write a macro for it... 😅

jthandy:
Yeah I was going to mention that also but glad to see you got there on your own :D Always happy to bounce ideas around before you put together a full PR including documentation and unit tests just to make sure that we're all on the same page about how to accomplish something!

PR #82, opened 2018-09-11
make union work on bq

drewbanin:
todo:
- [x] bq compatibility
- [x] fix tests

### DIFF
--- a/integration_tests/models/sql/test_get_column_values.sql
+++ b/integration_tests/models/sql/test_get_column_values.sql
@@ -3,6 +3,7 @@
 
 
 select
+    {% set columns = columns if columns is iterable else [] %}
     {% for column in columns -%}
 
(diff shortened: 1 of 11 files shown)


integration_tests/models/sql/test_get_column_values.sql
```diff
@@ -3,6 +3,7 @@
 
 
 select
+    {% set columns = columns if columns is iterable else [] %}
```
drewbanin: i will remove this before merging

drewbanin:
this is blocking on the 0.11.1 release of dbt

drewbanin:
Awesome, good catches @jtcohen6! I think we're actually in good shape to run integration tests on snowflake/bq/redshift these days. I'm going to merge this, but I just made an issue to track those two topics.

Thanks! merging

PR #83, opened 2018-09-12
fix for excluding last column in unpivot macro

alexcampana:
Excluding the last column of a table in the unpivot macro would add an additional `union all` to the end.

### DIFF
--- a/macros/sql/unpivot.sql
+++ b/macros/sql/unpivot.sql
@@ -13,6 +13,8 @@ Arguments:
 
   {%- set exclude = exclude if exclude is not none else [] %}
 
+  {%- set include_cols = [] %}
+
   {%- set table_columns = {} %}
 
   {%- set _ = table_columns.update({table: []}) %}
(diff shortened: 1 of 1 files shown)


drewbanin:
Thanks @alexcampana - you can ignore the Circle failure -- i think that's unrelated to your PR.

I'm looking at the original logic though, and I'm confused how this even works!!

```
{%- for col in cols -%}
    {%- if col.column not in exclude -%}
      {% set _ = include_cols.append(col) %}
    {%- endif %}
{%- endfor %}

{%- for col in include_cols -%}

    select
      {%- for exclude_col in exclude %}
        {{ exclude_col }},
      {%- endfor %}
      cast('{{ col.column }}' as varchar) as field_name,
      {{ dbt_utils.safe_cast(field=col.column, type=cast_to) }} as value
    from {{ table }}
    {% if not loop.last -%}
      union all
    {% endif -%}
  {%- endfor -%}
```

drewbanin:
@alexcampana just got tests passing over here: https://github.com/fishtown-analytics/dbt-utils/pull/82

I'm waiting on a PR. Once that's merged, if you can pull in `master`, then your tests will pass and we can merge this!

drewbanin:
@alexcampana thanks for your patience! I think if you merge master, these tests should now pass!

alexcampana:
@drewbanin no problem, looks like the tests are passing now!

drewbanin:
@alexcampana I just released [0.1.16](https://github.com/fishtown-analytics/dbt-utils/releases/tag/0.1.16). Thanks for the fix!

PR #84, opened 2018-10-01
remove list_relations call in materialization

beckjake:
The caching PR I wrote removes `list_relations` from the DBWrapper, so remove the explicit call from the materialization.

This change is backwards compatible: Currently in the absence of a `relations_list`, the adapter code calls `list_relations` on its own.


integration_tests/models/materializations/test_insert_by_period.sql
```diff
@@ -0,0 +1,16 @@
+{{
```
drewbanin: Can we do something hacky to get these tests to pass? Let's just not configure this model for Postgres, making it use the `view` materialization instead

```suggestion
{% if target.type != 'postgres' %}
  {{
  ...
  }}
{% endif %}
```

drewbanin: ^ I guess this fails during parsing? Boo. Might be worth just implementing that `if/else` block

beckjake: It turns out that the circlei script doesn't run the makefile, I just made the circle config match.

PR #87, opened 2018-11-08
Add slug option to pivot macro

abelsonlive:
# What

This P.R. adds a `slug` option to the `pivot` macro so that users can generate nicely formatted aliases for their pivoted columns.

### DIFF
--- a/macros/sql/pivot.sql
+++ b/macros/sql/pivot.sql
@@ -59,7 +59,7 @@ Arguments:
     )
     {% if alias %}
       {% if slug %}
-        {% let v = v|replace(' ', '_')|lower|trim %}
+        {% let v = v|trim|replace(' ', '_') %}
       {% endif %}
       as {{ adapter.quote(prefix ~ v ~ suffix) }}
     {% endif %}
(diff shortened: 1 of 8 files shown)


macros/sql/pivot.sql
```diff
@@ -56,6 +58,9 @@ Arguments:
       end
     )
     {% if alias %}
+      {% if slug %}
+        {% set v = v|trim|replace(' ', '_') %}
```
mplovepop: Should we lowercase as well?

macros/sql/pivot.sql
```diff
@@ -41,6 +42,7 @@ Arguments:
 {% macro pivot(column,
                values,
                alias=True,
+               slug=False,
```
mplovepop: I would default this to True as well!

abelsonlive: I agree, but I didn't want to break anybody's existing implementations

drewbanin:
@abelsonlive cool! I just followed up on #70. If you're able to work with @ltchikindas to merge these two PRs, that would be really great!

clrcrl:
Closing since this is stale

PR #97, opened 2018-12-07
Feature: get url host

axelazaid:
created two macros:
* `replace`
-- functions the same as the `replace()` function, and is adaptable for redshift, snowflake, postgres, and bigquery
* `get_url_host`
-- calls the `replace` macro two times that looks for the `http://` or `https://` portion of a url--should it exist-- and replaces it with a 'blank space'. Also calls the `split_part` macro to return everything before the first `/`, or alternatively, everything before the `?` of a url.

created two models in `integration_tests`:
* `test_replace`
* `test_url_host`

updated schema files to include tests for `test_replace` and `test_url_host`
* ran and passed all tests


integration_tests/data/web/data_url_host.csv
```diff
@@ -0,0 +1,2 @@
+original_url,parsed_url
+www.outdoorvoices.com/collections/tennis/,www.outdoorvoices.com
```
jthandy: @axelazaid can you include other example URLs that I had included in my original query that I sent to you? I'd like to make sure that all of those URLs pass this test :)

PR #98, opened 2018-12-09
Feature: right macro

axelazaid:
created `right` macro:
* functions the same as the `right()` function
* is adaptable for redshift, snowflake, postgres, and bigquery

added seed file and inserted it into a test model
* seed file: `data_right`
* test file: `test_right`

updated `schema.yml` file to include tests for model
* ran and passed all tests


jthandy:
This code looks awesome. My only note is that it doesn't look like BigQuery standard SQL has a `right()` function. Their old dialect had it, but their new dialect doesn't. You have to unselect Options >> Use Legacy SQL to choose this mode--it's what we use for all BQ projects (and it's what dbt uses). 

I'm a little surprised that the tests didn't fail on this. @drewbanin -- the circle job tests everything on BQ, right? And it uses standard SQL?

This is obviously not at all urgent--I just happened to be sitting at my computer when this came in and I was excited to take a look at it :)

PR #100, opened 2018-12-10
Feature: length macro

axelazaid:
created `length` macro
* functions same as the `len()` / `length()` function
* adaptable for redshift, snowflake, postgres, and bigquery

created test seed file and inserted it into test model:
* `data_length.csv`
* `test_length`

updated `schema.yml` file to include tests for test model
* ran and passed all tests


macros/cross_db_utils/length.sql
```diff
@@ -0,0 +1,39 @@
+{% macro length(expression) -%}
+    {{ adapter_macro('dbt_utils.length', expression) }}
+{% endmacro %}
+
+
+{% macro default__length(expression) %}
```
jthandy: this would be shorter if you made the default `length({{expression}})` and then just made a redshift version that differed from that :) functionally equivalent, less code!

PR #101, opened 2018-12-10
WIP fix: updated macros for bq compatibility

axelazaid:
* three outstanding compatibility issues

### DIFF
--- a/integration_tests/models/sql/test_groupby.sql
+++ b/integration_tests/models/sql/test_groupby.sql
@@ -1,8 +1,23 @@
+with test_data as (
+    
+    select
+        'a' as column_1,
+        'b' as column_2
+    
+),
+
+grouped as (
+
+    select 
+        *,
+        count(*) as total
+
+    from test_data
+    {{ dbt_utils.group_by(2) }}
+    
+)
+
+select * from grouped
 
 
-select
-    'a' as column_1,
-    'b' as column_2,
-    count(*) as total
 
-{{ dbt_utils.group_by(2) }}
--- a/integration_tests/models/sql/test_nullcheck_table.sql
+++ b/integration_tests/models/sql/test_nullcheck_table.sql
@@ -8,8 +8,17 @@ with nulled as (
 )
 
 select
-    field_1::varchar as field_1,
-    field_2::varchar as field_2,
-    field_3::varchar as field_3
+
+{{ dbt_utils.safe_cast('field_1',
+    dbt_utils.type_string()
+)}} as field_1,
+
+{{ dbt_utils.safe_cast('field_2',
+    dbt_utils.type_string()
+)}} as field_2,
+
+{{ dbt_utils.safe_cast('field_3',
+    dbt_utils.type_string()
+)}} as field_3
 
 from nulled
--- a/integration_tests/models/sql/test_unpivot.sql
+++ b/integration_tests/models/sql/test_unpivot.sql
@@ -1,1 +1,6 @@
-{{ dbt_utils.unpivot(table=ref('data_unpivot'), cast_to='varchar', exclude=['customer_id','created_at']) }}
+{{ dbt_utils.unpivot(
+    table=ref('data_unpivot'), 
+    cast_to=dbt_utils.type_string(), 
+    exclude=['customer_id','created_at']
+
+) }}
--- a/macros/cross_db_utils/date_trunc.sql
+++ b/macros/cross_db_utils/date_trunc.sql
@@ -7,5 +7,5 @@
 {% endmacro %}
 
 {% macro bigquery__date_trunc(datepart, date) %}
-    date_trunc({{date}}, {{datepart}})
+    timestamp_trunc(cast({{date}} as timestamp), {{datepart}})
 {% endmacro %}
--- a/macros/cross_db_utils/dateadd.sql
+++ b/macros/cross_db_utils/dateadd.sql
@@ -16,8 +16,8 @@
 
 {% macro bigquery__dateadd(datepart, interval, from_date_or_timestamp) %}
 
-    date_add(
-        cast( {{ from_date_or_timestamp }} as date),
+    datetime_add(
+        cast( {{ from_date_or_timestamp }} as datetime),
         interval {{ interval }} {{ datepart }}
         )
 
--- a/macros/cross_db_utils/datediff.sql
+++ b/macros/cross_db_utils/datediff.sql
@@ -16,8 +16,8 @@
 
 {% macro bigquery__datediff(first_date, second_date, datepart) %}
 
-    date_diff(
-        {{second_date}},
+    datetime_diff(cast(
+        {{second_date}} as datetime),
         {{first_date}},
         {{datepart}}
         )
--- a/macros/sql/star.sql
+++ b/macros/sql/star.sql
@@ -23,7 +23,7 @@
 
     {%- for col in include_cols %}
 
-        {% if relation_alias %} {{ relation_alias }}.{% endif %}"{{ col }}" {% if not loop.last %},
+        {% if relation_alias %} {{ relation_alias }}.{% endif %} {{ col }} {% if not loop.last %},
         {% endif %}
 
     {%- endfor -%}
--- a/integration_tests/data/cross_db/data_split_part.csv
+++ b/integration_tests/data/cross_db/data_split_part.csv
@@ -1,4 +1,4 @@
 parts,split_on,result_1,result_2,result_3
-a|b|c,|,a,b,c
-1|2|3,|,1,2,3
-,|,,
+"a|b|c","|",a,b,c
+"1|2|3","|",1,2,3
+,"|",,,
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -2,7 +2,7 @@
 name: 'dbt_utils_integration_tests'
 version: '1.0'
 
-profile: 'integration_tests'
+profile: 'realself'
 
 source-paths: ["models"]
 analysis-paths: ["analysis"] 
--- a/macros/sql/unpivot.sql
+++ b/macros/sql/unpivot.sql
@@ -39,7 +39,7 @@ Arguments:
       {%- for exclude_col in exclude %}
         {{ exclude_col }},
       {%- endfor %}
-      cast('{{ col.column }}' as varchar) as field_name,
+      cast('{{ col.column }}' as {{ dbt_utils.type_string() }}) as field_name,
       {{ dbt_utils.safe_cast(field=col.column, type=cast_to) }} as value
     from {{ table }}
     {% if not loop.last -%}


drewbanin:
> Database Error in model test_get_tables_by_prefix_and_union (models/sql/test_get_tables_by_prefix_and_union.sql)
  Syntax error: Unexpected "|" at [4:36]

Check out the [macro code](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/get_tables_by_prefix.sql). This is macro is a wrapper over `information_schema.tables`, which does not exist in BigQuery. We could try querying the [TABLES_SUMMARY](https://cloud.google.com/bigquery/docs/datasets#tables-meta-table) table with something like:
```
select * from dbt_dbanin.__TABLES_SUMMARY__
```
----

> Database Error in model test_insert_by_period (models/materializations/test_insert_by_period.sql)
  Syntax error: Unexpected extra token COMMIT at [2:5]
  compiled SQL at target/compiled/dbt_utils_integration_tests/materializations/test_insert_by_period.sql

Insert by period is only a thing on Redshift. You're going to want to include [this](https://github.com/fishtown-analytics/dbt-utils/pull/99/files#diff-850d4e9abce8359a9efd74b0725a6acfR4) change in this PR. I'd recommend in fact branching off of that `tests-on-all-adapters` branch.

>Database Error in model test_nullcheck_table (models/sql/test_nullcheck_table.sql)
  Within standard SQL views, table and view references require explicit project IDs, but table or view integration_tests.data_nullcheck_table is not project-qualified
  compiled SQL at target/compiled/dbt_utils_integration_tests/sql/test_nullcheck_table.sql

BigQuery doesn't like this one because the macro selects from `<schema>.<table>`, whereas it's expecting `<project_id>.<schema>.<table>`. To make this work cross-platform, I'd try creating a Relation object (example [here](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/materializations/insert_by_period_materialization.sql#L61)). That would look like:

```sql
{% macro nullcheck_table(schema, table) %}

  {% set cols = adapter.get_columns_in_table(schema, table) %}
  {%- set relation = api.Relation.create(schema=schema, identifier=table) -%}

  select {{ dbt_utils.nullcheck(cols) }}
  from {{ relation }}
  
{% endmacro %}
```

I _think_ that will work here, as Relations should always interpolate a Project ID when run against BigQuery.

drewbanin:
Get tables by prefix (BQ):
```
{% macro default__get_tables_by_prefix_sql(schema, prefix, exclude='') %}
....
{% endmacro %}

{% macro bigquery__get_tables_by_prefix_sql(schema, prefix, exclude='') %}

select distinct
  concat(dataset_id, '.', table_id) as ref
  
from dbt_dbanin.__TABLES_SUMMARY__
where dataset_id = 'dbt_dbanin'
  and lower(table_id) like lower('events_%')
  and lower(table_id) not like lower('exclude')

{% endmacro %}
```


```

{% macro get_tables_by_prefix(schema, prefix, exclude='') %}

    {%- call statement('tables', fetch_result=True) %}

      {{ dbt_utils.get_tables_by_prefix_sql(schema, prefix, exclude) }}

    {%- endcall -%}

    {%- set table_list = load_result('tables') -%}

    {%- if table_list and table_list['data'] -%}
        {%- set tables = table_list['data'] | map(attribute=0) | list %}
        {{ return(tables) }}
    {%- else -%}
        {{ return([]) }}
    {%- endif -%}

{% endmacro %}
```

jthandy:
@drewbanin how should we get the dataset? you have `from dbt_dbanin.__TABLES_SUMMARY__`. do we need to modify the parameters to pass that as well? or should it be from the context?

integration_tests/models/sql/test_groupby.sql
```diff
@@ -1,8 +1,23 @@
+with test_data as (
+    
+    select
+        'a' as column_1,
```
drewbanin: try using the `safe_cast` macro here to cast this as a string field. That should fix any pg/redshift errors

drewbanin:
@jthandy we're doing a live code review - just adding some notes in here :)

integration_tests/models/sql/test_unpivot.sql
```diff
@@ -1 +1,6 @@
-{{ dbt_utils.unpivot(table=ref('data_unpivot'), cast_to='varchar', exclude=['customer_id','created_at']) }}
+{{ dbt_utils.unpivot(
+    table=ref('data_unpivot'), 
+    cast_to=dbt_utils.type_string(),
```
drewbanin: nice

macros/cross_db_utils/dateadd.sql
```diff
@@ -16,8 +16,8 @@
 
 {% macro bigquery__dateadd(datepart, interval, from_date_or_timestamp) %}
 
-    date_add(
-        cast( {{ from_date_or_timestamp }} as date),
+    datetime_add(
```
drewbanin: I think this would be better if we cast to a `timestamp` instead of a `datetime`, and then use `timestamp_add` accordingly. Timestamps have timezones, whereas datetimes do not. More info on that one [here](https://stackoverflow.com/questions/47713945/bigquery-datetime-vs-timestamp).

macros/cross_db_utils/date_trunc.sql
```diff
@@ -7,5 +7,5 @@
 {% endmacro %}
 
 {% macro bigquery__date_trunc(datepart, date) %}
-    date_trunc({{date}}, {{datepart}})
+    date_trunc(cast({{date}} as date), {{datepart}})
```
drewbanin: do you remember why you made this change? If we cast the argument to a `date`, then it will lose its hours/mins/seconds. If you call this with:
```
{{ date_trunc("'2018-01-01 12:34:56'", "hour") }}
```

I think the result will be `2018-01-01` instead of `2018-01-01 12:00:00`. Can we cast this to a `timestamp` and use `timestamp_trunc` instead?

macros/cross_db_utils/datediff.sql
```diff
@@ -16,8 +16,8 @@
 
 {% macro bigquery__datediff(first_date, second_date, datepart) %}
 
-    date_diff(
-        {{second_date}},
+    datetime_diff(cast(
```
drewbanin: same here, let's use timestamps instead of datetimes. I imagine we'll want to cast both the first and second date below, right?

macros/datetime/date_spine.sql
```diff
@@ -44,13 +44,13 @@ with rawdata as (
...
     select (
-        {{
+        cast({{
             dbt_utils.dateadd(
                 datepart,
                 "row_number() over (order by 1) - 1",
                 start_date
             )
-        }}
+        }} as date)
```
drewbanin: if we cast this to a date, we'll drop the "time" component. Can you make this work by casting as a timestamp instead of a `date`?

macros/schema_tests/cardinality_equality.sql
```diff
@@ -2,6 +2,55 @@
 
 {% set column_name = kwargs.get('column_name', kwargs.get('from')) %}
 
+
```
drewbanin: see the `equality` test below

macros/schema_tests/equality.sql
```diff
@@ -18,7 +18,7 @@
 
 
 
--- core SQL
+{% if target.type == 'bigquery' %}
```
drewbanin: The only difference between BQ and the other adapters is `except distinct`, right? Can we make a cross-platform implementation of `except` that returns `except distinct` for bigquery and `except` for everything else?

macros/sql/get_tables_by_prefix.sql
```diff
@@ -1,13 +1,12 @@
 {% macro get_tables_by_prefix(schema, prefix, exclude='') %}
+  {{ return(adapter_macro('dbt_utils.get_tables_by_prefix', schema, prefix, exclude)) }}
```
drewbanin: i don't think we need to use the `adapter_macro` here because there is only a single implementation of `get_tables_by_prefix` for all adapters.

macros/sql/star.sql
```diff
@@ -23,7 +23,7 @@
 
     {%- for col in include_cols %}
 
-        {% if relation_alias %} {{ relation_alias }}.{% endif %}"{{ col }}" {% if not loop.last %},
+        {% if relation_alias %} {{ relation_alias }}.{% endif %} {{ col }} {% if not loop.last %},
```
drewbanin: this change might break a lot of models! The `star` macro is frequently used on source tables, and those frequently have columns named things like `date` or `timestamp` that can be reserved words. Rather than always removing these quotes, can we add an argument called `quote` that defaults to True? That way, you can disable quoting where required.

integration_tests/models/sql/test_get_column_values.sql
```diff
@@ -12,3 +14,18 @@ select
...
 
 from {{ ref('data_get_column_values') }}
+
+{% else %}
+
+select
+    {% set columns = columns if columns is iterable else [] %}
+    {% for column in columns -%}
+
+        {{dbt_utils.safe_cast("sum(case when field = '" ~ column ~ "' then 1 else 0 end)", dbt_utils.type_string()) }} as count_{{ column }}
```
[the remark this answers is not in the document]
axelazaid: yup, that's exactly what's happening...

macros/cross_db_utils/dateadd.sql
axelazaid: so when i cast it to `timestamp` and use `timestamp_add` i get:
`TIMESTAMP_ADD does not support the MONTH date part` ?? haven't dug too deep as to why, but just letting you know that's the error i get

macros/cross_db_utils/date_trunc.sql
axelazaid: so in the .csv file, it looks like the expected result is looking for just the date without the timestamp associated with it. if i cast the argument to `timestamp_trunc` it's going to look for the some timestamp. that's why i cast the argument to `date_trunc`....

axelazaid: using the `timestamp_trunc` function throws this error: `No matching signature for operator != for argument types: TIMESTAMP, DATE` when i try to test `test_date_trunc`

macros/cross_db_utils/datediff.sql
axelazaid: same thing here...maybe we should talk a little more when you have time to discuss these...i might be getting the wrong idea about how the tests should work, but my general thought process around these date/time casts is based on how the result should look in the end

macros/datetime/date_spine.sql
axelazaid: see above lol

macros/schema_tests/equality.sql
axelazaid: this one's kinda tricky, along with talking about date/time casts, we should talk about this one too. :)

macros/sql/get_tables_by_prefix.sql
axelazaid: 👍

integration_tests/models/sql/test_nullcheck_table.sql
```diff
@@ -1,15 +1,45 @@
...
+
+    {{ dbt_utils.safe_cast('field_3',
+        dbt_utils.type_string()
+    )}} as field_3
+
+from nulled
+
+{% else %}
+
+with nulled as (
```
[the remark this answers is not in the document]
axelazaid: 👍

drewbanin:
Good call on that Snowflake update. This LGTM when the test pass!

PR #105, opened 2018-12-21
Adds equal_sum schema test

clausherther:
Adds a schema test macro to test whether two columns sum up to the same total. 
For example, this can be helpful in auditing transaction level fact tables vs their aggregate fact tables.


clausherther:
(Integration tested on Snowflake)

drewbanin:
Hey @clausherther, cool macro! Have you seen [expression_is_true](https://github.com/fishtown-analytics/dbt-utils#expression_is_true-source)? I think you could implement this test (for some use cases anyway) using the more general form of this test.

I think this is a totally reasonable test, but I also am thinking hard about what kinds of macros belong in dbt-utils. We definitely need better rules as to what can live in here and what should live in some other package. I'm very open to discussing this, but my instinct is that this is a pretty specific test, and it might be better suited for some other repo. I'd totally recommend that you create your own repo of tests if you find that you have a critical mass of these macros that you share across projects!

So, very happy to chat. Let me know if you have any questions!

clausherther:
I thought about that, but does `expression_is_true` work against another model? The use case I've run into quite a bit is having an order and an order line item table, and needing to make totals match up. Does that make sense? Are these too specific for the utils package?

drewbanin:
Yeah, you're right that `expression_is_true` won't work in that use case. The orders/orders_items example is a good one.... i think i've just been convinced :)

I can kick off the CI build for this PR in a moment. Will also drop some comments in the code

macros/schema_tests/equal_sum.sql
```diff
@@ -0,0 +1,34 @@
+{% macro test_equal_sum(model) %}
```
drewbanin: I think you'll want to make the signature here:
```
{% macro test_equal_sum(model, column_name, compare_model, compare_column) %}
```

Then you can get rid of the `kwargs` lines below. We needed to do the `kwargs` trick to make schema tests compatible with the `v1` and `v2` schema.yml specs, but I don't think we want to support the v1 spec for net-new tests.

macros/schema_tests/equal_sum.sql
```diff
@@ -0,0 +1,34 @@
...
+
+{% set column_name = kwargs.get('column_name', kwargs.get('arg')) %}
+{% set compare_model = kwargs.get('compare_model', kwargs.get('arg')) %}
+{% set compare_column_name = kwargs.get('compare_column', kwargs.get('arg')) %}
+
+{#-- Prevent querying of db in parsing mode. This works because this macro does not create any new refs. #}
```
drewbanin: You can remove this part -- it's only useful if the macro tries to execute a query (like getting the columns in a table) during parsing.

macros/schema_tests/equal_sum.sql
```diff
@@ -0,0 +1,34 @@
...
+{% endif %}
+
+with a as (
+
+    select sum({{ column_name }}) as sum_a from {{ model }}
+
+),
+b as (
+
+    select sum({{ compare_column_name }}) as sum_b from {{ compare_model }}
```
drewbanin: make sure to change this to `compare_column` if you change the macro signature above

clausherther:
Thx for the feedback! I also incorporated that into the `equal_rowcount` schema test macro.

clausherther:
@drewbanin actually, thought about this a bit over the holidays and I think I'll close this PR in favor of a more generalized version that likely would also replace the `equal_rowcount` schema test you merged the other day. I'll submit that today or tomorrow. Not sure how you deal with deprecations here, but it's likely no one other than me used this one anyway.

Issue #89, opened 2018-11-12
try_cast only works for strings


jwerderits:
[dbt_utils.safe_cast](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/cross_db_utils/safe_cast.sql) does not work for snowflake.

`try_cast` only works for string expressions. 

Got error `dict object` has not attribute `string_literal` when using `dbt_utils.union_tables`

Solution was to change https://github.com/fishtown-analytics/dbt-utils/blob/d2139c265f4adfb2cb9647bf933bc95c4a3e0fff/macros/sql/union.sql#L67 
to `{{ col_name }}::{{ col_type }} as {{ col.quoted }} {% if not loop.last %},{% endif %}`

jtcohen6:
Reference: https://docs.snowflake.net/manuals/sql-reference/functions/try_cast.html

dapearce:
I just upgraded dbt utils and am hitting this same issue with `dbt_utils.union_table`. The error starts in our project at version 0.16

PR #112, opened 2019-01-10
Remove try_cast

dapearce:
Per issue #89 `try_cast` only works on strings (https://docs.snowflake.net/manuals/sql-reference/functions/try_cast.html). As a result, this is causing errors in other macros (eg `union_tables`)

### DIFF
--- a/macros/cross_db_utils/safe_cast.sql
+++ b/macros/cross_db_utils/safe_cast.sql
@@ -10,11 +10,6 @@
 {% endmacro %}
 
 
-{% macro snowflake__safe_cast(field, type) %}
-    try_cast({{field}} as {{type}})
-{% endmacro %}
-
-
 {% macro bigquery__safe_cast(field, type) %}
     safe_cast({{field}} as {{type}})
 {% endmacro %}


drewbanin:
Thanks for the PR @dapearce. I think you're probably right -- the `safe_cast` functionality isn't quite where it needs to be for this to work as we expect. 

I tried to see if i could do something like
```
case
    when typeof(col) like 'VARCHAR%' then safe_cast(col as type)
    else cast(col as type)
end
```

but it looks like `typeof` works for everything _except_ for strings 🙃

I'm ok with this change, but would like to get @jthandy's thoughts too, as he's spent more time in here recently than I have.

dapearce:
The argument for the use of `safe_cast` makes sense, and yes our issue (which is preventing us from updating this package, and thus updating `dbt`), is just with the use of `safe_cast` in the `union_tables` macro. So however we fix that is fine by me. Just FYI it does also appear in the `unpivot` macro too. My jinja isn't good enough to understand what's going on or if it's correctly used there or not.

drewbanin:
#113 was released in [0.1.21](https://github.com/fishtown-analytics/dbt-utils/releases/tag/0.1.21)

dapearce:
Thanks!

PR #121, opened 2019-02-26
made changes to right() function

axelazaid:


jthandy:
@drewbanin it turns out that both Bigquery and Snowflake implement their versions of the `right()` function differently than RS / PG. This was causing downstream data to be inconsistent between the two adapters. On RS/PG, `right('test', 0) = ''` and on SF/BQ, `right('test', 0) = 'test'`. The RS/PG version is much more sensible IMO and we needed to make these output the same thing in all cases. This PR accomplishes that.

drewbanin:
oh, roger! ok, totally reasonable. Approve

PR #123, opened 2019-03-08
creates get_column_from_sql and fetch macro

### LINKED ISSUE #122: get_column_from_statement or get_column_from_sql macro
Sometimes people want to use statements to return results as a list (in particular when they can't use the `get_column_values` macro)

And then when they do, they have to handle all this weirdness around getting the values from the table object that gets returned, and handling the error that gets thrown on compilation. It either looks something like:
```sql
 {%- call statement('payment_methods', fetch_result=True) -%}

   select distinct
   lower(payment_method)
   from app_data.payments
   order by 1

  {%- endcall -%}

  {%- if execute -%}
  
  {%- set payment_methods = load_result('payment_methods').table.columns['payment_method'].values() -%}
	
  {%- else -%}
  
  {%- set payment_methods = [] -%}
      
  {%- endif -%}
```
or
```sql
    {%- call statement('payment_methods', fetch_result=True) -%}

     select distinct
     lower(payment_method)
     from app_data.payments
     order by 1

    {%- endcall -%}

    {%- set payment_method_results = load_result('payment_methods') -%}

    {%- if payment_method_results and payment_method_results['data'] -%}
        {%- set payment_methods = value_list['data'] | map(attribute=0) | list %}
    {%- else -%}
         {%- set payment_methods = [] -%}
    {%- endif -%}
```

And that's dumb.

I'd like a macro that simplifies this, so that a user can just:
```sql
   {%- call statement('payment_methods', fetch_result=True) -%}

   select distinct
   lower(payment_method)
   from app_data.payments
   order by 1

  {%- endcall -%}

  {% set payment_methods = get_column_from_statement('payment_methods', 'payment_method') %}
```
Or maybe even one step further, and make it `get_column_from_sql`?
```sql
{% set payment_methods_sql %}
   select distinct
   lower(payment_method)
   from app_data.payments
   order by 1
{% endset %}

{% set payment_methods = get_column_from_sql('payment_methods_sql', 'payment_method') %}
```

ChrisLawliss:
Opening a draft PR for feedback.

To-do:
- [x] Documentation in README.md
- [x] Tests

Closes #122


macros/sql/get_column_from_sql.sql
```diff
@@ -0,0 +1,25 @@
...
+        {% set sql_results_table = load_result('get_sql_results').table.columns %}
+        {% for column, column_values in sql_results_table.items() %}
+            {% do sql_results.update({column: column_values.values()}) %}
+        {% endfor %}
+    {%- endif -%}
+    
+    {{ return(sql_results) }}
+    
+{% endmacro %}
+
```
clrcrl: I think for utils we generally have one macro per file! Can you split this into two files? That way we can document them separately too.

macros/sql/get_column_from_sql.sql
```diff
@@ -0,0 +1,25 @@
...
+{% macro get_column_from_sql(sql, column_name) %}
+    {{ return(fetch(sql)[column_name]) }}
```
clrcrl:
```suggestion
    {{ return(dbt_utils.fetch(sql)[column_name]) }}
```

clrcrl: We namespace macros within packages 🙂

macros/sql/fetch.sql
```diff
@@ -0,0 +1,21 @@
...
+    {%- call statement('get_sql_results', fetch_result=True) -%}
+
+        {{sql}}
+
+    {%- endcall -%}
+    
+    {% set sql_results={} %}
+    
+    {%- if execute -%}
+        {% set sql_results_table = load_result('get_sql_results').table.columns %}
```
drewbanin: We can probably just return the table here, right? What's the rationale behind pushing these values into a dict?

macros/sql/get_column_from_sql.sql
```diff
@@ -0,0 +1,3 @@
+{% macro get_column_from_sql(sql, column_name) %}
+    {{ return(dbt_utils.fetch(sql)[column_name]) }}
```
drewbanin: one of the big challenges associated with dbt's current compilation scheme is that macros are called twice, once in parsing (when `execute` is false) and then again when dbt is _really_ running the macro.

The implementation of `get_column_sql` shown here will work at _runtime_ but it will return `none` at parse-time, which probably won't be expected by the caller. If we return a `table` in the `fetch` macro, then we can pluck out the specified column in `get_column_from_sql`. If the table returned by fetch is `none` (ie. we're in parsing), then we can just return an empty list here! That should simplify the implementation logic for callers of this macro

macros/sql/fetch.sql
clrcrl: That was my rationale – to operate on the Table, you need to know about agate's [Table object](https://agate.readthedocs.io/en/1.6.1/api/table.html), which is one more set of docs to go look up

I can see someone finding `fetch` to be generally useful separate from `get_column_from_sql`, so wanted to return an object that was familiar to a user (i.e. a dictionary).

I am happy either way though!

macros/sql/fetch.sql
```diff
@@ -0,0 +1,21 @@
...
+        {% for column, column_values in sql_results_table.items() %}
```
clrcrl: I was rushing out the door last time I was looking at this. I think this is a slightly nicer naming (tbh I probably named these in the first place)
```suggestion
        {% for column_name, column in sql_results_table.items() %}
        {% do sql_results.update({column_name: columns.values()}) %}
```

macros/sql/fetch.sql
EdPre: Yes +1 as a pythonista I want a dict pl0x

EdPre:
This is awesome

macros/sql/get_column_from_sql.sql
clrcrl: I still don't get this comment...

ChrisLawliss:
Closing this PR to open a fetch macro branch/PR that is more limited in scope

PR #131, opened 2019-04-04
Add utils for logging to the command line

clrcrl:
I thought about how to test this, but it gets pretty hard given the output depends on the clock time. I wanted to do a regex match on the output, but Jinja doesn't support this.

In the end, I thought it would just be easiest to test if the output is a string 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,6 +393,37 @@ Usage:
 ```
 
 ---
+### Logger
+#### pretty_time ([source](macros/logger/pretty_time.sql))
+This macro returns a string of the current timestamp, optionally taking a datestring format.
+```sql
+{#- This will return a string like '14:50:34' -#}
+{{ dbt_utils.pretty_time() }}
+
+{#- This will return a string like '2019-05-02 14:50:34' -#}
+{{ dbt_utils.pretty_time(format='%Y-%m-%d %H:%M:%S') }}
+```
+
+#### pretty_log_format ([source](macros/logger/pretty_log_format.sql))
+This macro formats the input in a way that will print nicely to the command line when you `log` it.
+```sql
+{#- This will return a string like:
+"11:07:31 + my pretty message"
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -393,6 +393,38 @@ Usage:
...
+```
+
+It is expected that this macro will be used in conjuction with the [log](https://docs.getdbt.com/reference#log) macro.
+```
+{{ log(dbt_utils.pretty_output_msg("my pretty message"), info=True) }}
+```
+
+```
+11:07:28 | 1 of 1 START table model analytics.fct_orders........................ [RUN]
+         + 11:07:31 | my pretty message
```
drewbanin: When I saw this PR, I imagined the output would instead look like:
```
11:07:31 + my pretty message
```

What's the motivation behind placing the timestamp on the right side of the divider?

clrcrl: Ah, so I was largely following the pattern from [here](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/materializations/insert_by_period_materialization.sql#L118) and in fact means to swap that out.

I like having it indented as it makes it clear that it's being called from within a model (to me at least). But happy to change it if you feel it's unintuitive.

clrcrl:
Yah can for sure add a `log_info` macro – I back-and-forth-ed as to whether I should actually call the `log` function or not, so very happy to do that :)

Is the naming okay? I took inspiration from `pretty print`, but am also happy to rename it.

drewbanin:
I'd maybe just consider changing `pretty_output_msg` to `pretty_log_format` or similar, but otherwise I'm happy with these `pretty` names :)

README.md
drewbanin: yeah, I'd go ahead and swap it around I think unless you feel strongly that it should work the way you've already implemented

PR #132, opened 2019-04-18
replaced get_columns_in_table with get_columns_in_relation

Nerdenberger:


macros/sql/nullcheck_table.sql
```diff
@@ -1,8 +1,8 @@
 {% macro nullcheck_table(schema, table) %}
 
-  {% set cols = adapter.get_columns_in_table(schema, table) %}
+  {% set cols = adapter.get_columns_in_relation(this)) %}
```
jtcohen6: @Nerdenberger I think you have an extra close paren here:
```suggestion
  {% set cols = adapter.get_columns_in_relation(this) %}
```

macros/schema_tests/equality.sql
```diff
@@ -13,7 +13,7 @@
 {% set schema = model.schema %}
 {% set model_a_name = model.name %}
 
-{% set dest_columns = adapter.get_columns_in_table(schema, model_a_name) %}
+{% set dest_columns = adapter.get_columns_in_relation(this) %}
```
drewbanin: I think we're going to want to use `model` here -- `this` will refer to the currently running model, whereas we want the mode specified as an argument to the macro.

macros/sql/nullcheck_table.sql
drewbanin: in addition the extra paren, we'll want to construct a [Relation object](https://docs.getdbt.com/reference#section-creating-relations) here from the `schema` and `table` arguments passed into the macro.

This is kind of tricky because Relations should be constructed with a database name, but the `nullcheck_table` macro doesn't accept a database. @jtcohen6 do you think we should add an arg? Unsure how much this macro is used these days

macros/sql/star.sql
```diff
@@ -12,7 +12,7 @@
     {%- endif -%}
 
     {%- set include_cols = [] %}
-    {%- set cols = adapter.get_columns_in_table(schema_name, table_name) -%}
+    {%- set cols = adapter.get_columns_in_relation(this) -%}
```
drewbanin: If `from` is a Relation, then we can use that directly. I think we should change the contract of this macro to _require_ that a Relation is provided. It's going to be really hard for us to marshal a `schema.table` string into a Relation in an intelligent way. @jtcohen6 agree?

macros/sql/union.sql
```diff
@@ -22,7 +22,7 @@
             {%- set schema, table_name = (table | string).split(".") -%}
         {%- endif -%}
 
-        {%- set cols = adapter.get_columns_in_table(schema, table_name) %}
+        {%- set cols = adapter.get_columns_in_relation(this) %}
```
drewbanin: same here - we should use `table` directly if it's a Relation object

macros/sql/unpivot.sql
```diff
@@ -25,7 +25,7 @@ Arguments:
     {%- set schema, table_name = (table | string).split(".") -%}
   {%- endif -%}
 
-  {%- set cols = adapter.get_columns_in_table(schema, table_name) %}
+  {%- set cols = adapter.get_columns_in_relation(this) %}
```
drewbanin: see above

macros/sql/star.sql
jtcohen6: Agree. The only way I could imagine accepting schema + identifier args is if we called the api get (create) relation function as well, which seems a worse pattern than enforcing that the argument be a relation.

macros/sql/nullcheck_table.sql
jtcohen6: I think we could update line 6 below to be `from {{relation}}` and thereby accept a (default) database above. The [`nullcheck` macro](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/nullcheck.sql) just accepts a Jinja list of columns, it doesn't do any introspective work itself.

To answer your larger question, tbqh I've never seen this macro used, and it looks like it was last touched a while ago.

PR #133, opened 2019-04-26
Replace deprecated adapter methods with relations

### LINKED ISSUE #129: 0.13.0 compatibility
Updates required to the following macros to remove deprecation warnings:
- macros/schema_tests/equality.sql
- macros/sql/nullcheck_table.sql
- macros/sql/star.sql
- macros/sql/unpivot.sql
- macros/sql/union.sql

jtcohen6:
* Fixes #129
* Closes #132
* Some breaking changes to contracts used by helper macros

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,7 +440,7 @@ models:
     project-name:
         post-hook: "grant select on {{ this }} to db_reader"
 ```
-A useful workaround is to change the above post-hook t  o:
+A useful workaround is to change the above post-hook to:
 ```yaml
         post-hook: "grant select on {{ this.schema }}.{{ this.name }} to db_reader"
 ```
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ from {{ref('my_model')}}
 ```
 
 #### union_tables ([source](macros/sql/union.sql))
-This macro implements an "outer union." The list of tables provided to this macro will be unioned together, and any columns exclusive to a subset of these tables will be filled with `null` where not present. The `column_override` argument is used to explicitly assign the column type for a set of columns. The `source_column_name` argument is used to change the name of the`_dbt_source_table` field.
+This macro implements an "outer union." The list of relations provided to this macro will be unioned together, and any columns exclusive to a subset of these tables will be filled with `null` where not present. The `column_override` argument is used to explicitly assign the column type for a set of columns. The `source_column_name` argument is used to change the name of the`_dbt_source_table` field.
 
 Usage:
 ```
@@ -440,7 +440,7 @@ models:
     project-name:
         post-hook: "grant select on {{ this }} to db_reader"
 ```
-A useful workaround is to change the above post-hook to:
+A useful workaround is to change the above post-hook t  o:
 ```yaml
         post-hook: "grant select on {{ this.schema }}.{{ this.name }} to db_reader"
 ```
--- a/macros/cross_db_utils/_is_relation.sql
+++ b/macros/cross_db_utils/_is_relation.sql
@@ -1,5 +1,5 @@
-{% macro _is_relation(arg, macro) %}
-    {%- if execute and arg.name is none -%}
-        {%- do exceptions.raise_compiler_error("Macro " ~ macro ~ " expected a Relation but received the value: " ~ arg) -%}
+{% macro _is_relation(obj, macro) %}
+    {%- if not (obj is mapping and obj.get('metadata', {}).get('type', '').endswith('Relation')) -%}
+        {%- do exceptions.raise_compiler_error("Macro " ~ macro ~ " expected a Relation but received the value: " ~ obj) -%}
     {%- endif -%}
 {% endmacro %}
--- a/macros/cross_db_utils/_is_relation.sql
+++ b/macros/cross_db_utils/_is_relation.sql
@@ -1,0 +1,5 @@
+{% macro _is_relation(arg, macro) %}
+    {%- if execute and arg.name is none -%}
+        {%- do exceptions.raise_compiler_error("Macro " ~ macro ~ " expected a Relation but received the value: " ~ arg) -%}
+    {%- endif -%}
+{% endmacro %}
--- a/macros/sql/get_tables_by_prefix.sql
+++ b/macros/sql/get_tables_by_prefix.sql
@@ -1,16 +1,24 @@
-{% macro get_tables_by_prefix(schema, prefix, exclude='') %}
+{% macro get_tables_by_prefix(schema, prefix, exclude='', database=target.database) %}
 
-    {%- call statement('tables', fetch_result=True) %}
+    {%- call statement('get_tables', fetch_result=True) %}
 
       {{ dbt_utils.get_tables_by_prefix_sql(schema, prefix, exclude) }}
 
     {%- endcall -%}
 
-    {%- set table_list = load_result('tables') -%}
+    {%- set table_list = load_result('get_tables') -%}
 
-    {%- if table_list and table_list['data'] -%}
-        {%- set tables = table_list['data'] | map(attribute=0) | list %}
-        {{ return(tables) }}
+    {%- if table_list and table_list['table'] -%}
+        {%- set tbl_relations = [] -%}
+        {%- for row in table_list['table'] -%}
+            {%- set tbl_relation = adapter.get_relation(database, row.table_schema, row.table_name) -%}
+            {%- if not tbl_relation -%}
+                {%- set tbl_relation = api.Relation.create(database, row.table_schema, row.table_name) -%}
+            {%- endif -%}
+            {%- do tbl_relations.append(tbl_relation) -%}
+        {%- endfor -%}
+                
+        {{ return(tbl_relations) }}
     {%- else -%}
         {{ return([]) }}
     {%- endif -%}
--- a/macros/sql/get_tables_by_prefix.sql
+++ b/macros/sql/get_tables_by_prefix.sql
@@ -2,7 +2,7 @@
 
     {%- call statement('get_tables', fetch_result=True) %}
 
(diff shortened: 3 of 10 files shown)


README.md
```diff
@@ -440,7 +440,7 @@ models:
     project-name:
         post-hook: "grant select on {{ this }} to db_reader"
 ```
-A useful workaround is to change the above post-hook to:
+A useful workaround is to change the above post-hook t  o:
```
drewbanin: 👁 👁

macros/sql/get_tables_by_prefix.sql
```diff
@@ -1,16 +1,24 @@
...
-    {%- set table_list = load_result('tables') -%}
+    {%- set table_list = load_result('get_tables') -%}
 
-    {%- if table_list and table_list['data'] -%}
-        {%- set tables = table_list['data'] | map(attribute=0) | list %}
-        {{ return(tables) }}
+    {%- if table_list and table_list['table'] -%}
+        {%- set tbl_relations = [] -%}
+        {%- for row in table_list['table'] -%}
+            {%- set tbl_relation = adapter.get_relation(database, row.table_schema, row.table_name) -%}
```
drewbanin: What's the rationale behind calling `get_relation` here?

macros/sql/get_tables_by_prefix_sql.sql
```diff
@@ -5,7 +5,7 @@
 {% macro default__get_tables_by_prefix_sql(schema, prefix, exclude='') %}
 
         select distinct 
-            table_schema || '.' || table_name as ref
+            table_schema as "table_schema", table_name as "table_name"
```
drewbanin: Should we try to update this macro to accept a database name? As it stands, `get_tables_by_prefix` accepts a database, but this macro only queries for tables in the database that dbt is connected to

macros/sql/star.sql
```diff
@@ -5,14 +5,8 @@
...
 
-    {%- if from.name -%}
-        {%- set schema_name, table_name = from.schema, from.name -%}
-    {%- else -%}
-        {%- set schema_name, table_name = (from | string).split(".") -%}
-    {%- endif -%}
-
     {%- set include_cols = [] %}
-    {%- set cols = adapter.get_columns_in_table(schema_name, table_name) -%}
+    {%- set cols = adapter.get_columns_in_relation(from) -%}
```
drewbanin: I think there might be merit to adding a check here that returns an error if a non-relation is provided to this macro (and others). This would have the added benefit of raising an error if someone passes an ephemeral model into `star`. 

Maybe:
```
{% if from.get('name') is none %}
  {% do exceptions.raise_compiler_error("Macro star() expected a Relation but received the value: " ~ from) %}
{% endif %}
```

or something. We could wrap this up in its own little macro as a helper :)

macros/sql/get_tables_by_prefix.sql
jtcohen6: I was under the (likely false) impression that, if a relation already exists in dbt's cache, `get_relation` is a less expensive operation than `api.Relation.create`. If these are functionally equivalent, or if the extra lines of code are not worth the minuscule benefit, I'll happily go for the latter

macros/sql/star.sql
jtcohen6: Good idea! We might add this error to all of the macros that now expect relation arguments in a break-y sort of way. Two q's:
    * Is there a better way to infer the argument type, or is `arg.get('name') != none` the best way to confirm that `arg` is a relation?
    * Is there a Jinja method you know of that lets us pass the name of the current macro? Relevant if we were to make `error_if_not_relation(arg)` its own macro, and wanted to pretty print a dynamic compiler error `"Macro " ~ my_macro ~ " expected a Relation for argument " ~ arg"`

README.md
jtcohen6: 👁‍🗨

macros/sql/get_tables_by_prefix_sql.sql
jtcohen6: Good call, will update

README.md
drewbanin:
```suggestion
A useful workaround is to change the above post-hook to:
```

macros/sql/get_tables_by_prefix.sql
drewbanin: I don't think there would be a penalty for calling `get_relation` here (you're right, it should be cached) but I think in this case it's not necessary

macros/sql/star.sql
drewbanin: These are really good questions!

I don't think we have a great mechanism in place to assert the types of objects from the jinja context. We could conceivably add a context function or a jinja test (like `abc is Relation`) but that doesn't exist today :/. I think our best option is going to be to look for a `name` attribute. Here's some example code I played around with:

```
{% set things = [
    'abc.def',
    api.Relation.create(schema='ghi', identifier='jkl'),
] %}


{% for thing in things %}
    {{ log(thing ~ " is relation? " ~ (thing is mapping and thing.name is defined), info=true) }}
{% endfor %}
```

Let's be sure to test this with https://github.com/fishtown-analytics/dbt/pull/1416

We also don't have a way to get the macro name from the macro context. I think we should just anticipate making a macro to handle raising the exception, then pass in a string to use as the macro name from the caller.

jtcohen6: Right on. Is an `is_relation()` macro something that I should add to utils now, or do you anticipate creating it within the context of dbt proper?

drewbanin: Yeah - let's add it to dbt-utils! Maybe name it something like `_is_relation()` since I wouldn't be comfortable with folks using this externally (yet). If we had some way of locally scoping the macro I'd suggest that, but we can just indicate its private-ness with a naming convention for now.

jtcohen6: Possibly related to the PR you posted: `from.name` _does_ return a value, but `from.get('name')` _does not_. The macro fails the Relation test as written above.

drewbanin: Hmm, yeah, we'll probably want to call `from.name` then. @beckjake off the top of your head, can you think of a better way to tell if an object is a Relation in the jinja context than `obj is mapping and obj.name is not none`?

beckjake: Yeah, `.name` is a property that actually resolves to `.path.get('identifier')`.

I think I'd do ``obj is mapping and 'type' in obj.get('metadata', {}))`. We should probably add a jinja test for this, or something.

beckjake: Ew, maybe you have to even check the type string:

`obj is mapping and obj.get('metadata', {}).get('type', '').endswith('Relation')`

jtcohen6: Implemented the most rigorous logic in a macro `_is_incremental()`, which returns the following exception:
```
Compilation Error
  Macro star expected a Relation but received the value: hello

  > in macro _is_relation (macros/cross_db_utils/_is_relation.sql)
  > called by macro star (macros/sql/star.sql)
  > called by <Unknown>
```

drewbanin:
@jtcohen6 is this ready for a final review?

macros/sql/get_tables_by_prefix_sql.sql
```diff
@@ -1,25 +1,25 @@
...
 {% endmacro %}
 
-{% macro default__get_tables_by_prefix_sql(schema, prefix, exclude='') %}
+{% macro default__get_tables_by_prefix_sql(schema, prefix, exclude='', database=target.database) %}
 
         select distinct 
-            table_schema || '.' || table_name as ref
-        from information_schema.tables
+            table_schema as "table_schema", table_name as "table_name"
+        from {{database}}.information_schema.tables
```
jtcohen6: Fully agreed. I had tried using the `adapter.quote()` method (necessary in the BQ version below, given our BQ test project name), but it caused errors on Snowflake. Noted for our future selves.

macros/sql/get_tables_by_prefix_sql.sql
```diff
@@ -1,25 +1,25 @@
-{% macro get_tables_by_prefix_sql(schema, prefix, exclude='') %}
+{% macro get_tables_by_prefix_sql(schema, prefix, exclude='', database=target.database) %}
```
drewbanin: I think this will work great, but there's an important thing to know about python (and therefore jinja, I assume?) argument defaults. Check out this article on [mutable default arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

This code is A-OK because `target.database` is a string, which is immutable. It would only be a problem if the code read:
```
{% macro get_tables_by_prefix_sql(schema, prefix, exclude='', databases=[]) %}
```
or similar.

macros/sql/get_tables_by_prefix_sql.sql
drewbanin: We spent _a lot_ of time getting this working right with quoting in dbt-core. I think this code is fine for now, but in general, we'll probably want to create some sort of object (analogous to Relations) which encapsulates a database + schema and any applicable quoting configs.

This code won't work if a database is named `my-database` on eg. Snowflake. While a user could pass in their database string as `"my-database"`, that would cause problems when we create the Relation object in `get_tables_by_prefix`. I'm ok with shipping this as-is, but let's keep examples like this in mind as we improve dbt's understanding of databases and quoting

macros/cross_db_utils/_is_relation.sql
```diff
@@ -0,0 +1,5 @@
+{% macro _is_relation(obj, macro) %}
```
drewbanin: This is the intended behavior, but it's important to note: This will raise if the `ref` of an ephemeral model is passed in here. The `ref()` function returns a string for ephemeral models, vs. a Relation for any non-ephemeral model (or source). I believe all of the macros here that use `_is_relation` are expecting a table/view, so that is A-OK!

PR #135, opened 2019-05-08
Added quote_identifiers argument

carlyfk:
Indicates whether to surround column aliases with double quotes or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,6 +330,7 @@ Arguments:
     - suffix: Column alias postfix, default is blank
     - then_value: Value to use if comparison succeeds, default is 1
     - else_value: Value to use if comparison fails, default is 0
+    - quote_identifiers: Whether to surround column aliases with double quotes, default is true
 
 #### unpivot ([source](macros/sql/unpivot.sql))
 This macro "un-pivots" a table from wide format to long format. Functionality is similar to pandas [melt](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.melt.html) function.
--- a/macros/sql/pivot.sql
+++ b/macros/sql/pivot.sql
@@ -36,6 +36,7 @@ Arguments:
     suffix: Column alias postfix, default is blank
     then_value: Value to use if comparison succeeds, default is 1
     else_value: Value to use if comparison fails, default is 0
+    quote_identifiers: Whether surround column aliases with double quotes, default is true
 #}
 
 {% macro pivot(column,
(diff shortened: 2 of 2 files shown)


drewbanin:
Thanks for making this PR @carlyfk! Can you tell me more about the problem you had which was solved by this PR? It looks to me like a Snowflake column caps/quoting thing, right?

dbt-utils doesn't do a great job of consistently quoting columns/tables/schemas/databases in general, and I'm super interested in trying to fix that! I think PRs like this are a good idea, but I'm also interested in thinking about a more general approach to solving this type of problem for _every_ macro in here!

Let me know about the use case (for my edification). Can you also update the `README.md` to document this new argument? I can see about that test failure -- it's something on our end and not your code I think :)

carlyfk:
@drewbanin Thanks for your feedback! I think the issue was that one of my columns being generated is called *NAME*. The macro then outputs the SQL and NAME becomes "NAME". And then the rest of my downstream code fails because I'm referring to the column by NAME and not "NAME". By removing the forced quotes in the macro, I don't have to change any of my references downstream. Btw, this only became an issue with Snowflake and was not an issue in Redshift.

I'll update the README.md document as well.  :)

As a side note, this is my first PR and I'm a relative beginner in git. I've managed to completely jack up my codebase as a result and am not sure how to deal with the whole submodule concept. Any recs for blogs I can read on how to do this cleanly?

drewbanin:
ok, thanks for the additional context!

> this is my first PR and I'm a relative beginner in git

That's amazing - this is super well done and I assumed this wasn't your first time around the block. Nice work!

> I've managed to completely jack up my codebase as a result and am not sure how to deal with the whole submodule concept

Hmmm, submodules are definitely painful, but you shouldn't need one here! Did you clone dbt-utils into your `dbt_modules` directory? Instead, you should be able to use a `local` package for development. Check out an example [here](https://docs.getdbt.com/docs/package-management). We could definitely do a better job of documenting this type of workflow! Let me know if that doesn't fix this for you -- happy to help!

I just kicked off the tests here - are you ready to merge this when they pass?

carlyfk:
Sorry, I'm not getting email notifications when you reply so I no idea. Let me know I need to do to merge!

drewbanin:
This is merged! We'll cut a new release, `v0.2.0`, which includes this change (and some others) in the coming days!

Thanks for your contribution to dbt-utils!!

PR #136, opened 2019-05-18
Resolve "get_column_values returns an error if passed a `ref`, but the model has not been previously built"

emilieschario:
This is a **draft** PR

Closes https://github.com/fishtown-analytics/dbt-utils/issues/119

This PR update the `get_column_values` macro to handle cases where the model has not previously been built. 

This introduces two new arguments to the macro: `fail` and `default`.
* I thought "fail" was a bit more explicit than just default. The default setting for fail is True, meaning that this does not change the current behavior. 
* When fail is set to false and the model doesn't exist, the default value is passed in as a result. This will allow compilation to run successfully, but it will fail when running. 
* If the model exists, there are no changes to the behavior. 

Questions:
* Should the "fail" argument be assumed based on the value of default? (none would be fail = True, anything else would be fail = False). 
* I'm not sure what the guidelines are for logging tone/syntax; let me know if these need to change. 
* What tests, if any, does this need? I don't know that we need to test the failure state, but your call!

Not sure how important commit messages are to reviews, but I worked on this in https://gitlab.com/gitlab-data/analytics/merge_requests/958/.


macros/sql/get_column_values.sql
```diff
@@ -11,19 +10,45 @@ Returns:
...
+{#-- Prevent querying of db in parsing mode. This works because this macro does not create any new refs. #}
+    {%- if not execute -%}
+        {{ return('') }}
+    {% endif %}
+{#--  #}
+
+    {%- set table_value = adapter.get_relation(database=table.database,
+                                          schema=table.schema,
+                                         identifier=table.identifier) -%}
+
```
drewbanin: Let's move the `default`-related logic out of the statement.  Can you do something like:
```
{% if target_relation is none and default is none %}
  -- raise an exception here
{% elif target_relation is none and default is not none %}
  -- return here to short circuit evaluation of the macro
{% endif %}
```

Both `exceptions.raise_compiler_error` and `return()` will halt execution of this macro. While `exceptions.raise_compiler_error` will bubble up an error into the dbt-core code, `return()` will just yield execution back to the calling jinja code. You can do `{{ return(default) }}` to avoid running a `select` statement against the database here too!

Rather than querying the database, we can use the `return()` function to short-circuit evaluation of this macro. This has two benefits:
1. It avoids running an unnecessary query
2. It preserves the type supplied by the caller

So, If you execute:
```
{% set values = dbt_utils.get_column_values(ref('other_model'), 'type', default=[]) %}
```
The `values` variable will be a list object if `ref('other_model')` does not exist. The code you have here would execute `select '[]' as value`, which would return a string like `"[]"`, which would probably be harder to operate on.

macros/sql/get_column_values.sql
```diff
@@ -11,19 +10,45 @@ Returns:
...
+    {%- call statement('get_column_values', fetch_result=true) %}
+
+        {%- if not table_value and fail -%}
+
+          {{ log("Table doesn't exist..") }}
```
drewbanin:
```suggestion
```

I don't think we need this log line, as the exception you have here will show up in the debug logs too

macros/sql/get_column_values.sql
```diff
@@ -11,19 +10,45 @@ Returns:
...
+          {{ exceptions.raise_compiler_error("Table to pull columns from doesn't exist. Try running " ~ table ~ " and its dependencies first.") }}
```
drewbanin:
```suggestion
          {{ exceptions.raise_compiler_error("In get_column_values(): relation " ~ table ~ " does not exist and no default value was provided.") }}
```

I try to avoid being prescriptive in error messages unless we can be really sure about why the failure occurred. While the error message you included here is perfect if this macro is pointing to an unbuilt model, this error can also happen if:
1. a source table does not exist
2. the user made a typo when manually creating a relation

The other annoying thing to say here is that this relation might be a `view`. I tend to use the word `relation` which encompasses both tables and views, but is probably less readily understood by the majority of dbt users. Super open to keeping `table` here for clarity - call that how you see it!

emilieschario: I really like your version (and have updated this to reflect that). I'm curious about `relation` vs `model` for word choice, since the docs seem to reference "your dbt models" as an inclusive term.

macros/sql/get_column_values.sql
emilieschario: I've implemented this feedback! Thanks so much for it. I hadn't considered preserving type, but that makes so much sense. Just as an FYI, that does introduce weirdness about how to supply the default value _into_ the macro, so I've done that in the if/else statement surrounding value list

emilieschario:
TODO
- [ ] Add test

emilieschario:
@drewbanin it looks like it's passing the postgres tests but failing on a missing redshift-related env var. Can you help me with this? 

Here's the error
```
Running with dbt=0.14.0
Encountered an error:
Compilation Error
  Env var required but not provided: 'CI_REDSHIFT_DBT_HOST'
Exited with code 2
```

emilieschario:
Also, I'm not sure if I'm supposed to update my original message with my final write up or post a new one. 😅

emilieschario:
Before default value added:
```
20:09:42 ~/repos/dbt-utils/integration_tests/models (master %) $ dbt compile ; say beep
Running with dbt=0.13.1
Found 34 models, 48 tests, 0 archives, 0 analyses, 209 macros, 0 operations, 38 seed files, 0 sources

20:14:48 | Concurrency: 8 threads (target='dev')
20:14:48 |
Encountered an error:
Runtime Error
  Compilation Error in model test_get_column_values_relation_does_not_exist (models/sql/test_get_column_values_relation_does_not_exist.sql)
    In get_column_values(): relation ANALYTICS.emilie_scratch.sample does not exist and no default value was provided.

    > in macro statement (macros/core.sql)
    > called by macro get_column_values (macros/sql/get_column_values.sql)
    > called by model test_get_column_values_relation_does_not_exist (models/sql/test_get_column_values_relation_does_not_exist.sql)
```
Compilation should fail because the model doesn't exist yet. 

After default value added:
```
20:14:58 ~/repos/dbt-utils/integration_tests/models (master %) $ dbt compile ; say beep
Running with dbt=0.13.1
Found 34 models, 48 tests, 0 archives, 0 analyses, 209 macros, 0 operations, 38 seed files, 0 sources

20:16:16 | Concurrency: 8 threads (target='dev')
20:16:16 |
20:16:28 | Done.
```
We've set a default value, so even though it doesn't fail, it still compiles.

drewbanin:
Thanks @emilieschario! Running the test with database creds presently :)

drewbanin:
This test failure isn't your fault -- we need to merge [PR for 0.14.0 compatibility](https://github.com/fishtown-analytics/dbt-utils/pull/149) to fix that test. Let me see about doing that...

drewbanin:
Ok - if you pull `master`, that should fix these tests! The latest commit on `master` makes this package compatible with dbt >= 0.14.0 - let me know if you have any issues with the upgrade

emilieschario:
@drewbanin Looks like the same env var problem is causing the failure again. I just pulled the latest master.

drewbanin:
Hey @emilieschario - just kicked off the tests!

integration_tests/models/sql/sample.sql
```diff
@@ -0,0 +1 @@
+SELECT 1 as sample_column
```
drewbanin: I think the tests are failing because `sample` is a reserved work on Snowflake. Can you call this model something different (and change any downstream references)? I can kick off the tests again once that's done

integration_tests/models/sql/test_get_column_values_relation_does_not_exist.sql
```diff
@@ -0,0 +1,29 @@
...
+    {% for column in columns -%}
+
+        sum(case when field = '{{ column }}' then 1 else 0 end) as count_{{ column }}
+        {%- if not loop.last %},{% endif -%}
+
+    {%- endfor %}
+
+from {{ ref('data_get_column_values') }}
+
+{% else %}
```
drewbanin: what's up with the cast to `string` here?

integration_tests/models/sql/test_get_column_values_relation_does_not_exist.sql
```diff
@@ -0,0 +1,29 @@
+{% set columns = dbt_utils.get_column_values(ref('test_model'), 'example_column', default='') %}
```
drewbanin:
```suggestion
{% set columns = dbt_utils.get_column_values(ref('test_model'), 'example_column', default=[]) %}
```

Not super important, but the default value should probably be a list. That lets you remove the `{% set columns = columns if columns is iterable else [] %}` line, which sort of masks the behavior that this test is trying to exercise!

macros/sql/get_column_values.sql
```diff
@@ -11,19 +10,44 @@ Returns:
...
+
+    {%- call statement('get_column_values', fetch_result=true) %}
+
+        {%- if not target_relation and default is none -%}
+
+          {{ exceptions.raise_compiler_error("In get_column_values(): relation " ~ table ~ " does not exist and no default value was provided.") }}
+
+        {%- elif not target_relation and default is not none -%}
+
+          {{ log("Table doesn't exist..") }}
```
drewbanin:
```suggestion
          {{ log("Relation " ~ table ~ " does not exist. Returning the default value: " ~ default) }}
```

macros/sql/get_column_values.sql
```diff
@@ -11,19 +10,44 @@ Returns:
...
+          {{ log("Trying to not fail..") }}
```
drewbanin:
```suggestion
```

drewbanin:
Also if you don't mind merging master before rebasing, I just merged that integration test fix here: #151

PR #142, opened 2019-06-14
Adds 3 parameters to unpivot: remove, field_name, value_name

remigabillet:
This makes `dbt_utils.unpivot` a bit more like `pandas.melt` by adding the ability :
- to name the resulting columns
- to remove some columns from the operation

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,7 +337,7 @@ This macro "un-pivots" a table from wide format to long format. Functionality is
 
 Usage:
 ```
-{{ dbt_utils.unpivot(table=ref('table_name'), cast_to='datatype', exclude=[<list of columns to exclude from unpivot>]) }}
+{{ dbt_utils.unpivot(table=ref('table_name'), cast_to='datatype', exclude=[<list of columns to exclude from unpivot>], remove=[<list of columns to remove>], field_name=<column name for field>, value_name=<column name for value>) }}
 ```
 
 Example:
@@ -364,7 +364,10 @@ Arguments:
 
     - table: Table name, required
     - cast_to: The data type to cast the unpivoted values to, default is varchar
-    - exclude: A list of columns to exclude from the unpivot.
+    - exclude: A list of columns to exclude from the unpivot operation but keep in the resulting table.
+    - remove: A list of columns to remove from the resulting table.
+    - field_name: column name in the resulting table for field
+    - value_name: column name in the resulting table for value
 
 ---
 ### Web
--- a/integration_tests/models/sql/schema.yml
+++ b/integration_tests/models/sql/schema.yml
@@ -5,7 +5,7 @@ models:
     tests:
       - dbt_utils.equality:
           arg: ref('data_generate_series')
-        
+
   - name: test_get_column_values
     columns:
       - name: count_a
@@ -13,43 +13,43 @@ models:
           - accepted_values:
               values:
                 - '1'
-            
+
       - name: count_b
         tests:
           - accepted_values:
               values:
                 - '1'
-            
+
       - name: count_c
         tests:
           - accepted_values:
               values:
                 - '1'
-            
+
       - name: count_d
         tests:
           - accepted_values:
               values:
                 - '1'
-            
+
       - name: count_e
         tests:
           - accepted_values:
               values:
                 - '1'
-            
+
       - name: count_f
         tests:
           - accepted_values:
               values:
                 - '1'
-            
+
       - name: count_g
         tests:
           - accepted_values:
               values:
                 - '5'
-          
+
   - name: test_get_tables_by_prefix_and_union
     columns:
       - name: event
@@ -60,42 +60,47 @@ models:
           - dbt_utils.at_least_one
           - not_null
           - unique
-      
+
   - name: test_nullcheck_table
     columns:
       - name: field_1
         tests:
           - not_empty_string
-      
+
       - name: field_2
         tests:
           - not_empty_string
-      
+
       - name: field_3
         tests:
           - not_empty_string
-      
+
   - name: test_pivot
     tests:
       - dbt_utils.equality:
           compare_model: ref('data_pivot_expected')
-        
+
+  - name: test_unpivot_original_api
+    tests:
+      - dbt_utils.equality:
+          compare_model: ref('data_unpivot_original_api_expected')
+
   - name: test_unpivot
     tests:
       - dbt_utils.equality:
           compare_model: ref('data_unpivot_expected')
-        
+
   - name: test_star
     tests:
       - dbt_utils.equality:
           compare_model: ref('data_star_expected')
-          
+
   - name: test_surrogate_key
     tests:
       - assert_equal:
           actual: actual
           expected: expected
-        
+
   - name: test_union
     tests:
       - dbt_utils.equality:
--- a/integration_tests/models/sql/test_unpivot_original_api.sql
+++ b/integration_tests/models/sql/test_unpivot_original_api.sql
@@ -1,4 +1,7 @@
 
+-- unpivot() was enhanced with 3 new parameters
+-- This test targets the original API.
+
 -- snowflake messes with these tests pretty badly since the
 -- output of the macro considers the casing of the source
 -- table columns. Using some hacks here to get this to work,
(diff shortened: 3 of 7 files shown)


drewbanin:
Hey @remigabillet - thanks for the heads up! I'll give this a look today

(cc @dwallace0723, the original author of this macro 😉)

macros/sql/unpivot.sql
```diff
@@ -23,22 +27,26 @@ Arguments:
...
 
+  {%- for col in include_cols -%}
     select
       {%- for exclude_col in exclude %}
         {{ exclude_col }},
       {%- endfor %}
-      cast('{{ col.column }}' as {{ dbt_utils.type_string() }}) as field_name,
-      cast({{ col.column }} as {{ cast_to }}) as value
+
+      cast('{{ col.column }}' as {{ dbt_utils.type_string() }}) as "{{ field_name }}",
```
drewbanin:
```suggestion
      cast('{{ col.column }}' as {{ dbt_utils.type_string() }}) as {{ field_name }},
```

I don't think we necessarily want to add quotes here. This is sensible on Redshift/pg, but causes problems on BigQuery and Snowflake. In short:
- BigQuery uses backticks (not double quotes) to quote identifiers
- Snowflake treats unquoted identifiers as upper-case, so quoting a lower-case identifier constitutes a breaking change. More info on this (bizarre and upsetting) behavior here: https://docs.getdbt.com/docs/configuring-quoting#section-snowflake

macros/sql/unpivot.sql
```diff
@@ -23,22 +27,26 @@ Arguments:
...
+      cast({{ col.column }} as {{ cast_to }}) as "{{ value_name }}"
```
drewbanin:
```suggestion
      cast({{ col.column }} as {{ cast_to }}) as {{ value_name }}
```
Same as above - let's skip the quotes and defer the quoting responsibilities to the caller. You can always do:
```
{{ unpivot(table=ref('users'), field_name='"my=quoted-value"') }}
```
if quotes are necessary in the resulting column names

integration_tests/models/sql/test_unpivot_original_api.sql
```diff
@@ -0,0 +1,32 @@
...
+    {% set exclude = ['customer_id', 'created_at'] %}
+{% endif %}
+
+select
+    customer_id,
+    created_at,
+    case
+        when '{{ target.name }}' = 'snowflake' then lower(field_name)
+        else field_name
+    end as field_name,
```
drewbanin: I think this is what caused the tests to fail. On Snowflake, this column is evaluated as `FIELD_NAME`, but the result of the `unpivot` macro is `field_name`. This is an error on Snowflake:
```
select field_name from (
  select 1 as "field_name"
)
```

integration_tests/models/sql/schema.yml
```diff
@@ -60,42 +60,42 @@ models:
...
-      
+
   - name: test_pivot
     tests:
       - dbt_utils.equality:
           compare_model: ref('data_pivot_expected')
-        
-  - name: test_unpivot
+
+  - name: test_unpivot_original_api
```
drewbanin: I think you want to add another test for the `test_unpivot` model here, right? I can see the model code and the  expected seed file, but i don't see the corresponding test. Thanks for preserving the test for the existing behavior!

README.md
```diff
@@ -337,7 +337,7 @@ This macro "un-pivots" a table from wide format to long format. Functionality is
 
 Usage:
 ```
-{{ dbt_utils.unpivot(table=ref('table_name'), cast_to='datatype', exclude=[<list of columns to exclude from unpivot>]) }}
+{{ dbt_utils.unpivot(table=ref('table_name'), cast_to='datatype', exclude=[<list of columns to exclude from unpivot>], remove=[<list of columns to remove>], field_name=<column name for field>, value_name=<column name for value>) }}
```
drewbanin: can you just enumerate the new arguments in the documentation here

macros/sql/unpivot.sql
remigabillet: got it @drewbanin

integration_tests/models/sql/schema.yml
remigabillet: duh, I forgot the other test.

README.md
remigabillet: done

integration_tests/models/sql/test_unpivot_original_api.sql
remigabillet: I think I may have fixed it but I can't get the CI to run, can you try @drewbanin?

drewbanin:
Cool! Just kicked off a build :)

I think we'll be good to merge this when the tests pass!

integration_tests/models/sql/test_unpivot.sql
```diff
@@ -11,19 +11,15 @@
...
 
-select
-    customer_id,
-    created_at,
-    case
-        when '{{ target.name }}' = 'snowflake' then lower(field_name)
-        else field_name
-    end as field_name,
-    value
+select *
```
drewbanin: looks like the test on this model is failing on Snowflake:
Screen Shot 2019-06-20 at 11 56 52 AM

Can you copy in the hack from `integration_tests/models/sql/test_unpivot_original_api.sql` here?

Maybe something like:
```sql
select
    customer_id,
    created_at,
    case
        when '{{ target.name }}' = 'snowflake' then lower(prop)
        else prop
    end as prop,
    value
```

This will just lowercase the `prop` column of the `unpivot` macro so that the seed files will match up in the equality test. Lmk if you want me to help out with that - i know it's hard to test snowflake-specific functionality without access to a snowflake database!

remigabillet: thanks for the help @drewbanin. I think I've fixed it but I still cant get the CI working. Can you re-trigger please?

drewbanin: hmm, this still doesn't look right to me (and the tests are still failing!)

can you try using the logic i sent over above in `test_unpivot.sql`?
```
select
    customer_id,
    created_at,
    case
        when '{{ target.name }}' = 'snowflake' then lower(prop)
        else prop
    end as prop,
    value
```

We need to lowercase the `prop` column to get the tests to pass on Snowflake

remigabillet:
hey @drewbanin it still doesn't work. Sorry to keep bothering you. I'm happy to iterate on it but I don't have snowflake so I need the CI to run automatically. Is there a way for me to do that?

drewbanin:
Hey @remigabillet - we disabled builds on PRs to prevent nefarious users from logging out our database credentials! We've permissioned everything accordingly, but we still review PRs before we kick off CI builds. Would love to figure out a setup (dedicated redshift+snowflake+bq instances) that would let us make this CI process more friendly for contributors though.

Not sure if you saw my comment above, but i think all you need to do to get these tests to pass is lowercase the `prop` column returned by the `unpivot` macro in `test_unpivot.sql` when `target.name == 'snowflake'`. Something like:
```
select
    customer_id,
    created_at,
    case
        when '{{ target.name }}' = 'snowflake' then lower(prop)
        else prop
    end as prop,
    value

from (
    {{ dbt_utils.unpivot(
        table=ref('data_unpivot'),
        cast_to=dbt_utils.type_string(),
        exclude=exclude,
        remove='name',
        field_name='prop',
        value_name='val'
    ) }}
) as sbq
```

Once you make that change I can kick off the CI build and get this thing merged! This is also super minor, lmk if you'd prefer for me to make that particular change.

remigabillet:
Got it re: CI.

I actually wrote the test a little differently, removing that conditional. Isn't it the same thing?

```
select
    customer_id,
    created_at,
    prop,
    val

from (
    {{ dbt_utils.unpivot(
        table=ref('data_unpivot'),
        cast_to=dbt_utils.type_string(),
        exclude=exclude,
        remove='name',
        field_name='prop',
        value_name='val'
    ) }}
) as sbq
```

remigabillet:
I'm still unclear on snowflake's casing, do I need to add conditional around the `'prop'` and `'val'` parameters?

drewbanin:
The conditional really just needs to be applied to `prop`. We're going to fetch the column names in the `data_unpivot` model from Snowflake. Snowflake will return upper-case identifiers like `'SEGMENT'` and `'STATUS'`. In order for the equality comparison against the expected output to pass, we need to lowercase these identifiers. We can actually remove the conditional and just _always_ lowercase that `prop` column, eg:
```
select
    customer_id,
    created_at,
    lower(prop) as prop,
    val
```

but I like the idea of making it explicit that Snowflake is the odd one out here.

remigabillet:
Thanks for the explanation @drewbanin. I think I get it now, Snowflake uppercases column names so I need to lowercase before I compare my test output with the expected results.
I agree that it's better to be explicit also! 👍 
I think it'll work now.

drewbanin:
cool! This last change LGTM - looks like there's just a tiny typo where you project the column as `'prop'` instead of `prop`. Fix that and we'll be good to go!

drewbanin:
thanks @remigabillet, merging!

remigabillet:
Yes !

PR #145, opened 2019-06-21
Adds width_bucket macro, tests and readme

### LINKED ISSUE #138: width_bucket
might it be possible to implement this:

https://docs.snowflake.net/manuals/sql-reference/functions/width_bucket.html

in a cross-database way?? would be super-cool :D

clausherther:
Closes #138 

Adds:
- `width_bucket` macro
- integration test
- updated README


drewbanin:
Thanks @clausherther - just kicked off the tests 🤞

macros/cross_db_utils/width_bucket.sql
```diff
@@ -0,0 +1,59 @@
...
+{% endmacro %}
+
+
+{% macro default__width_bucket(expr, min_value, max_value, num_buckets) -%}
+
+    {% set bin_size -%}
+    (( {{ max_value }} - {{ min_value }} ) / {{ num_buckets }} )
+    {%- endset %}
+    (
+        -- to break ties when the amount is eaxtly at the bucket egde
```
clrcrl:
```suggestion
        -- to break ties when the amount is exactly at the bucket edge
```

macros/cross_db_utils/width_bucket.sql
```diff
@@ -0,0 +1,59 @@
...
+        case
+            when 
+                mod(
+                    {{ dbt_utils.safe_cast(expr, 'decimal(28,6)') }},
```
clrcrl: So, unfortunately `decimal(28, 6)` is not a valid data type on BigQuery. Instead:
```suggestion
                   safe_cast(bin_size as numeric) }}
```
You might have to create a separate `bigquery__width_bucket` macro to implement this logic 😬

macros/cross_db_utils/width_bucket.sql
```diff
@@ -0,0 +1,59 @@
...
+                    {{ dbt_utils.safe_cast(bin_size, 'decimal(28,6)') }}
```
clrcrl:
```suggestion
                    safe_cast({{ bin_size }} as numeric)
```

macros/cross_db_utils/width_bucket.sql
```diff
@@ -0,0 +1,59 @@
...
+
+{% macro redshift__width_bucket(expr, min_value, max_value, num_buckets) -%}
+
+    {% set bin_size -%}
+    (( {{ max_value }} - {{ min_value }} ) / {{ num_buckets }} )
+    {%- endset %}
+    (
+        -- to break ties when the amount is eaxtly at the bucket egde
+        case
+            when
```
clrcrl: AFAIK, `mod` works on Redshift. Is there a reason you aren't using it here?

clrcrl: Oh I got it -- because `mod` only works on two integers. Yeesh.
```
Function "mod(numeric,numeric)" not supported.
```
Even though the [docs](https://docs.aws.amazon.com/redshift/latest/dg/r_MOD.html) indicate it should work:
> The first input parameter is an INTEGER, SMALLINT, BIGINT, or DECIMAL number. If either parameter is a DECIMAL type, the other parameter must also be a DECIMAL type.

Yikes!

PR #147, opened 2019-06-28
support extra parameter 'condition' on 'expression_is_true' macro

### LINKED ISSUE #146: Extend 'expression_is_true' test to support subset asertions
`expression_is_true` test is really useful, but sometimes I am missing the option of using this test to assert on a subset of the rows of the model. For example, I would like to be able to assert that `price` is higher than 5K  if `product_type` is a car. For these reason I ended up creating a custom test for this like the one below, thought it feels like this kind of functionality could be in the `expression_is_true` test already.

```sql
{#
  Checks that an expression is true on those rows that meet a previous condition
  This is different from the 'expression_is_true' test from dbt-utils in that can
  limit the input dataset to be tested by using a 'condition'

  Arguments:
    - model: the name of the model to be tested
    - expression: the expression to be asserted in the data
    - condition: filters to a subset of the model rows
#}
{% macro test_true_on_condition(model, expression, condition) %}

with meet_condition as (
  select * from {{ model }} where {{ condition }}
)
select count(*)
from meet_condition
where not({{ expression }})

{% endmacro %}
```

Proposal: extend the `expression_is_true` test with an optional parameter `condition` that takes a `none` default value. When nothing is passed in the `condition` argument the expression is tested against the whole model. When a `condition` is passed the the model is subset by the condition prior to asserting the expression.

I'd be happy to send a PR this fits.

dcereijodo:
An optional parameter 'condition' can be passed to the 
'expression_is_true' macro to assert the expression for all records 
which meet a condition. Closes #146

**Checklist**
- [x] My commits are related to the pull request and look clean.
- [x] My SQL follows the [Fishtown Analytics style guide](https://github.com/fishtown-analytics/corp/blob/master/dbt_coding_conventions.md).
- [X] I have added appropriate tests and documentation to any new models.
- [x] I have updated the README file.

### DIFF
--- a/macros/schema_tests/expression_is_true.sql
+++ b/macros/schema_tests/expression_is_true.sql
@@ -1,7 +1,6 @@
-{% macro test_expression_is_true(model) %}
+{% macro test_expression_is_true(model, condition='true') %}
 
 {% set expression = kwargs.get('expression', kwargs.get('arg')) %}
-{% set condition = kwargs.get('condition', 'true') %}
 
 with meet_condition as (
 
--- a/macros/schema_tests/expression_is_true.sql
+++ b/macros/schema_tests/expression_is_true.sql
@@ -1,12 +1,18 @@
 {% macro test_expression_is_true(model) %}
 
 {% set expression = kwargs.get('expression', kwargs.get('arg')) %}
+{% set condition = kwargs.get('condition', 'true') %}
 
-with validation_errors as (
+with meet_condition as (
+
+    select * from {{ model }} where {{ condition }}
+
+),
+validation_errors as (
 
     select
         *
-    from {{model}}
+    from meet_condition
     where not({{expression}})
 
 )
--- a/README.md
+++ b/README.md
@@ -128,6 +128,23 @@ models:
 
 ```
 
+The macro accepts an optional parameter `condition` that allows for asserting
+the `expression` on a subset of all records.
+
+Usage:
+```yaml
+version: 2
+
+models:
+  - name: model_name
+    tests:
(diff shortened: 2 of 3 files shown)


dcereijodo:
CI failed :( 
Looks like configuration stuff though: `Env var required but not provided: 'CI_REDSHIFT_DBT_HOST'`

macros/schema_tests/expression_is_true.sql
```diff
@@ -1,12 +1,18 @@
 {% macro test_expression_is_true(model) %}
 
 {% set expression = kwargs.get('expression', kwargs.get('arg')) %}
+{% set condition = kwargs.get('condition', 'true') %}
```
drewbanin: I think `condition` can be supplied as a proper keyword argument here, with the default set to `true`. We use this `kwargs.get` pattern to support both the model-level and column-level test specification syntax, but in the general case, I think it makes to keep arguments in the method signature where possible.

dcereijodo: I was actually a bit confused by the use of `kwargs` here, but I thought there should be a reason for it and kept it the same. I agree a normal argument is better. I'll send the changes with your suggestions. Thanks!

drewbanin:
Cool, thanks for making that change! This will be good to merge when the tests pass.

Side note: love your PR description.... we're going to add something like that as the default for PRs to this repo :)

dcereijodo:
Thanks :)

drewbanin:
Merging this! Will cut a new release with this change on Monday :)

🎉

PR #154, opened 2019-07-27
Fix/get column values

### LINKED ISSUE #153: [Issue] get_column_values now only accepts a relation
## Steps to reproduce
1. Install dbt-utils v0.2.1
2. Call `get_column_values` with a string table reference
```
{{ dbt_utils.get_column_values(
    table="raw_jaffle_shop.customers",
    column='first_name',
    max_records=50
) }}
```
## Expected results
* Macro compiles correctly (this was the behaviour <0.2.1)

## Actual results
```
Running with dbt=0.14.0
Found 10 models, 28 tests, 0 snapshots, 0 analyses, 249 macros, 0 operations, 0 seed files, 3 sources

16:53:00 | Concurrency: 1 threads (target='dev')
16:53:00 |
Encountered an error:
Runtime Error
  Compilation Error in macro get_column_values (macros/sql/get_column_values.sql)
    'str object' has no attribute 'database'
```

## Commentary
Prior to v0.2.1, you could pass `get_column_values` a table like this `my_schema.my_table`.

However #152 now means you need to pass the macro a `relation` instead. All of these are valid:
```
{{ dbt_utils.get_column_values(
    table=ref('stg_jaffle_shop__customers'),
    column='first_name',
    max_records=50
) }}
```
```
{{ dbt_utils.get_column_values(
    table=source('raw_jaffle_shop', 'customers'),
    column='first_name',
    max_records=50
) }}
```
```
{%- set source_relation = adapter.get_relation(
      database=target.database,
      schema='raw_jaffle_shop',
      identifier="customers") -%}

{{ dbt_utils.get_column_values(
    table=source_relation,
    column='first_name',
    max_records=50
) }}

```

I'm totally 100% behind this change, but there needs to be some cleaning up here to communicate it.
- [ ] The README.md still says shows an example of passing a hardcoded table reference in the [pivot example](https://github.com/fishtown-analytics/dbt-utils/blob/master/README.md#pivot-source)
- [ ] Techhhhhnically we should replace the `table` argument name with `relation`, but that's likely going to break things (on the fence about whether we should fix it)
- [ ] There's also a little bit of redundant code in `get_column_values` which would be good to clean up:

```
    {%- set target_relation = adapter.get_relation(database=table.database,
                                          schema=table.schema,
                                         identifier=table.identifier) -%}
```
can be replaced with
```
    {%- set target_relation = table -%}
```

clrcrl:
Fixes #153 
I've left the argument name as `table` – this terminology is used consistently throughout utils (e.g. `get_tables_by_prefix`, `union_tables`, `unpivot`), so if we want to fix that, we should do it as a separate PR to fix all of them.

### DIFF
--- a/macros/sql/get_column_values.sql
+++ b/macros/sql/get_column_values.sql
@@ -18,9 +18,7 @@ Returns:
     {% endif %}
 {#--  #}
 
-    {%- set target_relation = adapter.get_relation(database=table.database,
-                                          schema=table.schema,
-                                         identifier=table.identifier) -%}
+    {%- set target_relation = table -%}
 
     {%- call statement('get_column_values', fetch_result=true) %}
 
--- a/macros/sql/get_column_values.sql
+++ b/macros/sql/get_column_values.sql
@@ -18,7 +18,9 @@ Returns:
     {% endif %}
 {#--  #}
 
-    {%- set target_relation = table -%}
+    {%- set target_relation = adapter.get_relation(database=table.database,
+                                          schema=table.schema,
+                                         identifier=table.identifier) -%}
 
     {%- call statement('get_column_values', fetch_result=true) %}
 
--- a/README.md
+++ b/README.md
@@ -343,7 +343,14 @@ This macro "un-pivots" a table from wide format to long format. Functionality is
 
 Usage:
 ```
-{{ dbt_utils.unpivot(table=ref('table_name'), cast_to='datatype', exclude=[<list of columns to exclude from unpivot>], remove=[<list of columns to remove>], field_name=<column name for field>, value_name=<column name for value>) }}
+{{ dbt_utils.unpivot(
+  table=ref('table_name'),
+  cast_to='datatype',
+  exclude=[<list of columns to exclude from unpivot>],
+  remove=[<list of columns to remove>],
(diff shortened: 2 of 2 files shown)


macros/sql/get_column_values.sql
```diff
@@ -18,9 +18,7 @@ Returns:
     {% endif %}
 {#--  #}
 
-    {%- set target_relation = adapter.get_relation(database=table.database,
-                                          schema=table.schema,
-                                         identifier=table.identifier) -%}
+    {%- set target_relation = table -%}
```
drewbanin: Calling `get_relation` here serves a purpose - it checks to see if the target relation actually exists in the database. If it doesn't exist, then `target_relation` will be `none` and this macro will return the `default`.

This can happen if you're doing `dbt compile` on a fresh project, for instance. Previously, the query in this macro would fail -- dbt would try to select from a nonexistent model.

I think ideally, we'd have a different adapter method which checks that a given Relation exists. Today though, we need to pull apart the Relation to pass it into `get_relation`, which is pretty weird!

clrcrl: Ah got it! I'll revert this commit then

PR #156, opened 2019-08-27
Fixes for query transaction flow - where to begin and commit

davehowell:
Remove redundant `begin` inside call statement
Use `adapter.commit()` instead of commit inside a call statement
Move `get_period_sql` string prep outside the DB `call statement` block
Commit each iteration of the loop

When comparing this materialization with the core ones like `table` and `incremental` there's a few things I noticed should be changed for consistency as well as when `begin` and `commit` are actually handled in the control flow.

--------------------------------------

On ln85
```
-- `begin` happens here, so `commit` after it to finish the transaction
  {{run_hooks(pre_hooks, inside_transaction=True)}}
  {% call statement() -%}
    begin; -- make extra sure we've closed out the transaction
    commit;
  {%- endcall %}
```
• It is directly calling the sql begin and commit within a `call statement()` block whereas the core materializations use adapter.commit(), which appears to be a safer call; it checks if there is an open transaction. On deeper inspection, `call statement()` opens a transaction by default - `auto_begin=True`,  so calling a statement with "commit" is self-committing. Similarly if you `call statement()` with `begin; end;` then it opens two transactions and commits one of them. I have searched and can't find any information on Redshift and nested transactions, it is possible that they aren't nested, maybe a commit or rollback will apply to all previous transactions.

• It calls `begin` inside run_hooks(pre_hooks, inside_transaction=True) and then calls `begin` again, and then immediately commit which seems like it is starting two transactions and ending the inner one. I get the idea that since this materialization's body is a big loop of queries, then running the entire loop along with hooks in one big transaction is a bad idea. It would be a long running transaction, and successful iterations of the loop would be rolled back if any subsequent failed, not ideal. On the other hand, running the pre and post hooks on every iteration would also be overkill. So the sensible decision was to commit the transactional pre & post hooks independently.

I dug in to the run_hooks code:

• If `inside_transaction=True`, and there is no current open transaction, it will `begin` a new one,  but commit has to be handled separately.

•  If  `inside_transaction=False`, it will call commit; prior to looping `call statement(auto_begin=False)` for each hook sql statement.  

Therefore the call to `begin` before these post_hook calls and the call to `commit` between these post_hook calls are redundant  - but not harmful:
```
{% call statement() -%}
    begin;
 {%- endcall %}

{{run_hooks(post_hooks, inside_transaction=True)}}

  {% call statement() -%}
    commit;
  {%- endcall %}

  {{run_hooks(post_hooks, inside_transaction=False)}}
```

• Prior to the big loop it has this comment:
```
-- commit each period as a separate transaction
  {% for i in range(num_periods) -%}
...
```

I don't see anything that explicitly begins or commits each period iteration of the loop, although I see that `call statement()` has a default `auto_begin=True` which implies that it will attempt to begin or use an existing open transaction _on every loop_. That does suggest that each iteration should call  {{ adapter.commit() }} to create a save point. I have added this.

On a separate note, I think this run_hooks macro should also use the safer call to {{ adapter.commit() }} https://github.com/fishtown-analytics/dbt/blob/399b33822a85493a8790df787eed59e951aecf6c/core/dbt/include/global_project/macros/materializations/helpers.sql#L5
That's in the core project so I'll leave that to your judgement.

### DIFF
--- a/macros/materializations/insert_by_period_materialization.sql
+++ b/macros/materializations/insert_by_period_materialization.sql
@@ -84,10 +84,7 @@
 
   -- `begin` happens here, so `commit` after it to finish the transaction
   {{run_hooks(pre_hooks, inside_transaction=True)}}
-  {% call statement() -%}
-    begin; -- make extra sure we've closed out the transaction
-    commit;
-  {%- endcall %}
+  {{ adapter.commit() }}
 
   -- build model
   {% if force_create or old_relation is none -%}
@@ -112,7 +109,6 @@
   {%- set target_cols_csv = target_columns | map(attribute='quoted') | join(', ') -%}
   {%- set loop_vars = {'sum_rows_inserted': 0} -%}
 
-  -- commit each period as a separate transaction
   {% for i in range(num_periods) -%}
     {%- set msg = "Running for " ~ period ~ " " ~ (i + 1) ~ " of " ~ (num_periods) -%}
     {{ dbt_utils.log_info(msg) }}
@@ -120,19 +116,22 @@
     {%- set tmp_identifier = model['name'] ~ '__dbt_incremental_period' ~ i ~ '_tmp' -%}
     {%- set tmp_relation = api.Relation.create(identifier=tmp_identifier,
                                                schema=schema, type='table') -%}
+    {% set tmp_table_sql = dbt_utils.get_period_sql(target_cols_csv,
+                                                    sql,
+                                                    timestamp_field,
+                                                    period,
+                                                    start_timestamp,
+                                                    stop_timestamp,
+                                                    i) %}
+
+    -- statement() will begin or re-use an existing transaction
     {% call statement() -%}
-      {% set tmp_table_sql = dbt_utils.get_period_sql(target_cols_csv,
-                                                       sql,
-                                                       timestamp_field,
-                                                       period,
-                                                       start_timestamp,
-                                                       stop_timestamp,
-                                                       i) %}
       {{dbt.create_table_as(True, tmp_relation, tmp_table_sql)}}
     {%- endcall %}
 
     {{adapter.expand_target_column_types(from_relation=tmp_relation,
                                          to_relation=target_relation)}}
+
     {%- set name = 'main-' ~ i -%}
     {% call statement(name, fetch_result=True) -%}
       insert into {{target_relation}} ({{target_cols_csv}})
@@ -142,6 +141,7 @@
           from {{tmp_relation.include(schema=False)}}
       );
     {%- endcall %}
+
     {%- set rows_inserted = (load_result('main-' ~ i)['status'].split(" "))[2] | int -%}
     {%- set sum_rows_inserted = loop_vars['sum_rows_inserted'] + rows_inserted -%}
     {%- if loop_vars.update({'sum_rows_inserted': sum_rows_inserted}) %} {% endif -%}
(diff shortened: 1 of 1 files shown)


davehowell:
The circleci failure `Env var required but not provided: 'CI_REDSHIFT_DBT_HOST'` seems like missing configuration on the CI.

macros/materializations/insert_by_period_materialization.sql
```diff
@@ -84,10 +84,7 @@
 
   -- `begin` happens here, so `commit` after it to finish the transaction
   {{run_hooks(pre_hooks, inside_transaction=True)}}
-  {% call statement() -%}
-    begin; -- make extra sure we've closed out the transaction
-    commit;
-  {%- endcall %}
+  {{ adapter.commit() }}
```
drewbanin: this materialization previously used 
```
begin;
commit;
```
to make sure that the following code does _not_ run inside of a transaction. We can't just call `commit` here directly, as the database will return an error if a transaction isn't currently open. I think the previous approach was suboptimal, but I think this change lead to the test failure that you can see in the CircleCI build.

drewbanin: It's also worth saying: `adapter.commit()` and `commit;` do slightly different things. Calling `adapter.commit()` will tell dbt that the active connection has a transaction open, whereas running `commit;` explicitly will not. I'm very open to using `adapter.commit()` here, but I do think there's some other work required to make this work correctly in practice.

davehowell: Interesting, I thought that `run_hooks` with `inside_transaction=True` would either begin a transaction or continue an existing one, perhaps it commits or otherwise ends the transaction internally. If there is no transaction at this point then the `{{ adapter.commit() }}` might be redundant.

I've looked a bit more and see now that Redshift and postgres before v11 [don't support nested transactions](https://stackoverflow.com/a/25428060/1335793), and Redshift has no support for savepoint or subtransactions either. I've misunderstood the behaviour. So you could `begin` a bunch of times and only need to `commit` or `rollback` once. The previous code here is therefore not harmless at worst and safe at best - doing what the comment said. I'll revert it.

davehowell:
> Do you think that each insert should be made inside of a transaction? Or do you think we're better off just letting the database run each of these queries independently outside the scope of a transaction?

The one thing to avoid is a long running transaction. I do think the queries should be outside the transactions and only the run_hook for pre and post hooks that are supposed to be in a transaction should be in transactions. 

I am under the impression that `call statement` will begin or continue a transaction due to this default arg `auto_begin=True`, so perhaps this should be called with `auto_begin=False`

drewbanin:
I do think you're right about that @davehowell! I was surprised to see in those code that `auto_begin` was _not_ explicitly provided in this materialization!

Totally agree that the hooks should be transactional - we just need to clear out the transaction before the "building" section of the code here

drewbanin:
I believe we discussed this on Slack (?) but what was the actual change you were interested in making here? I'm all for cleaning up this code, just want to make sure I understand what your goals are here

davehowell:
My original goal was an issue with this materialization taking a really long time (>15mins) to run one iteration, where I could run the query for one iteration myself e.g. via psql or another client, and it was fast (< 30s).  I ruled out caching and other effects. I thought perhaps there were issues with the transactions being held open a long time then went through cleaning up the transaction flow.  I believe my issues were WLM queue related but thought the clean up here was still worthwhile. 

It works for me on a model in production, but I realise that it is on v 0.13.1 so I will do the upgrade and test it and will likely find the same error that Circle CI is throwing. After I do the upgrade I'll make the changes discussed above.

clrcrl:
Closing since this is stale

PR #161, opened 2019-09-17
Add a new schema test: relationships_where

MartinGuindon:
Hello dbt-utils maintainers!

I'm submitting this PR to add a new schema test that has proven to be extremely useful. 

Note that I'm _not_ the original author of the macro. It was shared on Slack last May and I had preserved a copy of it in my project. I unfortunately didn't write down who was the original contributor... so my apologies to him/her.

The need for this macro came up a few times on Slack over these past few months so I thought I'd issue a PR anyway.

To this unknown person who shared it originally on Slack:  thank you!


README.md
```diff
@@ -190,6 +190,26 @@ models:
...
+This test validates the referential integrity between two tables (same as the core relationships schema test) with an added predicate to filter out some rows from the test. This is useful to exclude records such as test entities, rows created in the last X minutes/hours to account for temporary gaps due to ETL limitations, etc.
+
+Usage:
+```yaml
+version: 2
+
+models:
+  - name: model_name
+    columns:
+      - name: from_column
```
drewbanin: cool test! I'd change `from_column` to `id` here, then remove the `from:` argument to the `relationships_where` test. I think that will reflect a more typical usage of a test like this.

It's really great that you included the `from` argument here as well -- that makes it possible to define a test that _does not_ apply to a particular column, eg. a test on an expression

macros/schema_tests/relationships_where.sql
```diff
@@ -0,0 +1,48 @@
+{% macro test_relationships_where(model, to, field) %}
+
+{% set column_name = kwargs.get('column_name', kwargs.get('from')) %}
+{% set from_condition = kwargs.get('from_condition', True) %}
```
drewbanin: this is pretty minor, but I'd use `"true"` (a string) as the default value for `from_condition` and `to_condition` here

MartinGuindon: Done.

README.md
MartinGuindon: Good point. Done.

drewbanin:
hey @MartinGuindon can you pull `fishtown-analytics/dbt-utils` master into your branch? I think if you do that, then push your changes, it will fix these tests!

MartinGuindon:
Thanks @drewbanin. Done that, and fixed a small mistake at the same time.
Postgres tests are now successful but Redshift tests are failing:
```Compilation Error
  Env var required but not provided: 'CI_REDSHIFT_DBT_HOST'
Exited with code 2```

File: integration_tests/models/schema_tests/schema.yml [hunk 3]
```diff
@@ -32,11 +32,21 @@ models:
...
+
+  - name: data_test_relationships_where_table_2
+    columns:
+      - name: id
+        tests:
+          - dbt_utils.relationships_where:
+              from: id
+              to: ref('data_test_relationships_where_table_1')
+              field: id
+              from_condition: id <> '4'
```
drewbanin: I think this test is failing on BigQuery because the `id` column is being loaded as an integer. Can you change `'4'` to `4` here? I think we should be good to go after that!

MartinGuindon: Oops. Bad habit from Snowflake where `4 = '4'`... 
Fixed!

macros/schema_tests/relationships_where.sql
tconbeer: Great feedback. Thanks, Drew

PR #162, opened 2019-09-19
Add get_column_names_from_sql macro

clrcrl:
Inspired by a Slack thread, didn't want to lose it.

I've wanted this before, but have always worked around it by creating things as views upstream.

TBH I'm not sure of the utility of it unless we work it into `union` and `star` so that both work with upstream ephemeral models.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ This macro is useful when you need to iterate through columns in a CTE (includin
 {% set query_sql %}
 select * from {{ ref('my_ephemeral_model') }}
 {% endset %}
-{% set column_names=get_columns_in_relation(query_sql) %}
+{% set column_names=get_column_names_from_sql(query_sql) %}
 
 select
 {% for column_name in column_names %}
--- a/README.md
+++ b/README.md
@@ -230,6 +230,24 @@ Usage:
 ...
 ```
 
+#### get_column_names_from_sql ([source](macros/sql/get_column_names_from_sql.sql))
+This macro returns a list of column names (_not_ `Column`[objects](https://docs.getdbt.com/docs/api-variable#section-column)) for the results of a provided query.
+
+This macro is useful when you need to iterate through columns in a CTE (including an ephemeral model) since the information_schema cannot be queried. If you need to find the columns in a [Relation](https://docs.getdbt.com/docs/api-variable#section-relation) (generally a table or view), you should use [get_columns_in_relation](https://docs.getdbt.com/docs/adapter#section-get_columns_in_relation) instead.
+
+```sql
+{% set query_sql %}
+select * from {{ ref('my_ephemeral_model') }}
+{% endset %}
+{% set column_names=get_columns_in_relation(query_sql) %}
+
+select
+{% for column_name in column_names %}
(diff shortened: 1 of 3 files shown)


bastienboutonnet:
I tried this on my end, and the expected behaviour of `if execute` seems a bit odd.
When I run compile, the sql passed to `run_query()` gets ran agains the database. Could it be that `execute` evaluates as `True` even during `dbt compile`? 

This might also be something wrong in my settings (also I am on Snowflake in case there is an exception) and I'm not really able to traceback the logic that sets that flag with the amount of time I have these days but I thought it would be worth looking into if you guys are going to merge this.

clrcrl:
@bastienboutonnet -- `execute` is a bit of a fickle beast. AFAIK (and Drew can correct me if I'm wrong), when you `dbt compile`, dbt:
1. Parses the project --> `execute == False` / no queries are run. If we tried to access `results.column_names` at this stage, we'd get back an error like this:
```
Encountered an error:
Compilation Error in macro get_column_names_from_sql (macros/sql/get_column_names_from_sql.sql)
  'None' has no attribute 'table'
```
This is why we have to nest some of the code in `if execute`.
2. Compiles the project --> `execute == True`. This means that even on `compile`, the query gets executed, so the behaviour you're seeing is expected :)

bastienboutonnet:
Ok. So for a use case when you're trying to get cols from a CTE which will not have been generated then the compile fails. In my use case it breaks the run as it will throw things like `<CTE_NAME> does not exist`

clrcrl:
LOL! Turns out this already exists [here](https://github.com/fishtown-analytics/dbt/blob/dev/louisa-may-alcott/core/dbt/include/global_project/macros/adapters/common.sql#L31). So you can just access it using `adapter.get_columns_in_query`. Silly me :)

bastienboutonnet:
@clrcrl This macro doesn't seem to work for me sadly.
I'm using it as so:
```sql
with test_cte as (
    select
        *
    from dwh.data_warehouse.fact_booking
)
select
{{get_columns_in_cte('select * from test_cte')}}
```

clrcrl:
Hey @bastienboutonnet 👋 

I got it to work like so:
```
{% if execute %}
    {%- set query_to_go -%}

    with test_cte as (
        select
            *
        from raw_jaffle_shop.customers
    )
    select * from test_cte
    {%- endset -%}

    {% set cols=get_columns_in_query(query_to_go) %}
{% else %}

{% set cols=[] %}

{% endif %}

{{ log(cols, info=True) }}
{{ cols }}

```

The output looks like this:
```
$ dbt compile
Running with dbt=0.14.2
Found 9 models, 28 tests, 0 snapshots, 1 analyse, 138 macros, 0 operations, 0 seed files, 3 sources

11:32:51 | Concurrency: 1 threads (target='dev')
11:32:51 |
['id', 'first_name', 'last_name', 'email', '_elt_updated_at']
11:32:52 | Done.
```

And the compiled query looks like this:
```
    ['id', 'first_name', 'last_name', 'email', '_elt_updated_at']
```

bastienboutonnet:
@clrcrl Amazing that totally works! This is going to be suuuch a nice thing to have. Thanks again!

PR #164, opened 2019-09-25
Show correct usage information for the dateadd macro

### LINKED ISSUE #163: Update documentation for the datedifff macro: add another set of quotes
Hey everyone,
it is my understanding from using `dbt_utils.dateadd` this morning on a Redshift cluster and failing that the macro is not supported for Redshfit.

Error
```bash
Completed with 1 errors:

Database Error in model labelling_set (models/marts/ml-entities/labelling_set.sql)
  function date_add("unknown", integer, integer) does not exist
  HINT:  No function matches the given name and argument types. You may need to add explicit type casts.
  compiled SQL at target/compiled/of_models/marts/ml-entities/labelling_set.sql
```

Code https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/cross_db_utils/dateadd.sql

I'm willing to open a PR for this and dig deeper into the logic of `adapter_macro` etc. if that sounds like something that could be useful (for my own usecase I just wrote manually the `date_add` redshift statement).

louisguitton:
Fix #163

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ This macro adds a time/day interval to the supplied date/timestamp. Note: The `d
 
 Usage:
 ```
-{{ dbt_utils.dateadd(datepart='day', interval=1, from_date_or_timestamp='2017-01-01') }}
+{{ dbt_utils.dateadd(datepart='day', interval=1, from_date_or_timestamp="'2017-01-01'") }}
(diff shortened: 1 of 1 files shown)


drewbanin:
Thanks @louisguitton! Merging!!

PR #166, opened 2019-10-09
Feature/test mutually exclusive ranges

clrcrl:
Test failing because of [this issue](https://github.com/fishtown-analytics/dbt/issues/1817). Will wait for that fix rather than working around the issue.


README.md
```diff
@@ -231,6 +231,49 @@ models:
...
+version: 2
+
+models:
+  # test that age ranges do not overlap
+  - name: age_brackets
+    tests:
+      - dbt_utils.mutually_exclusive_ranges:
+          lower_bound_column: min_age
+          upper_bound_column: max_age
+          collectively_exhaustive: true
```
clrcrl: Is collectively exhaustive the right word here?

jthandy:
@drewbanin I'm not going to be able to look at this for like another week. Any chance you're able to help out here? Sorry, I know I'm useless right now.

README.md
```diff
@@ -231,6 +231,49 @@ models:
...
+* `partition_by` (optional): If a subset of records should be mutually exclusive
+(e.g. all periods for a single subscription_id are mutually exclusive), use this
+argument to indicate which column to partition by. `default=none`
+* `collectively_exhaustive` (optional): True if the upper bound of one row should
+be the lower bound of the next row. `default=false`
+
+**Understanding `collectively_exhaustive`:**
+Take the example of subscriptions, where a customer can only have one subscription
+at a time. This customer may have periods where they _do not_ have any active
+subscription, in which case the ranges would not overlap, but they might.
```
drewbanin: > but they might.

i think this could be reworded for clarity

README.md
drewbanin: in the docs below, you write:
> (i.e. there are gaps)

What do you think about `allow_gaps`? That would be the converse of what this flag currently does, but I think it would be pretty clear!

macros/schema_tests/test_mutually_exclusive_ranges.sql
```diff
@@ -0,0 +1,57 @@
...
+    select * from calc
+    where (
+        -- lower_bound should always be before upper_bound
+        not(lower_bound < upper_bound)
+
+        -- upper_bound should always be before/the same as the next lower_bound
+        -- (coalesce it to handle null cases for the last record)
+        or not(
+            coalesce(
+                upper_bound {{ collectively_exhaustive_operator }} next_lower_bound,
```
drewbanin: what happens if `next_lower_bound` is `null`? Should that be an error here, or should users implement not_null tests on these columns too?

macros/schema_tests/test_mutually_exclusive_ranges.sql
```diff
@@ -0,0 +1,57 @@
...
+            {{ partition_clause }}
+            order by {{ lower_bound_column }}
+        ) as next_lower_bound
+
+    from {{ model }}
+
+),
+
+validation_errors as (
+    -- we want to return records where our assumptions are NOT true, so we'll use
```
drewbanin: is this correct? It seems like you did the opposite of this in the code below! I think it would make sense to put one big `not` around the entire predicate

macros/schema_tests/test_mutually_exclusive_ranges.sql
```diff
@@ -0,0 +1,57 @@
...
+    -- the `not` function here so we can write our assumptions nore cleanly
+    select * from calc
+    where (
+        -- lower_bound should always be before upper_bound
+        not(lower_bound < upper_bound)
```
drewbanin: if `lower_bound` or `upper_bound` are `null` here, then this will evaluate to `not(null)` which will be `null`, and the row will be excluded from the `select` results. This is sensible for the first and last record in the partition, but outside of these specific places, it could represent an error that may be worth catching in this particular test.

clrcrl: Fixed!

macros/schema_tests/test_mutually_exclusive_ranges.sql
clrcrl: Rewrote this to make it cleaner

clrcrl:
A few other things:
1. Is `mutually_exclusive` the right term here? Is it known enough? Should we call it `non_overlapping_ranges` instead?
2. I thought about getting extremely persnickety and implementing three operators, i.e.:
```
gaps:not_allowed --> lower_bound  = previous_upper_bound
gaps:allowed     --> lower_bound <= previous_upper_bound
gaps:required    --> lower_bound <  previous_upper_bound
```
example of `gaps:not_allowed`:
| lower_bound | upper_bound | 
|-------------|-------------| 
| 0           | 1           | 
| 1           | 2           | 
| 2           | 3           | 


example of `gaps:allowed`
| lower_bound | upper_bound | 
|-------------|-------------| 
| 0           | 1           | 
| 2           | 3           | 
| 3           | 4           | 

example of `gaps:required`
| lower_bound | upper_bound | 
|-------------|-------------| 
| 0           | 1           | 
| 1           | 2           | 
| 3           | 4           | 


There's probably a better way to express that, not sure if it's worth implementing. I can imagine date ranges where you actually want things to not overlap so you don't have a fan out in any downstream joins, but 🤷‍♀

Issue #120, opened 2019-02-15
`date_spine` no longer works on BigQuery


dolfth:
Using date_spine this way on BigQuery

`    {{ dbt_utils.date_spine(
        "day",
        "parse_date('%F', '2000-01-01')",
        "date_add(current_datetime, interval 53 week)"
       )
    }}
`
errors with the following message:
`No matching signature for operator <= for argument types: DATETIME, DATE. Supported signatures: ANY <= ANY`
It seems that the errors results from the comparison in [this line](https://github.com/fishtown-analytics/dbt-utils/blob/ce8daeb6e01490f2833b8cd6cb06656b891f667b/macros/datetime/date_spine.sql#L63) that refers to dateadd. Looking at that file it seems that [this commit to dateadd](https://github.com/fishtown-analytics/dbt-utils/commit/eefa8441ac484f47e332c22f316379038f72af41) changed the type of the output of the macro to datetime in the case of BQ. Reverting to the older version solves the problem.

drewbanin:
hey @dolfth - thanks for the report. Just wanted to acknowledge that I saw this - will take a look this week!

dolfth:
I am currently working with [this version of date_add](https://github.com/fishtown-analytics/dbt-utils/blob/29050eec54ac044e96c4c0151b24d8cf4386fbac/macros/cross_db_utils/dateadd.sql) without any problems

drewbanin:
thanks for the update!

clausherther:
Hi, just wanted to check on this. Any plans to merge @dolfth's version of date_add to make the date_spine work?

clausherther:
Ah, looks like I have to call the date_spine macro like this
```
{{ dbt_utils.date_spine(
        datepart="day",
        start_date="cast('" ~ start_date ~ "' as datetime)",
        end_date="cast('" ~ end_date ~ "' as datetime)",
       )
    }}
```
since the `day_date column` is of type `datetime` instead of `date`.

jtcohen6:
Hey folks! I know this issue is from a while ago. I was able to get `date_spine` working on BQ with a _very_ slight tweak to @dolfth's original code:
```sql
{{ dbt_utils.date_spine( "day", "parse_date('%F', '2000-01-01')", "datetime_add(current_datetime, interval 53 week)" ) }}
```
[...]

PR #167, opened 2019-10-14
Fixes date_spine for BigQuery

clausherther:
This addresses the datatype issue when calling `date_spine` from BigQuery. See #120 
I've also added `datetime` to `dbt_utils.datatypes()` in case some (future) platforms have requirements or names for this.

You still have to call `date_spine` with parameters that _can_ be cast to a `datetime`, so either dates, strings or datetimes. 

So, this (from the README) still works:
```
{{ dbt_utils.date_spine(
    datepart="minute",
    start_date="to_date('01/01/2016', 'mm/dd/yyyy')",
    end_date="dateadd(week, 1, current_date)"
   )
}}
```

But I don't think something like _this_ would work:
```
{{ dbt_utils.date_spine(
    datepart="minute",
    start_date="2016-01-01",
    end_date="2022-12-31"
   )
}}
```
since those start/end dates would need be wrapped in quotes (`'`) to be parsed as strings. Not sure this is something we need to address.

### DIFF
--- a/macros/cross_db_utils/datatypes.sql
+++ b/macros/cross_db_utils/datatypes.sql
@@ -79,3 +79,30 @@
 {% macro bigquery__type_int() %}
     int64
 {% endmacro %}
+
+{# datetime  -------------------------------------------------     #}
+
+{%- macro type_datetime() -%}
+  {{ adapter_macro('dbt_utils.type_datetime') }}
+{%- endmacro -%}
+
+{%- macro default__type_datetime() -%}
+    datetime
+{%- endmacro -%}
+
+{%- macro bigquery__type_datetime() -%}
+    datetime
+{%- endmacro -%}
+
+{%- macro redshift__type_datetime() -%}
+    datetime
+{%- endmacro -%}
+
+{%- macro postgres__type_datetime() -%}
+    datetime
+{%- endmacro -%}
+
+{%- macro snowflake__type_datetime() -%}
+    datetime
+{%- endmacro -%}
+
--- a/macros/cross_db_utils/datatypes.sql
+++ b/macros/cross_db_utils/datatypes.sql
@@ -89,20 +89,3 @@
 {%- macro default__type_datetime() -%}
     datetime
 {%- endmacro -%}
-
-{%- macro bigquery__type_datetime() -%}
-    datetime
-{%- endmacro -%}
-
-{%- macro redshift__type_datetime() -%}
-    datetime
-{%- endmacro -%}
-
-{%- macro postgres__type_datetime() -%}
-    datetime
-{%- endmacro -%}
-
-{%- macro snowflake__type_datetime() -%}
-    datetime
-{%- endmacro -%}
-
--- a/macros/datetime/date_spine.sql
+++ b/macros/datetime/date_spine.sql
@@ -19,7 +19,7 @@
 
 {# private, don't call from anywhere else #}
 {%- macro date_spine__parse_end_date(end_date) -%}
-  {{ adapter_macro('dbt_utils.prep_end_date_', end_date) }}
+  {{ adapter_macro('dbt_utils.date_spine__parse_end_date', end_date) }}
 {%- endmacro -%}
 
 
--- a/macros/datetime/date_spine.sql
+++ b/macros/datetime/date_spine.sql
@@ -60,7 +60,7 @@ filtered as (
 
     select *
     from all_periods
-    where date_{{datepart}} <= {{ end_date }}
+    where date_{{datepart}} <= {{ dbt_utils.safe_cast(end_date, dbt_utils.datatypes.datetime()) }}
 
 )
 
--- a/macros/datetime/date_spine.sql
+++ b/macros/datetime/date_spine.sql
@@ -60,7 +60,7 @@ filtered as (
 
     select *
     from all_periods
-    where date_{{datepart}} <= {{ dbt_utils.safe_cast(end_date, dbt_utils.datatypes.datetime()) }}
+    where date_{{datepart}} <= {{ dbt_utils.safe_cast(end_date, dbt_utils.type_datetime()) }}
 
 )
 
--- a/macros/datetime/date_spine.sql
+++ b/macros/datetime/date_spine.sql
@@ -17,7 +17,20 @@
 
 {%- endmacro %}
 
+{# private, don't call from anywhere else #}
+{%- macro date_spine__parse_end_date(end_date) -%}
+  {{ adapter_macro('dbt_utils.prep_end_date_', end_date) }}
+{%- endmacro -%}
 
+
+{%- macro default__date_spine__parse_end_date(end_date) -%}
+    {{ end_date }}
+{%- endmacro -%}
+
+
+{%- macro bigquery__date_spine__parse_end_date(end_date) -%}
+    {{ dbt_utils.safe_cast(end_date, dbt_utils.type_datetime()) }}
+{%- endmacro -%}
 
 
 {% macro date_spine(datepart, start_date, end_date) %}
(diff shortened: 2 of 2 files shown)


macros/cross_db_utils/datatypes.sql
```diff
@@ -79,3 +79,30 @@
...
+    datetime
+{%- endmacro -%}
+
+{%- macro postgres__type_datetime() -%}
+    datetime
+{%- endmacro -%}
+
+{%- macro snowflake__type_datetime() -%}
+    datetime
+{%- endmacro -%}
```
drewbanin: I think we can remove all of these and just make dbt-utils use the default implementation (which returns `datetime`)

macros/datetime/date_spine.sql
```diff
@@ -60,7 +60,7 @@ filtered as (
 
     select *
     from all_periods
-    where date_{{datepart}} <= {{ end_date }}
+    where date_{{datepart}} <= {{ dbt_utils.safe_cast(end_date, dbt_utils.datatypes.datetime()) }}
```
drewbanin: I think we just want to call `dbt_utils.type_datetime()` here, but I love the idea of namespacing macros 😍 

```suggestion
    where date_{{datepart}} <= {{ dbt_utils.safe_cast(end_date, dbt_utils.type_datetime()) }}
```

clausherther: Whoops, that must have been wishful thinking! ;)

macros/datetime/date_spine.sql
```diff
@@ -60,7 +60,7 @@ filtered as (
 
     select *
     from all_periods
-    where date_{{datepart}} <= {{ end_date }}
+    where date_{{datepart}} <= {{ dbt_utils.safe_cast(end_date, dbt_utils.type_datetime()) }}
```
drewbanin: I've spent more time than I'd like to admit looking at this line of SQL!

This gives me some amount of pause -- the `safe_cast` macro is intended to cast a _string_ to a more restrictive type, like a datetime. It is _not_ intended to be used with a target type (like a `datetime`) directly. That is:

```
-- This will work great (a string literal)
{{ dbt_utils.safe_cast("'2019-01-01'", dbt_utils.type_datetime()) }}

-- This will not work on Snowflake
-- eg: select try_cast('2019-01-01'::datetime as datetime)
{{ dbt_utils.safe_cast("'2019-01-01'::timestamp", dbt_utils.type_datetime()) }}
```


So, this PR represents a breaking change to `date_spine` on Snowflake at least -- it appears that BigQuery is a little bit more permissive with the types supplied to `safe_cast` though.

I can think of two alternatives:
1. force the caller of this macro to use `{{ dbt_utils.safe_cast(end_date, dbt_utils.type_datetime()) }}` in the macro which calls `date_spine` (we can better document this)
2. add an argument to `date_spine` which, if provided, will use `safe_cast` to cast  `start_date` and `end_date` parameters? (I think this probably isn't a very good approach).


What say you, @clausherther? You buy it?

clausherther: @drewbanin hmm, I see. So, this passed the tests b/c the integration test passes string literals and not date functions like in the usage docs. 

How do you feel about casting `field` in `safe_cast(field, type)` to a varchar in the Snowflake case? Or is that too much brute force?

```
{% macro snowflake__safe_cast(field, type) %}
    try_cast({{field}}::varchar as {{type}})
{% endmacro %}
```

Alternatively, how about **not** safe casting?
```
where date_{{datepart}} <= cast({{ end_date }} as {{ dbt_utils.type_datetime() }})
```

clausherther: Or, would something like this work? Cast the `end_date` only in the BQ case?

```
{# private, don't call from anywhere else #}
{%- macro prep_end_date_(end_date) -%}
  {{ adapter_macro('dbt_utils.prep_end_date_', end_date) }}
{%- endmacro -%}

{%- macro default__prep_end_date_(end_date) -%}
    {{ end_date }}
{%- endmacro -%}

{%- macro bigquery__prep_end_date_(end_date) -%}
    {{ dbt_utils.safe_cast(end_date, dbt_utils.type_datetime()) }}
{%- endmacro -%}
```

Used like:
```
    where date_{{datepart}} <= {{ dbt_utils.prep_end_date_(end_date) }}
```

clausherther: @drewbanin just wanted to check in to see what you thought re: the above.

drewbanin: hey @clausherther - I was traveling over the past week - just getting caught up on PR comments now :)

I don't love the idea of casting to `varchar` in the Snowflake case, and I don't think auto-casting to datetime is a good idea either (ie. as shown in this example):
```
where date_{{datepart}} <= cast({{ end_date }} as {{ dbt_utils.type_datetime() }})
```

That last example you provided, `prep_end_date_` feels like the most reasonable option to me. This is really only necessary on BigQuery, so no need to introduce weirdness for other databases here. Maybe we can call it something explicit like `date_spine__parse_end_date` where the default implementation does nothing but the bq implementation casts to `datetime` as shown in your example.

I'll also say: I think there might be merit to better documenting the supported datatypes for the `date_spine` macro on each database, then requiring users to cast the `end_date` appropriately in their calling macro. That's certainly not as immediately convenient, but it could be a helpful constraint which simplifies the implementation of macros like `date_spine`.

Let me know what you think!

clausherther: Cool, makes sense. I like the "prep" macro option the best as well.
So, would it be best to just add this as a private macro inside of the date_spine macro file?

```

Comment (clausherther) [reply macros/datetime/date_spine.sql]: Cool, makes sense. I like the "prep" macro option the best as well.
So, would it be best to just add this as a private macro inside of the date_spine macro file?
```
{# private, don't call from anywhere else #}
{%- macro date_spine__parse_end_date(end_date) -%}
  {{ adapter_macro('dbt_utils.prep_end_date_', end_date) }}
{%- endmacro -%}

{%- macro default__ date_spine__parse_end_date(end_date) -%}
    {{ end_date }}
{%- endmacro -%}

{%- macro bigquery__ date_spine__parse_end_date(end_date) -%}
    {{ dbt_utils.safe_cast(end_date, dbt_utils.type_datetime()) }}
{%- endmacro -%}
```

and then update the date_spine macro to use it?
```
where date_{{datepart}} <= {{ dbt_utils.date_spine__parse_end_date(end_date) }}
```

File: macros/datetime/date_spine.sql [hunk 4]
```diff
@@ -17,8 +17,21 @@
 
 {%- endmacro %}
 
+{# private, don't call from anywhere else #}
+{%- macro date_spine__parse_end_date(end_date) -%}
+  {{ adapter_macro('dbt_utils.prep_end_date_', end_date) }}
```
drewbanin: I think this should be
```suggestion
  {{ adapter_macro('dbt_utils.date_spine__parse_end_date', end_date) }}
```

macros/datetime/date_spine.sql
drewbanin: This sounds great to me. I made one suggestion inline but will kick off the CI build when that's addressed!

clausherther: Ooops, sorry I missed that one, committed!

kyleries:
@drewbanin 👋  - I've tested this branch in our project and it's working well. Anything I can help with to get it merged? Docs, tests, etc?

@clausherther 👏  for this PR!

clrcrl:
We ended up changing the README here to something that _is_ cross-db compatible (afaik)
```
{{ dbt_utils.date_spine(
    datepart="day",
    start_date="cast('2019-01-01' as date)",
    end_date="cast('2020-01-01' as date)"
) }}
```
Closing this PR as a result, as docs are easier to maintain than extra macros :D

PR #168, opened 2019-10-16
Feature: add uniqueness_test option to cardinality equality test

kevinsanz93:
I needed something somewhere in between the relationship test and this cardinality test.

The test needed to compare the unique values across two different sources' models to ensure that the unique values were in lock-step across both systems. The current test fails if the total count of rows was different even if the distinct values were shared.

Technically having two relationship tests  (one pointing from A->B and one from B->A) would work, but I like the convenience of being able to run just one test.

This is my first attempt at forking a public repo and trying to contribute, let me know if I did anything wrong or missed a step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ models:
 ```
 
 #### cardinality_equality ([source](macros/schema_tests/cardinality_equality.sql))
-This schema test asserts if values in a given column have exactly the same cardinality as values from a different column in a different model.
+This schema test asserts if values in a given column have exactly the same cardinality as values from a different column in a different model. An option to test that the two columns share unique values is provided via the test_uniqueness option.
 
 Usage:
 ```yaml
@@ -209,6 +209,7 @@ models:
           - dbt_utils.cardinality_equality:
               field: other_column_name
               to: ref('other_model_name')
+              test_uniqueness: false
 
 ```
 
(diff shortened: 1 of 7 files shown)


kevinsanz93:
It seems like the build failed because of an issue with the redshift build env, and not due to the changes. :(

drewbanin:
hey @kevinsanz93! thanks for opening this PR! We'll check it out -- just wanted to give you a heads up that I saw it and that I think it's neat :)

Don't worried about those failing tests -- I'll kick those off for you when I review this PR!

drewbanin:
Hey @kevinsanz93 - thanks for opening up this PR!

You definitely did everything right here - nice work on your first PR to a public repo :) 

This PR is interesting - it sits right in between the functionality defined in the `cardinality_equality` and the `equality `tests defined in this package. I like that you noted something similar can be accomplished with the existing `cardinality_equality` test in conjunction with a couple of other relationships tests. While I recognize the benefit of the logic change in this PR (should the logic you've shown here not have been the default?) I hesitate to accept a PR for this particular macro in dbt-utils.

This package is becoming hard to maintain as macros become more complex over time. In general, I want all of the macros in here to be simpler and accept fewer arguments - this will make them easier to test and maintain. We need to do a better job of clarifying how this particular package is going to evolve over time (I think we should have more packages that have a tighter scope) but I did want to get back to you here sooner rather than later.

For the moment, I'm going to close this PR but please do feel free to let me know if you have any questions or comments. I'm really grateful you took the time to open this one up and very happy to discuss it further.

Thanks!

PR #169, opened 2019-10-18
Add a new schema test: numeric_constraints

soltanianalytics:
Hi all!

This is my first pull request, so please forgive amateur mistakes (but do point them out ruthlessly!). I used [this PR](https://github.com/fishtown-analytics/dbt-utils/pull/161) as guidance on how to structure the PR.

To quote @MartinGuindon directly, I'm also submitting this PR to add a new schema test that has proven to be extremely useful.

Essentially, I want to be able to easily test numeric columns for values in various ways. For instance, I need to be sure that VAT values are indeed percentages within a certain range, or that VAT is zero for international transactions.

Thought I'd share it with you :)

Best,
Bijan

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,50 @@ models:
 
 ```
 
+#### numeric_constraints ([source](macros/schema_tests/numeric_constraints.sql))
+This schema test asserts that a column of numerical value is within specified bounds.
+
+Usage:
+```yaml
+version: 2
+
+models:
+  - name: model_name
+    columns:
+      - name: a_number
+        tests:
+          - dbt_utils.numeric_constraints:
+              gte: 0
+              lt: 180
+```
+
+Possible options:
+```
+eq:   equal
+ne:   not equal
+gt:   greater than
+gte:  greater than or equal
+lt:   lower than
+lte:  lower than or equal
+```
+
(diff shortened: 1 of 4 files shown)


drewbanin:
Hey @soltanianalytics - I'm just getting around to checking this out. It's a really cool PR, thanks for making it! I think you did a great job here :)

I'll give this a more thorough review shortly :+1:

drewbanin:
Hey @soltanianalytics - this is a super cool PR!

I don't think it's something we're going to want to merge into dbt-utils, as there are slightly more complicated ways of accomplishing the same functionality with existing tests today. One example would be to provide an expression to the [expression_is_true](https://github.com/fishtown-analytics/dbt-utils#expression_is_true-source) macro.

I do believe that your test is super useful and I'm grateful that you opened up a PR to contribute it back! If you have other macros like this, you should definitely feel free to stand up your own package which contains this macro and others like this.

We're generally going to be a little more conservative around the introduction of macros to this package going forwards. It can be difficult to maintain all of these macros in a cross-database way, and ultimately, we think dbt's package management has matured enough to where it's tractable enough for folks to stand up their own packages for things like this :)

Let me know if you have any questions or comments - I do really appreciate you sending this through!

PR #174, opened 2019-11-07
Reducing whitespace in group_by

ltchikindas:
reducing whitespace in compiled SQL for legibility

select
    foo.a 
  , foo.b 
  , foo.c 
  , sum(foo.d)
from
  foo 
{{group_by(n=3)}} 

now results in 

select
    foo.a 
  , foo.b 
  , foo.c 
  , sum(foo.d)
from
  foo 
group by
    1,2,3

### DIFF
--- a/macros/sql/groupby.sql
+++ b/macros/sql/groupby.sql
@@ -1,9 +1,8 @@
-{% macro group_by(n) %}
+{%- macro group_by(n) -%}
 
   group by
-    {% for i in range(1, n + 1) %}
-      {{ i }}
-      {% if not loop.last %} , {% endif %}
-   {% endfor %}
+    {% for i in range(1, n + 1) -%}
+      {{ i }}{{ ',' if not loop.last }}   
+   {%- endfor -%}
 
-{% endmacro %}
+{%- endmacro -%}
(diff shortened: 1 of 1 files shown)


drewbanin:
Thanks @ltchikindas! This one was a long time coming :) 

Just kicked off a build here - this will be good to merge when the tests pass :)

macros/sql/groupby.sql
```diff
@@ -1,9 +1,8 @@
...
+{%- macro group_by(n) -%}
 
   group by
-    {% for i in range(1, n + 1) %}
-      {{ i }}
-      {% if not loop.last %} , {% endif %}
-   {% endfor %}
+    {% for i in range(1, n + 1) -%}
+      {{ i }}{{ ',' if not loop.last }}   
+   {%- endfor -%}
```
drewbanin: Actually - one minor nit - can we make this read:
```
{%- macro group_by(n) -%}
  group by {% for i in range(1, n + 1) -%}
      {{ i }}{{ ',' if not loop.last }}   
   {%- endfor -%}
{%- endmacro -%}
```
The difference is that the SQL renders out as
```
select
  1 as id,
  2 as name
group by 1,2
```
instead of
```
select
  1 as id,
  2 as name
group by
  1,2
```

ltchikindas: done!

PR #179, opened 2019-12-04
Feature/width bucket macro

clrcrl:
After speaking with @clausherther, we decided that I'd finish this PR off (since I'd done lots of the work anyway to figure out why the tests were failing). 

Original PR is #145


clausherther:
Nice work @clrcrl, thanks!

PR #182, opened 2019-12-20
Remove redundant SELECT

Fokko:
Spark does not like the additional `SELECT` in front of the `abs(..)`. Therefore I'd suggest removing it, if that is ok with the other database engines.

### DIFF
--- a/macros/schema_tests/equality.sql
+++ b/macros/schema_tests/equality.sql
@@ -50,12 +50,12 @@ unioned as (
 
 final as (
 
-    select (select count(*) from unioned) +
-        (select abs(
+    select (
+      select count(*) from unioned) +
+        abs(
             (select count(*) from a_minus_b) -
             (select count(*) from b_minus_a)
-            ))
-        as count
+      ) as count
 
 )
 


drewbanin:
hey @Fokko - thanks for making this PR! We could be clearer about this, but for now, dbt-utils only works with postgres, redshift, snowflake, and BigQuery. All of this code would become a _huge_ mess if we tried to support presto + spark + other plugins in the future too. There's just too much variance between database engines for this to be tractable IMO.

I'm going to close out this particular PR, but I'm very grateful you took the time to make it! Let's figure out a way to introduce code like this more directly into the spark integration tests. We should probably drop the dependency on dbt-utils in those tests...

Fokko:
Thanks, Drew for the elaborated answer. I understand your concern, but in this case, I would consider accepting this PR since it does not change the behavior of the query. I understand that Spark is a second class citizen right now, but supporting this would be nice anyway.

My experiences with Spark and DBT are in general very good. Personally I would not support Presto since it is only a query layer, instead of a processing framework such as Spark.

PR #190, opened 2020-02-13
add get_relations_by_suffix

ktshannon:
Copying the same methods as prefix, but replicating for suffix needs.


ktshannon:
Closing as this is an MVP meant for a single project, will look into proper refactor to use the same method for both prefix/suffix.

Issue #125, opened 2019-03-17
new macro: `safe_add()`


jthandy:
I'm a little surprised we don't have something for this yet! It's a simple but neat use case. Here's some code that I recently saw in a PR:

```
coalesce(orders.total_gross_revenue, 0) + 
            coalesce(refunds.total_refund_amount, 0) +
            coalesce(orders.total_discounts, 0) + 
            coalesce(refunds.total_refund_shipping_amount, 0) +
            coalesce(orders.total_shipping_cost, 0) + 
            coalesce(refunds.total_refund_tax_amount, 0) +
            coalesce(orders.total_tax, 0)
            as total_sales
```

Adding up the values for a bunch of different fields happens frequently in an ecommerce context (and likely in other contexts as well, although this is where I see it most). This could be solved by a function very similar to `surrogate_key()`--not fancy, but useful.

eogilvy12:
Love this idea!!

jtcohen6:
Really like this. Reminds me of a related concept: `safe_greatest`, `safe_least`.
```
nullif(greatest(
    coalesce(source_one.updated_at,'0000-01-01'),
    coalesce(source_two.updated_at,'0000-01-01'),
    coalesce(source_three.updated_at,0000-01-01')
),'0000-01-01')
```
```
nullif(least(
    coalesce(source_one.updated_at,'9999-01-01'),
    coalesce(source_two.updated_at,'9999-01-01'),
    coalesce(source_three.updated_at,9999-01-01')
),'9999-01-01')
```

PR #191, opened 2020-02-14
Add safe add macro

invinceyble:
#125 looked pretty straightforward and fun to do. :) 

I wasn't sure about how to actually format the output of the macro (re: spacing / style) though, so take a closer look there.

I'm also not super sure how running these tests work, though I'm hoping that CI will pick it up and run them? 🤞

### DIFF
--- a/macros/sql/safe_add.sql
+++ b/macros/sql/safe_add.sql
@@ -1,0 +1,15 @@
+{%- macro safe_add() -%}
+
+{% set fields = [] %}
+
+{%- for field in varargs -%}
+
+    {% set _ = fields.append(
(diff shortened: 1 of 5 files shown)


clrcrl:
I'll kick off the tests now!

macros/sql/safe_add.sql
```diff
@@ -0,0 +1,15 @@
+{%- macro safe_add() -%}
+
+{% set fields = [] %}
+
+{%- for field in varargs -%}
+
+    {% set _ = fields.append(
```
clrcrl: Here you can use `do`:
```
{% do fields.append("coalesce(" ~ field ~ ", 0)") %}
```

PR #197, opened 2020-03-19
Allow list input for surrogate key macro 

DylanBaker:
Per [this conversation](https://getdbt.slack.com/archives/C2JRRQDTL/p1584050118056100) in dbt Slack, I wanted to explore the idea of allowing the surrogate key macro to support lists, instead of the varargs method.

This PR:
* Adds a deprecation warning if the varargs method is used.
* Allows for a single list argument to the surrogate macro. 
* Updates the surrogate key tests.

### DIFF
--- a/macros/sql/surrogate_key.sql
+++ b/macros/sql/surrogate_key.sql
@@ -3,7 +3,7 @@
 
 {%- if varargs|length >= 1 %}
 
-{{ exceptions.warn("Warning: the `surrogate_key` macro now takes a single list argument instead of multiple string arguments. Multiple string arguments are longer supported and will be deprecated in a future release of dbt-utils.") }}
+{%- do exceptions.warn("Warning: the `surrogate_key` macro now takes a single list argument instead of multiple string arguments. Multiple string arguments are longer supported and will be deprecated in a future release of dbt-utils.") -%}
 
 {# first argument is not included in varargs, so add first element to field_list_xf #}
 {%- set field_list_xf = [field_list] -%}
--- a/macros/sql/surrogate_key.sql
+++ b/macros/sql/surrogate_key.sql
@@ -1,8 +1,28 @@
-{%- macro surrogate_key() -%}
+{%- macro surrogate_key(field_list) -%}
 
-{% set fields = [] %}
 
-{%- for field in varargs -%}
+{%- if varargs|length >= 1 %}
+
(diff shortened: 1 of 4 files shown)


macros/sql/surrogate_key.sql
```diff
@@ -1,8 +1,28 @@
-{%- macro surrogate_key() -%}
+{%- macro surrogate_key(field_list) -%}
 
-{% set fields = [] %}
 
-{%- for field in varargs -%}
+{%- if varargs|length >= 1 %}
+
+{{ log("Warning: the `surrogate_key` macro now takes a single list argument instead of multiple string arguments. Multiple string arguments are longer supported and will be deprecated in a future release of dbt-utils.", info=True) }}
```
clrcrl: Heya, check out this [nifty exception](https://docs.getdbt.com/docs/writing-code-in-dbt/jinja-context/exceptions/#warn)

This wasn't available to me when I last wrote code like this, but now you can do:

```suggestion
{{ exceptions.warn("Warning: the `surrogate_key` macro now takes a single list argument instead of multiple string arguments. Multiple string arguments are longer supported and will be deprecated in a future release of dbt-utils.") }}
```

clrcrl: But then again, implementing that will necessitate a change to `require-dbt-version` to `>=0.15.1` in the `dbt_project.yml` file

clrcrl: Let's do it! I'll also update my code to use this in a separate PR :)

clrcrl: See #198

clrcrl:
Oh, and can you also update the readme!

macros/sql/surrogate_key.sql
DylanBaker: Perfect. Makes sense! Have committed your suggested change.

DylanBaker:
Yeah, absolutely. Do you want me to write something about the old deprecated version or just the new desired input?

clrcrl:
I'm happy to just have the updated usage in the readme, and rely on the deprecation warning to communicate the change in arguments. In other cases where we've deprecated, we've documented it in the README, but that was because we renamed macros entirely!

BTW, I just updated the docs on this — can you use `{% do exceptions.warn() %} ` instead of `{{ exceptions.warn() }}`? Not using the `do` syntax can lead to some funky behavior. As soon as that's done 

I also just merged my PR so you can rebase off master if you want :)

clrcrl:
🤞 Just kicked off the tests

DylanBaker:
🙌

PR #206, opened 2020-04-07
Add epoch_to_timestamp macro

clrcrl:
Required for `dbt-slack` package :)


clrcrl:
Soooo I don't know what to do here.
My local postgres setup is completely borked, so I can't test this locally, which is why my email inbox looks like this:
Screen Shot 2020-06-29 at 4 51 46 PM

I don't know what to do here to fix it! Any suggestions?

clrcrl:
It's not worth it

PR #208, opened 2020-04-10
Adding logic to unpivot to handle boolean columns.

avishalom:
A boolean value will be replaced with 'true' or 'false'.
The rest remains unchanged.
added test

### DIFF
--- a/macros/sql/unpivot.sql
+++ b/macros/sql/unpivot.sql
@@ -52,7 +52,12 @@ Arguments:
       {%- endfor %}
 
       cast('{{ col.column }}' as {{ dbt_utils.type_string() }}) as {{ field_name }},
-      cast({{ col.column }} as {{ cast_to }}) as {{ value_name }}
+      cast(  {% if col.data_type == 'boolean' %}
+           case when {{ col.column }} then 'true' else 'false' end
+             {% else %}
+           {{ col.column }}
+             {% endif %}
+           as {{ cast_to }}) as {{ value_name }}
 
     from {{ relation }}
 
--- a/README.md
+++ b/README.md
@@ -556,6 +556,7 @@ Arguments:
 
 #### unpivot ([source](macros/sql/unpivot.sql))
 This macro "un-pivots" a table from wide format to long format. Functionality is similar to pandas [melt](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.melt.html) function.
+Boolean values are replaced with the strings 'true'|'false'
 
 Usage:
 ```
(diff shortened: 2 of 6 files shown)


macros/sql/unpivot.sql
```diff
@@ -52,7 +52,12 @@ Arguments:
       {%- endfor %}
 
       cast('{{ col.column }}' as {{ dbt_utils.type_string() }}) as {{ field_name }},
-      cast({{ col.column }} as {{ cast_to }}) as {{ value_name }}
+      cast(  {% if col.data_type == 'boolean' %}
```
drewbanin, line 55: I think we might want to make a cross-database adapter macro that renders this `case when` statement. That will let us use this logic in other macros. Additionally, I think that databases like BigQuery will use the type `BOOL`, not `boolean` (though I'd need to double-check this to be sure!). The adapter macro pattern will let us define custom macro implementations on a per-adapter basis here

avishalom: Should I wrap this in a targtt=redshift test? 
(I'm too lazy to connect to databases that I don't normally use to test. maybe other members have them ready.)

drewbanin: Nope - I'd just add a file in [this dir](https://github.com/fishtown-analytics/dbt-utils/tree/master/macros/cross_db_utils) using a file like [this one](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/cross_db_utils/hash.sql) as a template. You can call this file something like `bool_cast.sql`. The `default__` implementation can just return
```
cast({{ col }} as {{ type }})
```

and the Redshift implementation can use the `case when ...` statement you've shown here.

I actually did some more digging, and it looks like Redshift is the only database dbt supports which lacks the ability to cast a boolean to a string. Go figure.

Regardless, I'd rather encapsulate this logic in an adapter-macro than hardcode any `if target == redshift` logic in here. Code like that becomes a big mess real fast :)

avishalom: Thanks, I wasn't familiar with that pattern. 

oh, it looks like [safe_cast](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/cross_db_utils/safe_cast.sql)
already exists, though I'm not 100% sure what the behavior is,
can you override an sql function with a macro? 
or do you need to invoke it as <package>.<macro_name>() ? 
within the namespace of the package dbt_utils [line 19](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/cross_db_utils/safe_cast.sql#L19) is ambiguous with a macro named the same as a function. 
the fact that it isn't an error probably means that with no package qualifier the native function takes precedence.

clrcrl:
Closing in favor of #305

PR #209, opened 2020-04-10
Fix the `unpivot` test to correct behaviour of `remove`, Fix the data to align with it.

avishalom:
Remove expects an iterable, 
wrapping values with a [] so that it doesn't list('value') by accident


drewbanin:
Thanks @avishalom - I just kicked off the automated tests here

clrcrl:
Closing in favor of @249

PR #211, opened 2020-04-16
Simplify running the integration tests

bjgbeelen:
During the development and discussion of https://github.com/fishtown-analytics/dbt-utils/pull/205 I wanted to run the integration tests against a Spark backend. To make this happen, I made some changes presented here. While adding Spark integration is something for the long run, I already did some work that I think might be valuable for others users too. 

It's twofold:

- making it simpler to run the integration tests locally
- and, also make sure that they mimick what is happening in CircleCI as best as possible.

I have not seen the results yet on CircleCI of course, so it might need some extra attention at things I overlooked


integration_tests/ci/sample.profiles.yml
```diff
@@ -11,7 +11,7 @@ integration_tests:
   outputs:
     postgres:
       type: postgres
-      host: localhost
+      host: "{{ env_var('CI_DBT_HOST') }}"
```
bjgbeelen, line 14: This would mean btw that this environment variable would need to be set in CircleCI too, or we should put `localhost` as a default value, but that would not be my preference.

Now that I take another look on it, adding `POSTGRES` somewhere in these environment variables, e.g. `CI_POSTGRES_DBT_HOST` would probably make the naming more consistent in the profiles.yml

drewbanin:
hey @bjgbeelen - this is awesome!

Your PR here inspired me to play around with CircleCI's _workflows_. I made a branch for this over here: https://github.com/fishtown-analytics/dbt-utils/compare/pr/211-drew?expand=1

The key benefits are that:
1. tests for each adapter can run in parallel
2. we can enable build-on-PRs for PRs submitted from forks. We've historically disabled this because it's a vector for abuse, but I think workflows would let us at least kick off the Postgres tests here for forked PR builds

I'm pretty happy with this PR as-is. I'm going to do one more pass here, and then this will be good to merge :)

bjgbeelen:
Glad you like it! Sounds like great new benefits you're investigating!

PR #213, opened 2020-04-17
Add distinct argument to pivot macro

DylanBaker:
This PR adds a `distinct` argument to the pivot macro. When set to true, it adds `distinct` to the aggregate, immediately after the opening bracket.

### DIFF
--- a/macros/sql/pivot.sql
+++ b/macros/sql/pivot.sql
@@ -32,12 +32,12 @@ Arguments:
     alias: Whether to create column aliases, default is True
     agg: SQL aggregation function, default is sum
     cmp: SQL value comparison, default is =
-    dictinct: Whether to use distinct in the aggregation, default is False
     prefix: Column alias prefix, default is blank
     suffix: Column alias postfix, default is blank
     then_value: Value to use if comparison succeeds, default is 1
     else_value: Value to use if comparison fails, default is 0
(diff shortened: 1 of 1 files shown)


macros/sql/pivot.sql
```diff
@@ -44,13 +45,15 @@ Arguments:
                alias=True,
                agg='sum',
                cmp='=',
+               distinct=False,
```
clrcrl, line 48: I think new arguments should go at the bottom of the list, incase someone (silly) is relying on the order of arguments for optional args 😓

macros/sql/pivot.sql
```diff
@@ -32,6 +32,7 @@ Arguments:
     alias: Whether to create column aliases, default is True
     agg: SQL aggregation function, default is sum
     cmp: SQL value comparison, default is =
+    dictinct: Whether to use distinct in the aggregation, default is False
```
clrcrl, line 35:
```suggestion
    distinct: Whether to use distinct in the aggregation, default is False
```

macros/sql/pivot.sql
DylanBaker: Of course. Now amended!

clrcrl:
Closing in favor of #248, which I had to open to get the tests to pass (a rebase was needed)

PR #215, opened 2020-04-29
Return target relation in insert by period materialization

jtalmi:
```
* Deprecation Warning: The materialization ("insert_by_period") did not
explicitly return a list of relations to add to the cache. By default the target
relation will be added, but this behavior will be removed in a future version of
dbt.
For more information, see:
https://docs.getdbt.com/v0.15/docs/creating-new-
materializations#section-6-returning-relations
```


clrcrl:
Closing in favor of #236

PR #218, opened 2020-05-01
Bump required dbt version

tjengel:
## Description & motivation
This PR updates the required dbt version for this package, as requested in this issue https://github.com/fishtown-analytics/dbt-utils/issues/210


clrcrl:
So, uh, this was my mistake. I think we don't actually need 0.16.0, and I got the release notes wrong. I'm going to update the release notes rather than reduce the dbt version range.

PR #220, opened 2020-05-08
feature/get relations by schema prefix

JDW818:
## Description & motivation

This was inspired by our work with Alamo Drafthouse. We loaded data from 40+ nearly identical instances of SQL server into theater-specific schemas. The schemas follow this pattern - venue_0001, venue_0002, etc. Table names are identical across schemas. The catch is that not all theater schemas contain every single table. So, we needed a way to dynamically define relations and union like tables from all theaters.

To accomplish this,  I built a macro based around `get_relations_by_prefix`. Instead of looking for a table prefix, this macro looks for a schema prefix + explicit table name.

#### 6/29 Update:
After discussing with @clrcrl, we decided to make a combined version of this macro and `get_relations_by_prefix`, called `get_relations_by_pattern`. We will eventually deprecate `get_relations_by_prefix` and a warning has been added accordingly.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -492,7 +492,7 @@ get_relations_by_prefix, it's particularly handy paired with `union_relations`.
 
 **Args:**
 * `schema_pattern` (required): The schema pattern to inspect for relations.
-* `table_pattern` (required): The name of the table/view (case insensitive)
+* `table_pattern` (required): The name of the table/view (case insensitive).
 * `exclude` (optional): Exclude any relations that match this table pattern.
 * `database` (optional, default = `target.database`): The database to inspect
 for relations.
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ get_relations_by_prefix, it's particularly handy paired with `union_relations`.
 {% set relations = dbt_utils.get_relations_by_pattern('schema_pattern', 'table_pattern%') %}
 
 -- Returns a list of relations as above, excluding any that end in `deprecated`
-{% set relations = dbt_utils.get_relations_by_pattern('schema_pattern',', 'table_pattern%', '%deprecated') %}
+{% set relations = dbt_utils.get_relations_by_pattern('schema_pattern', 'table_pattern%', '%deprecated') %}
 
 -- Example using the union_relations macro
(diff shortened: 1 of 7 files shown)


JDW818:
@clrcrl just checking to get your thoughts on this one!

clrcrl:
Hey @JDW818 :wave:

This looks great! Definitely a valuable addition to utils.

Two thoughts:
1. I think we need to integration tests to get this merged — do you feel comfortable copying the pattern for the test for `get_relations_by_prefix` and adapting it for this macro?
2. Do you think we should instead generalize `get_relations_by_prefix` to do both things. e..g:

Before:
```
-- to return <my_schema>.<my_table_prefix>*
-- before
{% set relations = dbt_utils.get_relations_by_prefix('my_schema', 'my_table_prefix') %}
-- after
{% set relations = dbt_utils.get_relations_by_prefix('my_schema', 'my_table_prefix%') %}
-- (note new % sign)

-- to return <my_schema>*.<my_table_prefix>
-- before
{% set relations = dbt_utils.get_relations_by_schema_prefix('my_schema_prefix', 'my_table_name') %}
-- after
{% set relations = dbt_utils.get_relations_by_prefix('my_schema_prefix%', 'my_table_name') %}
```

This would be a breaking change, since people who are currently using the `get_relations_by_prefix` would have to change from `my_table_prefix` to `my_table_prefix%`. Perhaps a better option is to instead deprecate that macro, and create a more general macro `get_relations_by_pattern`?

What do you think?

JDW818:
@clrcrl Per our convo, I updated this to work for both table and schema prefixes. I also added a deprecation warning to `get_relations_by_prefix`.

clrcrl:
I've just rebased this from master so that the tests pass!

JDW818:
@clrcrl anything else on this or are we good to merge?

clrcrl:
Sorry, I literally have a to-do that looks like this:

Screen Shot 2020-07-08 at 2 47 55 PM

I decided to remove the warning (the most recent commit), because then we can make a nicer release rather than a breaking change release. This has really been a mess, and I'm very sorry about this :/

PR #223, opened 2020-05-29
Replace __TABLES_SUMMARY__ with INFORMATION_SCHEMA

preston-hf:
## Description & motivation
INFORMATION_SCHEMA is actually [documented](https://cloud.google.com/bigquery/docs/information-schema-tables) and it requires less permissions than `__TABLES_SUMMARY__` which is not standardized and seems to require the ability to read the table, where INFORMATION_SCHEMA is queryable with just metadata access.

## Checklist

### DIFF
--- a/macros/sql/get_tables_by_prefix_sql.sql
+++ b/macros/sql/get_tables_by_prefix_sql.sql
@@ -17,11 +17,11 @@
 {% macro bigquery__get_tables_by_prefix_sql(schema, prefix, exclude='', database=target.database) %}
     
         select distinct
-            dataset_id as table_schema, table_id as table_name
+            table_schema, table_name
 
-        from {{adapter.quote(database)}}.{{schema}}.__TABLES_SUMMARY__
-        where dataset_id = '{{schema}}'
-            and lower(table_id) like lower ('{{prefix}}%')
-            and lower(table_id) not like lower ('{{exclude}}')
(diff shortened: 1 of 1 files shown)


clrcrl:
@jtcohen6 — do you have an instinct here? I am not the best person to review BQ code

jtcohen6:
My impression is that `INFORMATION_SCHEMA` is newer and better in almost every way. The syntax is sugarier, for one, and more similar to how other databases store this info. I didn't know about the differences in required permissions, which is even more compelling.

I wouldn't say `__TABLES_SUMMARY__` is _undocumented_, exactly—just that it's older and thereby included in the documentation for older & odder features, such as [wildcard/sharded tables](https://cloud.google.com/bigquery/docs/querying-wildcard-tables).

My only qualm about `INFORMATION_SCHEMA` is this note in the [docs](https://cloud.google.com/bigquery/docs/information-schema-intro):
> This feature is in a pre-release state and might change or have limited support. For more information, see the product launch stages.

That said, I think it's worth going ahead with this. From the release notes, it looks like they've been in beta for [over a year](https://cloud.google.com/bigquery/docs/release-notes#February_04_2019), and the changes since then (in November and April) seem fairly minor.

preston-hf:
I'll just add that GCP Support just recommended that I use `INFORMATION_SCHEMA`, fwiw

clrcrl:
Closing in favor of #247 (I had to rebase before the tests would run)

PR #224, opened 2020-06-02
Add JSON not null test for columns.

elliottohara:
## Description & motivation
Added test to confirm that json paths are not null in json values. 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,19 @@ case we recommend using this test instead.
 ```
 
 ---
+#### json_paths_not_null ([source](macros/schema_tests/json_paths_not_null.sql))
+This test confirms that all rows in a model have json values for the paths specified in the column tested against.
+
+**Usage:**
+```yaml
+- name: subscriptions
+  columns:
+    - name: settings_json
+      tests:
+        - dbt_utils.json_path_not_null:
+            paths: ['preferences.email_type', 'preferences.email_address']
+```
+
 ### SQL helpers
 #### get_query_results_as_dict ([source](macros/sql/get_query_results_as_dict.sql))
 This macro returns a dictionary from a sql query, so that you don't need to interact with the Agate library to operate on the result
--- a/macros/schema_tests/json_paths_not_null.sql
+++ b/macros/schema_tests/json_paths_not_null.sql
@@ -1,0 +1,22 @@
+{% macro test_json_paths_not_null (model) %}
+    {% set column_name = kwargs.get('column_name') %}
+    {% set paths = kwargs.get('paths', []) %}
+
+    with validation_errors as (
+        select
+            {% for path in paths %}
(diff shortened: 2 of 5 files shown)


elliottohara:
I didn't add tests, because, well, I've not written tests for macros yet. 

Looking through examples now. 

This has been tested and used in our production snowflake db though.

clrcrl:
Hey @elliottohara — this is cool! I think this code is only going to work on Snowflake though, and we'd definitely love some tests here.

Do you feel well-placed to try to follow the pattern for tests? I can help out with making the code multi-warehouse compatible :)

elliottohara:
Test added.

clrcrl:
Dope! I'm pretty snowed under this week, but am putting time in my calendar to review all the open PRs on this repo on Friday 😄

README.md
```diff
@@ -375,6 +375,19 @@ case we recommend using this test instead.
...
+This test confirms that all rows in a model have json values for the paths specified in the column tested against.
+
+**Usage:**
+```yaml
+- name: subscriptions
+  columns:
+    - name: settings_json
+      tests:
+        - dbt_utils.json_path_not_null:
+            paths: ['preferences.email_type', 'preferences.email_address']
```
clrcrl, line 388: I think my preference would be to have this macro work like so:
```suggestion
        - dbt_utils.json_path_not_null:
            path: preferences.email_type
        - dbt_utils.json_path_not_null:
            path: preferences.email_address
```
That way, it will be easier to debug when this fails, as you can figure out exactly which path failed.

This will require changes to the macro code (see below).

FWIW, here's the way to write this on different warehouses:
```
-- snowflake
settings_json:preferences.email_address
settings_json:preferences:email_address

-- bigquery
settings_json.preferences.email_address

-- redshift
json_extract_path_text(settings_json, 'preferences', 'email_address'),

-- postgres
settings_json -> 'preferences' ->> 'email_address',
```

One side benefit of only testing one path at a time, is that we could then provide each path as an array:
```
        - dbt_utils.json_path_not_null:
            path: [preferences, email_type]
        - dbt_utils.json_path_not_null:
            path: [preferences, email_address]
```
(or at least do so more cleanly)

Having the keys in an array will make it easier for someone to (later) add multi-warehouse compatibility!

macros/schema_tests/json_paths_not_null.sql
```diff
@@ -0,0 +1,22 @@
+{% macro test_json_paths_not_null (model) %}
+    {% set column_name = kwargs.get('column_name') %}
+    {% set paths = kwargs.get('paths', []) %}
+
+    with validation_errors as (
+        select
+            {% for path in paths %}
```
clrcrl, line 7: As above, I think it might be better to get rid of this `for` loop, as it will make this test easier to debug if it fails

clrcrl:
Closing due to inactivity

PR #225, opened 2020-06-03
Add accepted values from source schema test

elliottohara:
## Description & motivation
I found myself doing a `SELECT DISTINCT(blah) FROM BOOM` and using that to add an `accepted_values` test. For obvious reasons, that's suboptimal, so I threw together this schema test.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,27 @@ models:
 
 ```
 
+#### accepted_values_from([source](macros/schema_tests/accepted_values_from.sql))
+Much like the built in accepted_values test this test asserts that all values of a given column are
+in a list. However the list comes from a source or relationship.  
+
+Usage:
+```yaml
+- name: test_accepted_values_from_with_rename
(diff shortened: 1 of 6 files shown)


clrcrl:
Hmm, why not use a [relationships](https://docs.getdbt.com/reference/resource-properties/tests/#relationships) test for this?

elliottohara:
Relationship test would work, but I always think of a relationship test as an actual relationship between entities. I feel like making it explicit that the datatype is some enum(ish) populated from some other source is a bit more explicit.

clrcrl:
Hmm, I see your point.

I think when we start using "relationships" to render information like an ERD, it would be valid to have a separate test (though, it would be my hope that we don't just rely on the presence of a test to build this information).

I'm a little hesitant to add this code as it _is_ functionally the same as the relationships test. As a result I'm going to close this issue.

PR #226, opened 2020-06-04
Added ability to use expression_is_true as a column test

elliottohara:
## Description & motivation
I love the expression_is_true schema test, but often it feels like the tests I use them for should be a child of the column I'm testing against, instead of the whole relationship. So I modified the expression_is_true test to do exactly that. 

Left the original tests as is (no regressions), and added 2 more tests to test with condition set and without it. 

## Checklist

### DIFF
--- a/macros/schema_tests/expression_is_true.sql
+++ b/macros/schema_tests/expression_is_true.sql
@@ -1,6 +1,7 @@
 {% macro test_expression_is_true(model, condition='true') %}
 
 {% set expression = kwargs.get('expression', kwargs.get('arg')) %}
+{% set column_name = kwargs.get('column_name') %}
 
 with meet_condition as (
 
@@ -12,7 +13,7 @@ validation_errors as (
     select
         *
     from meet_condition
-    where not({{expression}})
+    where not( {{ expression if column_name is none else [column_name, expression] | join(' ')  }})
 
 )
 
--- a/README.md
+++ b/README.md
@@ -181,6 +181,25 @@ models:
 
 ```
 
+This macro can also be used at the column level. When this is done, the `expression` is evaluated against the column.  
+
+```yaml
+version: 2
+models: 
+    - name: model_name
+      columns: 
+        - name: col_a
+          tests:
+            - dbt_utils.expression_is_true:
+                expression: '>= 1'
+        - name: col_b
+          tests:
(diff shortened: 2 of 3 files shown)


clrcrl:
Oh this is really cool! As I just mentioned on your other PR, I'm going to do a big review on Friday! 

I have a related thought bouncing around in my head:

What if we renamed the `expression_is_true` test to `assert_where_clause` (we have a pattern for deprecations) and created a new test `assert_having_clause`, something like this:

```
{% macro test_assert_having_clause(model, condition='true') %}

{% set expression = kwargs.get('expression', kwargs.get('arg')) %}

with

validation_errors as (

    select
        {{ column_name }}
    from meet_condition
    group by 1
    having not({{ expression }})

)

select count(*)
from validation_errors

{% endmacro %}
```

Then you could call it like:
```yml
version: 2

models:
  - name: orders_snapshot
    columns:
      - name: 
        tests:
          - assert_having_clause: "sum(case when is_current_version then 1 end) = 1"
```

I could imagine it being useful in other spots, e.g. does a ledger of transactions result in a positive entry:

```yml
version: 2

models:
  - name: transactions
    columns:
      - name: order_id
        tests:
          - assert_having_clause: "sum(amount) > 0"
```

Anyway, this is just a side-thought, and definitely not required for this PR at all, but wanted to chat through the thought with someone who is also thinking about this!

elliottohara:
LOVE the assert_having_clause idea! I can think of a few places we'd use it.

macros/schema_tests/expression_is_true.sql
```diff
@@ -12,7 +13,7 @@ validation_errors as (
     select
         *
     from meet_condition
-    where not({{expression}})
+    where not( {{ expression if column_name is none else [column_name, expression] | join(' ')  }})
```
clrcrl, line 16: Hmm this is a little hard to read for me, I think I'd prefer an `if` statement:
```suggestion
{% if column_name is none %}
where not({{ expression }})
{% else %}
where not({{ column_name }} {{ expression }})
{% endif %}
```

clausherther:
I was just looking for something like this today. What do we need to do to get this one over the finish line?

clrcrl:
Closing in favor of #313

PR #238, opened 2020-06-18
New env vars for integration tests

jtcohen6:
* Connect to sandboxed, cost-controlled instances of Redshift, Snowflake, BigQuery
* Use same env var nomenclature that dbt uses for its integration tests


jtcohen6:
Agreed, no release necessary just for this. The change will take effect going forward as long as people are branching off `master`.

PR #242, opened 2020-06-29
Ensure surrogate key handles single arg

### LINKED ISSUE #241: 0.3.0 `surrogate_key` does not support a single string argument
### Describe the bug
dbt_utils 0.3.0 gives off a warning if, when using the surrogate_key macro, you try to use multiple string arguments instead of one list and says that support for multiple string args will be dropped in a future version. this is great and makes sense 

but attempting to run any models that use one string argument results in an error! I think this because the check to see whether the macro is being used with a list or multiple strings mistakenly thinks that a single string is a list and then tries to process the string as a list [https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/surrogate_key.sql#L4-L15](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/surrogate_key.sql#L4-L15)

### Steps to reproduce
If you have a model that looks like this 
```
with foo as (
  select
    1 as column_1,
    2 as column_2
)

select
  {{  dbt_utils.surrogate_key('column_1') }} AS key, --string
  column_1,
  column_2
from foo
```

and you are using `dbt_utils` version `0.2.5` the model will run with no issues!

but if you upgrade to `dbt_utils` version `0.3.0` the model will fail with error 

```
Database Error in model example (models/base/example.sql)
  column "c" does not exist in foo
  compiled SQL at target/run/candid_data_model/models/base/example.sql
```

### Expected results
I'd expect running the model in this way to work the same as if there are multiple string arguments. Alternatively, a better warning message in this case could be helpful as a user who does not know how python processes strings might be a bit confused by the error. 

### Actual results
models fails with error:

```
Database Error in model example (models/base/example.sql)
  column "c" does not exist in foo
  compiled SQL at target/run/candid_data_model/models/base/example.sql
```

### System information
```
packages:
  - package: fishtown-analytics/dbt_utils
    version: [">=0.3.0", "<0.4.0"]
  - package: fishtown-analytics/redshift
    version: [">=0.1.0", "<0.2.0"]
  - package: fishtown-analytics/segment
    version: [">=0.2.0", "<0.3.0"]
  - package: fivetran/github
    version: 0.1.0
```

- [ ] postgres
- [X] redshift
- [ ] bigquery
- [ ] snowflake

```
installed version: 0.17.0
   latest version: 0.17.0

Up to date!

Plugins:
  - bigquery: 0.17.0
  - snowflake: 0.17.0
  - redshift: 0.17.0
  - postgres: 0.17.0
```

**The operating system you're using:**
Mac osX Mojave

**The output of `python --version`:**

```
Python 3.7.7
```

Sure!

clrcrl:
Fixes #241

Updated the tests for this case too


clrcrl:
Oh yikes, I'm leaning no just for simplicity, and so more people upgrade to 0.17.0

PR #251, opened 2020-06-30
Update release flow

clrcrl:
## Description & motivation
As per our discussions — fixing a process problem with documentation rather than fancy tech.

🚨 We need to update our default branches to get this right.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,35 @@
+# dbt-utils releases
+
+## When do we release?
+There's a few scenarios that might prompt a release:
+
+| Scenario                                   | Release type |
+|--------------------------------------------|--------------|
+| New functionality¹                         | minor        |
+| Breaking changes to existing macros        | minor        |
+| Fixes to existing macros                   | patch        |
+| dbt minor release with no breaking changes | patch        |
+| dbt minor release with breaking changes    | minor        |
+
+¹New macros were previously considered patch releases — we have brought them up to minor releases to make versioning for dependencies clearer.
+
+## Branching strategy
+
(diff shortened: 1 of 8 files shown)


clrcrl:
Moving back to draft while we make some new releases :)

run_test.sh
```diff
@@ -6,7 +6,7 @@ if [[ ! -f $VENV ]]; then
     . $VENV
 
     pip install --upgrade pip setuptools
-    pip install dbt
+    pip install  "dbt>=0.17.0,<0.18.0"
```
clrcrl, line 9: This is to placehold the fact that we _will_ increment it before doing a release

clrcrl:
@jtcohen6 — I rewrote almost all of this and just did a force push. I think there's probably still some details to hash out once you've read through it, especially w.r.t. branching strategy.

RELEASE.md
```diff
@@ -0,0 +1,35 @@
...
+| Fixes to existing macros                   | patch        |
+| dbt minor release with no breaking changes | patch        |
+| dbt minor release with breaking changes    | minor        |
+
+¹New macros were previously considered patch releases — we have brought them up to minor releases to make versioning for dependencies clearer.
+
+## Branching strategy
+
+At any point, there should be two long-lived branches:
+- `main`: This reflects the most recent release of dbt-utils
```
jtcohen6, line 19: I was thinking, "Shouldn't this be `dev/0.x-1.0` (currently `dev/0.5.0`)? But I think having a safe, clear, main branch to PR against for small fixes or README updates is going to be much more user-friendly.

PR #253, opened 2020-07-17
Allow for non-quote identifier column names in pivot even if values include non-alphanumeric symbols

dovstern:
## Checklist
- [ ] I have verified that these changes work locally

### DIFF
--- a/macros/sql/cleanse_jinja.sql
+++ b/macros/sql/cleanse_jinja.sql
@@ -1,0 +1,11 @@
+{% macro cleanse_jinja(text) -%}
+{#
+Clean templated text outputs, so that it is compilable without quote identifiers
+
+Arguments:
+    text: text_to_clean
+#}
+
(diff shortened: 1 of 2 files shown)


clrcrl:
Hey @dovstern — sorry for sitting on this one for so long. I like your thinking here, but I think there might be another way to achieve this that is more generic / broadly useful.

In dbt 0.19.0, we're introducing the `re` module to the Jinja context! I think once that's done, the best way to fix this problem is to introduce a `slugify` macro:

```
{% macro slugify(string) %}

{#- Lower case the string -#}
{% set string = string | lower %}
{#- Replace spaces and dashes with underscores -#}
{% set string = modules.re.sub('[ -]+', '_', string) %}
{#- Only take letters, numbers, and underscores -#}
{% set string = modules.re.sub('[^a-z0-9_]+', '', string) %}

{{ return(string) }}

{% endmacro %}
```

The we could call `slugify` instead of `cleanse_jinja`

I'm going to close this out for now, but I do agree that this is something we want to solve via utils!

PR #258, opened 2020-07-27
Add package and model name in surrogate_key warning

tayloramurphy:
## Description & motivation

This adds the package and model name to the surrogate key deprecation warning.

I tested by directly editing the macro in dbt modules after running deps. Produces:

> Running with dbt=0.17.1
Warning: the `surrogate_key` macro now takes a single list argument instead of multiple string arguments. Support for multiple string arguments will be deprecated in a future release of dbt-utils. The gitlab_snowflake.arr_changes_sfdc_account_monthly model triggered this warning.
Parsing on run start hooks...
Parsing on run end hooks...
Warning: the `surrogate_key` macro now takes a single list argument instead of multiple string arguments. Support for multiple string arguments will be deprecated in a future release of dbt-utils. The logging.stg_dbt_audit_log model triggered this warning.
Warning: the `surrogate_key` macro now takes a single list argument instead of multiple string arguments. Support for multiple string arguments will be deprecated in a future release of dbt-utils. The logging.stg_dbt_model_deployments model triggered this warning.

## Checklist
- [x] I have verified that these changes work locally
- [x] I have updated the README.md (if applicable)
- [x] I have added tests & descriptions to my models (and macros if applicable)

PR #259, opened 2020-07-27
Add package and model name in deprecation warnings (fka #258)

clrcrl:
Opened instead of #258


clrcrl:
I will merge this, but I need to catch up with Jeremy about our new release process :) (see #251 — utils has been a trash fire recently, we're trying to get better about it)

PR #263, opened 2020-07-30
Prepare for 0.5.1 release

clrcrl:
This is a:
- [x] bug fix PR with no breaking changes (please change the base branch to `main`)
- [ ] new functionality
- [ ] a breaking change

## Description & motivation
Update the changelog as part of an 0.5.1 release. Also make our tests more strict, in line with other packages we are updating.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
-## dbt-utils v0.6.0 (unreleased)
+## dbt-utils v0.5.1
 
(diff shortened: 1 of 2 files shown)


clrcrl:
I added `--warn-error` flags, but of course, the integration tests in this package purposely test that these warnings are thrown (e.g. we test to see that the previous format for `surrogate_key` is supported)

PR #264, opened 2020-07-31
`no_time_gap` schema test

hundredwatt:
This is a:
- [x] new functionality
- [ ] a breaking change

## Description & motivation
Add `no_time_gap` schema test 

(Open to suggestions on the name 😉)

This test asserts that the largest gap between records is less than the specified interval. For example, if you have a source table that is loaded once per day, you can use this test to find any gaps larger than 1 day in the data.

Example/Usage:
```yaml
version: 2

models:
  # tests that events exist in every 2 day interval in the past year
  - name: events
    tests:
      - dbt_utils.no_time_gap:
          field: created_at
          datepart: day
          interval: 2
          since: 365
```

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/schema_tests/no_time_gap.sql
+++ b/macros/schema_tests/no_time_gap.sql
@@ -1,0 +1,29 @@
+{% macro test_no_time_gap(model, datepart, interval) %}
+
+{% set column_name = kwargs.get('column_name', kwargs.get('field')) %}
+{% set since = kwargs.get('since', kwargs.get('since')) %}
+
+select
+    count(*) as error_result
+from {{model}} t1
+inner join {{model}} t2 on
+    t2.{{column_name}} = (
+        select
+            MAX(t3.{{column_name}})
+        from {{model}} t3
+        where
+            t3.{{column_name}} < t1.{{column_name}}
+        {% if since %}
+        and t3.{{column_name}} >=
+            {{dbt_utils.dateadd('year', -1, dbt_utils.current_timestamp())}}
+        {% endif %}
+    )
+where
+    {{dbt_utils.datediff('t2.' + column_name, 't1.' + column_name, datepart)}}
+    >= {{interval}}
+{% if since %}
+and t1.{{column_name}} >=
+    {{dbt_utils.dateadd('year', -1, dbt_utils.current_timestamp())}}
+{% endif %}
+
+{% endmacro %}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Fixes
 
 ## Features
+* Add `no_time_gap` schema test
 
 ## Quality of life
 
--- a/README.md
+++ b/README.md
@@ -406,6 +406,32 @@ case we recommend using this test instead.
           - product
 ```
 
+#### no_time_gap ([source](macros/schema_tests/no_time_gap.sql))
+This test asserts that a gap of the specified duration does not exist in a
+date or datetime column. For tables that are loaded incrementally, this test 
+can be used to verify that data loaded for all time intervals.
+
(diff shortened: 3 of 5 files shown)


clrcrl (commented):
I've been thinking about this. What do you think of adding a param to the `mutually_exclusive_ranges` test, with an arg named `exact_gap_width`:

```yml
models:
  # test that age ranges do not overlap
  - name: age_brackets
    tests:
      - dbt_utils.mutually_exclusive_ranges:
          lower_bound_column: min_age
          upper_bound_column: max_age
          gaps: required
          exact_gap_width: interval 1 day

```

Though I guess here, we're testing for `max_gap_width`, and others might want to test for `min_gap_width`, right? 

Maybe we should change gap width to be something like this:

```yml
models:
  # test that age ranges do not overlap
  - name: age_brackets
    tests:
      - dbt_utils.mutually_exclusive_ranges:
          lower_bound_column: min_age
          upper_bound_column: max_age
          gaps: required # note: allowed would be ok here too
          gap_widths:
              min_width: 0
              max_width: 20
```

Then to achieve an exact width, you'd set `min_width` and `max_width` to the same value.


But then I guess that doesn't work with date/time functionality since you need to use the `date_add` functionality to check the gaps. Hmmmmmmmmmmmm

(Yes this is an insight into my brain 😅 )

clrcrl (commented):
I really like the idea of this test — for me, a clearer name would be `sequential_values`.

I have a few other comments about SQL style and how this is tested — I can dig into those in more detail, but I did just want to reach out first to check if you're still interested in working on this, I did let it go stale for quite a while!

Let me know, and then if you're interested I can do a more fine-grained review — otherwise, I am happy to finish this off myself :)

hundredwatt:
@clrcrl yes, I'm still interested. Looking forward to your review :)

By renaming to `sequential_values`, do you see this as being used for other types besides DATETIME?

macros/schema_tests/no_time_gap.sql
```diff
@@ -0,0 +1,29 @@
...
+inner join {{model}} t2 on
+    t2.{{column_name}} = (
+        select
+            MAX(t3.{{column_name}})
+        from {{model}} t3
+        where
+            t3.{{column_name}} < t1.{{column_name}}
+        {% if since %}
+        and t3.{{column_name}} >=
+            {{dbt_utils.dateadd('year', -1, dbt_utils.current_timestamp())}}
```
hundredwatt, line 18: Oops, looks like I mistakenly hardcoded this (and line 22) to 1 year ago rather than a dynamic value based on `since` and `datepart`

clrcrl:
I made the decision here to take the functionality of this PR and rewrite it as a `sequential_values` test, in PR #318. 
Notably:
- the sequential values test works for both time intervals, and integer / decimal gaps
- It does _not_ have the concept of a "since" parameter — instead I'd implement a "min value" test in addition to sequential values

Typically, I try my best to provide feedback on pull requests rather than redo code myself — in this case I went against my typical behavior and wrote my own code. Your code was good, however it was a matter of style — the SQL style is really different to our SQL style, and we tend to use seed files for testing rather than a sequence (which have much nicer cross-db compatibility). Given that I also felt there should be design changes to the test (i.e. making it work with both time intervals and integer / decimal intervals), the back-and-forth of reviewing was likely going to be a frustrating process for both parties as there were a ton of changes I had in mind.

I would love to work with you on a future PR — if you have another test in mind, I encourage you to open an issue and we can work on it earlier on in the process. Again, thank you for this PR — the idea behind the test is really good, and I'm excited to add its sibling to our testing suite!

PR #267, opened 2020-08-15
Feature: use adapter.dispatch

jtcohen6:
This is a:
- [ ] bug fix PR with no breaking changes (please change the base branch to `main`)
- [ ] new functionality
- [x] a breaking change

## Description & motivation
* Replace all instances of `adapter_macro` with `adapter.dispatch`
    * Define a variable, `dbt_utils_dispatch_list`, whereby end users can provide a list of _other_ package namespaces that will dispatch to _before_ `dbt_utils`, whenever a `dbt_utils` macro is called
* While doing that, get tests passing
    * `date_spine` on postgres
    * refactor `get_relations_by_prefix` and `get_relations_by_pattern` since there was a lot of duplicated code, and the tests for the latter were failing on BQ
* We won't be able to run automated tests until v0.18.0rc1 is on pypi

## Checklist
- [x] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [x] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake
- [ ] I have updated the README.md (if applicable)
- [ ] I have added tests & descriptions to my models (and macros if applicable)
- [ ] I have added an entry to the changelog

PR #268, opened 2020-08-20
Remove some deprecated macros

clrcrl:
This is a:
- [ ] new functionality
- [x] a breaking change

## To do before merge
- [x] Update base branch to `dev/0.6.0` once #267 is merged
- [x] Update the changelog

## Description & motivation
I removed `union_tables` and `get_tables_by_prefix`, which we deprecated >6 months ago as part of [v0.2.4](https://github.com/fishtown-analytics/dbt-utils/releases/tag/0.2.4) in Dec 2019

I also removed `get_tables_by_pattern` — this was introduced in v0.5.0 and is functionally equivalent to `get_tables_by_pattern_sql`. Given few people would be using this macro (as it's a helper macro), we are going to go straight to removing it rather than deprecating it first.

We could also consider deprecating:
- The old arguments for `surrogate_key` macro: deprecated in v0.3.0 /  April 2020
- `identifier`: deprecation warning added in  v0.4.0 / April 2020 (deprecated before then but not marked as such)

Personally, I think we should wait until the _next_ version of utils to deprecate those

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/sql/union.sql
+++ b/macros/sql/union.sql
@@ -83,15 +83,3 @@
     {%- endfor -%}
 
 {%- endmacro -%}
-
-{%- macro union_tables(tables, column_override=none, include=[], exclude=[], source_column_name='_dbt_source_table') -%}
-    {%- set error_message = '
-        Warning: the `union_tables` macro is no longer supported and will be deprecated in a future release of dbt-utils. \
-        Use the `union_relations` macro instead. \
-        The {}.{} model triggered this warning. \
-        '.format(model.package_name, model.name) -%}
-    {%- do exceptions.warn(error_message) -%}
-
-    {{ return(dbt_utils.union_relations(tables, column_override, include, exclude, source_column_name)) }}
-
-{%- endmacro -%}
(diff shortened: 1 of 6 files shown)


macros/sql/union.sql
```diff
@@ -10,7 +10,6 @@
     {% endif -%}
 
     {%- set column_override = column_override if column_override is not none else {} -%}
-    {%- set source_column_name = source_column_name if source_column_name is not none else '_dbt_source_relation' -%}
```
clrcrl, line 13: @jtcohen6 — can you just sense check this for me (related to line #1). I looked at it a number of times and can't figure out why it's like this.

I guess similarly, why do we have `column_override=none` instead of `column_override={}`

jtcohen6: I think this is because of the python convention to set default arguments to `none`, always, out of fear of [mutable def](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)[ault arguments](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/).

It's less of an issue with strings and dicts than with arrays, so I think it's _probably_ fine to make this change.

PR #281, opened 2020-09-22
Make star macro case insensitive

mdimercurio:
This is a:
- [ ] new functionality
- [x] a breaking change

## Description & motivation

This change makes the `except` parameter of the `star` macro case insensitive.

We encountered this issue as we are migrating from Postgres to Snowflake and we lowercase all of our column names in the `except` argument.

This could be a breaking change if you have different columns with different casing but this seems like a very bad practice, for example `timestamp` and `Timestamp` in the same table. I haven't included an entry in the changelog yet because I am not sure if this is considered a fix/feature/breaking change.

See discussion here: https://github.com/fishtown-analytics/dbt-utils/issues/115

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
- [ ] I have added an entry to the changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -515,7 +515,7 @@ Usage:
 ```
 
 #### star ([source](macros/sql/star.sql))
-This macro generates a list of all fields that exist in the `from` relation, excluding any fields listed in the `except` argument. The construction is identical to `select * from {{ref('my_model')}}`, replacing star (`*`) with the star macro. This macro also has an optional `relation_alias` argument that will prefix all generated fields with an alias.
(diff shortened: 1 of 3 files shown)


clrcrl:
@drewbanin — you left the original comment about making the `except` parameter case-insensitive. I know it has probably been a while since you thought about it, but do you have any thoughts here?

drewbanin:
hey @mdimercurio - thanks for opening this PR!

Gosh... it has been a while. My sense is still that it would be more convenient to make the `except` argument case-insensitive, although it definitely is "messy" to create subtleties like these ad-hoc across the different macros in dbt-utils.

@clrcrl what do you reckon?

mdimercurio:
@drewbanin do you see other macros that could use this logic?
We could add an argument to enable it to make it more clear to users

clrcrl:
Closing in favor of #348

PR #284, opened 2020-09-25
Schema test/not accepted values

### LINKED ISSUE #283: Schema Test - Not Accepted Values
### Describe the feature
Exactly the opposite of `accepted_values` test. 
A way to test if a value is not inside a field. 

Using `dbt_utils.expression_is_true` could be achieved but it is harder if you have several values and doing a macro that works in a column level could be useful

### Additional context
It should work with any kind of database

### Who will this benefit?
Anyone that wants to check if specific values are not inside a column. 
Some examples:
- A table with users and their roles where I want to ensure that there are no "public" roles.
- A table with some cities, I want to check if some cities are not there

### Are you interested in contributing to this feature?
Yes, I would be more than happy to help by creating a PR with this.

My main doubt creating a PR is related to tests, I think I did one test correctly but I would like to do the test where the feature actually fails. Maybe that can be discussed better in a PR.

JavierMonton:
This is a:
- [x] new functionality
- [ ] a breaking change

## Description & motivation
This is a new macro that will allow us to check `not accepted values`, exactly the opposite of `accepted_values`. 
I'm already using it in a table with users and roles where I want to ensure that some roles are not inside that table.
This will close #283 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [x] Redshift
    - [ ] Snowflake

I have a doubt about testing this kind of macros, writing a test that uses the macro only tests when it works, but not when it fails. Is there any way of testing when it fails?

Also, this is my first PR, I hope I did everything ok 🙂


clrcrl:
Just did a small rebase on your branch so that we don't have merge conflicts!

This is really great — I'm still thinking a little about the name here: is "not accepted" best, or is something like disallowed, or barred, better?

JavierMonton:
When I was looking for this functionality I was mainly searching for "not accepted values", but disallowed or rejected for example could be fine too! I would say that barred seems less familiar to me, but up to you! 
I can change the name if you prefer.

clrcrl:
The more I sat with it, the more I thought "not allowed" felt very descriptive, let's merge!

clrcrl (approved):

PR #289, opened 2020-10-12
Add argument to get_column_values allow for alternate sorting of column values

### LINKED ISSUE #288: The order by clause in get_column_values should be configurable
Currently, the columns values returned by `get_column_values` are sorted by most frequent to least frequent via `order by count(*) desc` [here](
https://github.com/fishtown-analytics/dbt-utils/blob/9feaccc327a7409298a2bc362db53c2e597024fa/macros/sql/get_column_values.sql#L44),  which isn't always what the caller expects. This should be configurable by allowing callers to specify either a column, a sort direction or both.

clausherther:
This is a:
- [x] new functionality
- [ ] a breaking change

## Description & motivation
This PR adds a `sort_column` and `sort_direction` argument to `get_column_values` to allow for alternate sorting of column values. This based on work from @joshpeng-quibi
Closes #288 

## Checklist
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/sql/get_column_values.sql
+++ b/macros/sql/get_column_values.sql
@@ -10,7 +10,7 @@ Returns:
     A list of distinct values for the specified columns
 #}
 
-{% macro get_column_values(table, column, max_records=none, default=none) -%}
+{% macro get_column_values(table, column, sort_column=none, sort_direction=none, max_records=none, default=none) -%}
 
 {#-- Prevent querying of db in parsing mode. This works because this macro does not create any new refs. #}
     {%- if not execute -%}
@@ -22,6 +22,7 @@ Returns:
                                           schema=table.schema,
                                          identifier=table.identifier) -%}
 
+    {# If no sort column is supplied, we use the default descending frequency count. #}
     {%- call statement('get_column_values', fetch_result=true) %}
 
         {%- if not target_relation and default is none -%}
@@ -36,16 +37,27 @@ Returns:
 
         {%- else -%}
 
-            select
-                {{ column }} as value
+            with sorted_column_values as (
 
-            from {{ target_relation }}
-            group by 1
-            order by count(*) desc
+                select
+                    {{ column }} as value,
+                    {%- if sort_column is none %}
+                    count(*) as sort_column
+                    {% else %}
+                    {# We take the max sort value for each value to make sure 
+                        there are no duplicate rows for each value #}
+                    max({{ sort_column }}) as sort_column
+                    {% endif %}
+                from {{ target_relation }}
+                group by 1
+                order by 2 {{ sort_direction if sort_direction else "desc" }}
+                {% endif %}
+                {% if max_records is not none %}
+                limit {{ max_records }}
+                {% endif %}
 
-            {% if max_records is not none %}
-            limit {{ max_records }}
-            {% endif %}
+            )
+            select value from sorted_column_values
 
         {% endif %}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# dbt-utils v0.6.3
+
+## Features
+- Adds ability to specify a `sort_column` and `sort_direction` in `get_column_values.
+
 # dbt-utils v0.6.2
 
 ## Fixes
--- a/README.md
+++ b/README.md
@@ -442,6 +442,8 @@ group by 1
 This macro returns the unique values for a column in a given [relation](https://docs.getdbt.com/docs/writing-code-in-dbt/class-reference/#relation).
 It takes an options `default` argument for compiling when the relation does not already exist.
 
+The `sort_column` argument allows for sorting of values. The default is highest to lowest frequency of values. You can also specify a `sort_direction`.
+
 Usage:
 ```
 -- Returns a list of the top 50 states in the `users` table
@@ -453,6 +455,29 @@ Usage:
 
 ...
 ```
+
+```
+-- Returns a list of user names sorted by name from the `users` table
+{% set names = dbt_utils.get_column_values(table=ref('users'), column='name', sort_column='name', default=[]) %}
+
+{% for name in names %}
+    ...
+{% endfor %}
+
+...
+```
+
+```
+-- Returns a list of user cities sorted by name from the `users` table
+{% set cities = dbt_utils.get_column_values(table=ref('users'), column='city_name', sort_column='city_name', default=[]) %}
+
+{% for city in cities %}
+    ...
+{% endfor %}
+
+...
+```
+
 #### get_relations_by_prefix
 Returns a list of [Relations](https://docs.getdbt.com/docs/writing-code-in-dbt/class-reference/#relation)
 that match a given prefix, with an optional exclusion pattern. It's particularly


shedd:
@clausherther thanks for implementing! was looking for this exact feature just now!

macros/sql/get_column_values.sql
```diff
@@ -36,16 +37,27 @@ Returns:
...
+                    count(*) as sort_column
+                    {% else %}
+                    {# We take the max sort value for each value to make sure 
+                        there are no duplicate rows for each value #}
+                    max({{ sort_column }}) as sort_column
+                    {% endif %}
+                from {{ target_relation }}
+                group by 1
+                order by 2 {{ sort_direction if sort_direction else "desc" }}
+                {% endif %}
```
shedd, line 54: fyi - I created a local copy of this and hit and error on DBT v18:

```
Encountered an error:
Compilation Error in macro (macros/sorted_get_column_values.sql)
  Got an unexpected control flow end tag, got endif but never saw a preceeding if (@ 63:8)
```

clausherther: Thanks for highlighting that! Just pushed a fix to remove that extra `endif`

clausherther:
@clrcrl anything else we need to do here to get this merged?

macros/sql/get_column_values.sql
```diff
@@ -36,16 +37,26 @@ Returns:
...
+            with sorted_column_values as (
+
+                select
+                    {{ column }} as value,
+                    {%- if sort_column is none %}
+                    count(*) as sort_column
+                    {% else %}
+                    {# We take the max sort value for each value to make sure 
+                        there are no duplicate rows for each value #}
+                    max({{ sort_column }}) as sort_column
```
clrcrl, line 49: huh, this is interesting. I thought this PR would be to make it easier to alpha-sort, i.e. "order by {{ column }}" (which I typically achieve through a Jinja sort)

What kind of ways do you want to sort these results?

clausherther: This is intended to let you sort by **frequency** in **ascending** (or descending) order, which I think was was not possible previously and is not possible in Jinja with just the results since the counts aren't returned. 

(This is based on an internal version of this we had to implement at Quibi as a result.)

macros/sql/get_column_values.sql
```diff
@@ -10,7 +10,7 @@ Returns:
     A list of distinct values for the specified columns
 #}
 
-{% macro get_column_values(table, column, max_records=none, default=none) -%}
+{% macro get_column_values(table, column, sort_column=none, sort_direction=none, max_records=none, default=none) -%}
```
clrcrl, line 13: I had a comment here but it disappeared:

Let's make these new args at the end of the list, in case anyone was relying on order of arguments when previously using this macro 😬 
```suggestion
{% macro get_column_values(table, column, max_records=none, default=none, sort_column=none, sort_direction=none) -%}
```

macros/sql/get_column_values.sql
clrcrl, line 49: What do you think of instead doing something like:
```
{% set order_by = kwargs.get('order_by', 'count(*)')
```
Then people could also order alphabetically by providing:
```
{% set states = dbt_utils.get_column_values(table=ref('users'), column='state', order_by='state') %}

```

And by recency like so:
```
{% set states = dbt_utils.get_column_values(table=ref('users'), column='state', order_by='min(created_at)') %}
```

macros/sql/get_column_values.sql
```diff
@@ -36,16 +37,26 @@ Returns:
...
+                    {%- if sort_column is none %}
+                    count(*) as sort_column
+                    {% else %}
+                    {# We take the max sort value for each value to make sure 
+                        there are no duplicate rows for each value #}
+                    max({{ sort_column }}) as sort_column
+                    {% endif %}
+                from {{ target_relation }}
+                group by 1
+                order by 2 {{ sort_direction if sort_direction else "desc" }}
```
clrcrl, line 53: If we go with the above comment, it may be easier to do something like this:
```suggestion
                order by {{ order_by }} {{ sort_direction or "desc" }}
```

macros/sql/get_column_values.sql
clausherther: I like it! I'll push a commit in a sec. Thanks!

macros/sql/get_column_values.sql
clausherther: Still trying to figure out how to make this work with non-aggregate columns. E.g. you pass in `column_2` instead `count(*)` we need to wrap in an aggregation, like `max(column_2)`, or group by it as well.

clausherther:
@clrcrl lmk what you think of the way I implemented the `order_by` arg. Basically, if we're ordering by the `column`, I use `max` to force the aggregation to be 1:1, otherwise we use `count(*)` or whatever the called specified (e.g. `max(created_at)`

clrcrl (approved):
I played around with this locally:

Here's my testing file:
```
{% set list_default_order = dbt_utils.get_column_values(
        table=ref('stg_payments'),
        column='payment_method')
%}

{% do log('list_default_order: ' ~ list_default_order, info=True) %}

{% set list_alpha_order = dbt_utils.get_column_values(
        table=ref('stg_payments'),
        column='payment_method',
        order_by='payment_method')
%}

{% do log('list_alpha_order: ' ~ list_alpha_order, info=True) %}

{% set list_recency_order = dbt_utils.get_column_values(
        table=ref('stg_payments'),
        column='payment_method',
        order_by='max(created_at)',
        sort_direction='desc')
%}

{% do log('list_recency_order: ' ~ list_recency_order, info=True) %}
```
It all worked as expected!

```txt
list_default_order: ['credit_card', 'bank_transfer', 'coupon', 'gift_card']
list_alpha_order: ['bank_transfer', 'coupon', 'credit_card', 'gift_card']
list_recency_order: ['credit_card', 'bank_transfer', 'gift_card', 'coupon']
```

I did have a slightly different implementation in mind, which I think makes the sort order a little less confusing for the two cases — lmk what you think:

```sql
    {%- set order_by = order_by or 'count(*) desc' -%}
...
            with sorted_column_values as (

                select
                    {{ column }} as value -- note: no second column declared
                from {{ target_relation }}
                group by 1
                order by {{ order_by }}
                {% if max_records is not none %}
                
                limit {{ max_records }}
                {% endif %}
            )
            select value from sorted_column_values

```
By removing the second column in the query, we don't have to get tricky with the `order_by` column. I also just ended up 
removing the `sort_direction` input as I figured it could be pretty easily packaged up in the `order_by` definition.

The above implementations would end up looking like so (i.e. no "sort_direction" parameter)
```
{% set list_default_order = dbt_utils.get_column_values(
        table=ref('stg_payments'),
        column='payment_method')
%}

{% do log('list_default_order: ' ~ list_default_order, info=True) %}

{% set list_alpha_order = dbt_utils.get_column_values(
        table=ref('stg_payments'),
        column='payment_method',
        order_by='payment_method')
%}

{% do log('list_alpha_order: ' ~ list_alpha_order, info=True) %}

{% set list_recency_order = dbt_utils.get_column_values(
        table=ref('stg_payments'),
        column='payment_method',
        order_by='max(created_at) desc'
%}

{% do log('list_recency_order: ' ~ list_recency_order, info=True) %}
```
This does mean there's fewer built-in defaults, but may be a little easier to document and reason about?

PR #292, opened 2020-11-05
Add consecutive date ranges test

MarkMacArdle:
This is a:
- [x] new functionality
- [ ] a breaking change

## Description & motivation
Modified the `dbt_utils.mutually_exclusive_ranges` test to accept an additional argument for the `gaps` parameter of `consecutive_dates`. It tests that date ranges are not overlapping and that the start the date of a row is the day after the previous row's end date. Unlike the other options of `not_allowed, allowed, required`, `consecutive_dates` can only be used when the bound columns are dates.

It's kind of a combination of the existing `gaps: not_allowed` (lower bound of a row must equal upper bound of previous row) and `gaps: required` (lower bound of row must be after upper bound of previous) as a gap is required but it must always be one day. 

The motivation for me is that I get manually created tables, predominantly from our finance team, which look like the below. Need to be able to test the date ranges don't skip any days, as well as that they don't overlap.

| currency | start_date | end_date   | exchange_rate | 
|----------|------------|------------|---------------|
| SEK      | 2020-01-01 | 2020-03-16 | 12.123        |
| SEK      | 2020-03-17 | 2020-05-10 | 12.456        |
| SEK      | 2020-05-11 |            | 13.789        |

## Checklist
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake


clrcrl:
So I wrote up a huge long comment about this, and then put it on the wrong PR 🤦‍♀️ 

Here it is!

I really like the idea of this PR! I think we can make the args just a little tidier though! Here's some variations on this that I could imagine:

1. Testing for a "fixed width" gap, that works somehow for both integers and dateparts:
```yml
  - name: subscriptions
    tests:
      - dbt_utils.mutually_exclusive_ranges:
          lower_bound_column: started_at
          upper_bound_column: ended_at
          gaps:
             interval: 1
             datepart: day
```
(note: I just implemented some similar logic, that handles numeric and date gaps, [over here](https://github.com/fishtown-analytics/dbt-utils/commit/9843b9f31ce4763399941e8415223ec31e375479))

2. Testing for a min / max gap width:

```yml
  - name: subscriptions
    tests:
      - dbt_utils.mutually_exclusive_ranges:
          lower_bound_column: started_at
          upper_bound_column: ended_at
          gaps:
             min:
                 interval: 1
                datepart: day
             max:
                interval: 4
                datepart: day
```

(and combinations of just a min, or just a max, and then is in an exclusive or inclusive range?!)

I think your idea is right — we _just_ want to test for some fixed width. I can't think of a data source where I would want to test that the gap is within some range, but I can think of scenarios (like this one) where I want to test that that gap is an exact  number.

That being said, I think it would be nice if we do make the gap size configurable — is this something you're interested in taking on, or would you prefer that I push some commits to implement that logic?

MarkMacArdle:
I can't currently think of where I'd need a max/min interval either so decided against adding those options now. I've updated the argument naming to be amenable to adding them later though.

```
  - name: datetime_buckets
    tests:
      - dbt_utils.mutually_exclusive_ranges:
          lower_bound_column: valid_from
          upper_bound_column: coalesce(valid_to, '2099-01-01 00:00:00')
          partition_by: subscription_id
          gaps: interval_constrained
          gap_interval_constraints:
              fixed_interval: 2
              datepart: hour
```

Currently the only interval argument `gap_interval_constraints` takes is `fixed_interval`. My thinking is that in future args like `min_interval` and `max_interval` could be added (along with an if statement to block them being used with `fixed_interval`).

It now handles date parts like the example you linked too, thanks for that suggestion!

clrcrl:
Hey Mark, I'm going to admit defeat here. I've really stared at this code for a long time, and in the end, I think this is a somewhat bespoke use case that it is probably going to make the utils code harder for most people to understand, without benefitting many users.

This is something that I think we should have been more decisive on years ago — we've let a lot of "yeah why not?" code into utils, where people have adapted code to suit their specific need, and now we're in this state where it's super hard to maintain and document.

I think that if this schema test enabled you to write your own macro that achieved _exactly_ what you wanted it to do, then it has achieved its purpose. I love that your instinct was to contribute this code back, and I hate that my eventual answer here is going to be "utils is not the right fit". What we need (as a community) is a way for people to share code more easily that doesn't look like changing one of our most-installed packages. This is something we're definitely thinking hard about!

Thank you so much for taking the time to write this code — if nothing else, I hope that you feel that you got to learn some new Jinja along the way and flex some muscles that are pretty fun to flex! I apologise for the fact that I took _so long_ to resolve this, that's fully on me. Find me on dbt Slack and I'll send you some stickers as an apology 💕

PR #295, opened 2020-11-14
Added `where` parameter to the `unique_combination_of_columns` schema test

findinpath:
This is a:
- [x] new functionality
- [ ] a breaking change

## Description & motivation
This PR adds `where` parameter to the `unique_combination_of_columns` schema test in order to support a `WHERE` clause that can be used to isolate the tuples on which the uniqueness for the combination of columns needs to be verified.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake

### DIFF
--- a/macros/schema_tests/unique_combination_of_columns.sql
+++ b/macros/schema_tests/unique_combination_of_columns.sql
@@ -1,12 +1,11 @@
-{% macro test_unique_combination_of_columns(model, quote_columns = false) %}
+{% macro test_unique_combination_of_columns(model, combination_of_columns, quote_columns = false, where = None) %}
 
-{%- set columns = kwargs.get('combination_of_columns', kwargs.get('arg')) %}
 
 {% if not quote_columns %}
-    {%- set column_list=columns %}
+    {%- set column_list=combination_of_columns %}
 {% elif quote_columns %}
     {%- set column_list=[] %}
-        {% for column in columns -%}
+        {% for column in combination_of_columns -%}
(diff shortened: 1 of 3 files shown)


macros/schema_tests/unique_combination_of_columns.sql
```diff
@@ -1,6 +1,7 @@
 {% macro test_unique_combination_of_columns(model, quote_columns = false) %}
 
 {%- set columns = kwargs.get('combination_of_columns', kwargs.get('arg')) %}
+{%- set where = kwargs.get('where', kwargs.get('arg')) %}
```
clrcrl, line 4: ok so this one is on me — we don't actually need this line (or the above one tbh).

This line was because I was copying syntax from dbt core that handled the "v1" yaml syntax, which we haven't supported for a long time now!

Can you:
1. Remove lines 3 & 4
2. Add `where` as an arg in line 1 (`where=None`)

clrcrl (changes requested):
Thanks so much for this PR!

Small change requested.

Can you also:
1. Rebase from `dev/0.7.0`
2. Add an entry to `changelog.md`

(If those instructions don't make sense, lmk, and I can add more details!)

macros/schema_tests/unique_combination_of_columns.sql
findinpath: Sure.

clrcrl (approved):
Looks great!

PR #296, opened 2020-11-16
concat macro can use alternative_concat on Postgres too

ChristopheDuong:
This is a:
- [x] new functionality
- [ ] a breaking change

## Description & motivation
Using Postgres database, I have a model that needs to produce a surrogate key or hash on a large number of columns.
Unfortunately it breaks because, on postgres, we cannot pass more than 100 arguments to a function...

The current `dbt_utils.surrogate_key` is indeed calling the `dbt_utils.concat` which is using the `concat()` function that accepts a list of arguments.

However postgres has another alternative to concat string using the binary operator `||` instead as we can do on Redshift and on Snowflake and adding this adapter version to the dispatched function for the `dbt_utils.concat` would avoir crashes when passing more than 100 strings to concat together.

## Checklist
    - [ ] BigQuery
    - [X] Postgres
    - [ ] Redshift
    - [ ] Snowflake
- [ ] I have added an entry to the changelog


clrcrl (commented):
So it turns out that [Postgres has supported this for a long time](https://www.postgresql.org/docs/7.1/functions-string.html), and BigQuery has been supporting it [since March](https://twitter.com/bbhoss/status/1235721909230133248)

I think we should just rip out the "alternative" syntax and make `||` the default. Though, @jtcohen6, I'm curious whether you think that might break any other non-core adapters (e.g. does Spark support `||`?)

PR #298, opened 2020-11-18
Add all_columns_present_in_both_tables argument to equality test

MarkMacArdle:
This is a:
- [x] new functionality
- [ ] a breaking change

## Description & motivation
I want to use the equality test to make sure two tables have all the same columns and values. Equality will check all the values but if the table the test is called on (table_a) has less columns than the table compared against (table_b), this won't be cause a test to fail (although the reverse, table_b having less columns than table_a, would cause the test to fail).

Added the `all_columns_present_in_both_tables` argument which expects a `true` or `false` value to allow testing for this.

The equality test also has the `compare_columns` argument which allows comparing only a subset of the columns in the tables. It doesn't make sense to use this argument with `all_columns_present_in_both_tables` so added an if statement to the start of the equality test that will throw a compile error if both are used at the same time.

## Checklist
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# dbt-utils v0.6.3
+
+## Features
```
clrcrl, line 3: ok so this is on us — we changed how we're using branches in dbt-utils and forgot to update the PR template.

I've just switched the base branch for this to be `dev/0.7.0` — can you rebase this work from that branch, and add this entry into the `dev/0.7.0` changelog section?

macros/schema_tests/equality.sql
```diff
@@ -18,8 +25,15 @@ information schema — this allows the model to be an ephemeral model
...
-{% set compare_columns = kwargs.get('compare_columns', adapter.get_columns_in_relation(model) | map(attribute='quoted') ) %}
-{% set compare_cols_csv = compare_columns | join(', ') %}
+{% set compare_cols_a = kwargs.get('compare_columns', adapter.get_columns_in_relation(model) | map(attribute='quoted') ) %}
+{% set compare_cols_a_csv = compare_cols_a | join(', ') %}
+
+{%- if kwargs.get('all_columns_present_in_both_tables', false) -%}
+    {% set compare_cols_b = kwargs.get('compare_columns', adapter.get_columns_in_relation(compare_model) | map(attribute='quoted') ) %}
+    {% set compare_cols_b_csv = compare_cols_b | join(', ') %}
+{%- else -%}
+    {% set compare_cols_b_csv = compare_cols_a_csv %}
```
clrcrl, line 35: So this will fail if the columns are in a different order, right?:
| id | name   | 
|----|--------| 
| 1  | claire | 

| name | id |
|--|--|
| claire | 1 |

I'd  posit that these two tables are in fact equal though — curious for your thoughts here!

clrcrl (changes requested):

clrcrl:
Thanks so much for this PR! Agree that this is a change is the right direction! I've left some comments, let me know if they make sense or not :)

macros/schema_tests/equality.sql
MarkMacArdle: Yes order would cause the test to fail, but I would agree those two are equal in content. I think rather than just the `all_columns_present_in_both_tables` variable having two would be more appropriate:
```
identical_columns: true|false
identical_column_order: true|false
```

The case I currently want to use this test for is a query that must match a report format. So I want to test both content and column order. Appreciate not everyone would care about the column order though.

clrcrl: I spent way too long thinking about this this morning

I think if we were implementing the macro from scratch, we'd do something to be smarter about how the choose which columns to compare, and have something like:
```
comparison_type is one of ['all_columns_ordered', 'all_columns_unordered', 'columns_present_in_both', 'user_specifed']
```
Alas, as it stands, it's a little hard to change the default behavior of a utils macro, as it will cause breaking changes.

I think your proposal is a good one, and adding `identical_columns` and `identical_column_order` args with defaults of false (so as not to introduce breaking changes) is the way to go!

MarkMacArdle:
I added a further test for column order... and then kept going and also added datatype and case sensitivity tests as I realised you can check both of those too once you have the information schema data.

Once I realised all the tests I was adding were information schema tests I thought it made sense to group them into one `column_metadata_tests` list. They're all optional.

```yaml
 - name: model_name
    tests:
      - dbt_utils.equality:
          compare_model: ref('other_table_name')
          column_metadata_tests:
            - all_columns_present_in_both_tables
            - case_sensitive_names
            - matching_order
            - matching_data_types
```

Aside from `all_columns_present_in_both_tables` any/all of the other tests can be used in conjunction with the existing `compare_columns` argument.

```yaml
  - name: model_name
    tests:
      - dbt_utils.equality:
          compare_model: ref('other_table_name')
          compare_columns:
            - first_column
            - second_column
          column_metadata_tests:
            - case_sensitive_names
            - matching_order
            - matching_data_types
```

The downside to all this functionality is a big increase in complexity of the test so I have added testing. I would have liked to be able have tests that check it raises compile errors when it should but didn't see a way of doing this. I wished there was an equivalent of try/except in python. 

I did check it raises those compile errors when it should in my manual testing and just left those tests in the `schema.yml` but commented them out.

clrcrl:
Hey Mark — sorry for the huge delay here. This PR ended up being much more complex than I expected. I tried to review it today, but ultimately when I weighed the effort of reviewing it against the benefit that the changes would bring, I came up short. We tend to use the `equality` test for meta-testing (i.e. in the `integration_tests` for dbt-utils), rather than use it to test models in our project (it doesn't make a ton of sense that we'd have two models with the same output), so these don't feel like changes that a _ton_ of people would benefit from.

I have a sneaking suspicion that you've been using this macro to implement unit testing in your dbt project — that's a valiant effort, and a use case we've thought about in dbt-core. While we don't have plans to immediately tackle it, it is something that we want to tackle as a core-paradigm in the future.

PR #310, opened 2020-12-23
t-sql hacks that shouldn't break mainline behavior

swanderz:
This is a:

## Description & motivation
MSFT's T-SQL has some weird quirks to it such as:
- subquery aggregate columns need aliases; and
- there is no boolean type, so there's no such thing as `true` and `false`. So in a few schema tests, I've replaced `true` with `1=1`; and,
- `limit 0` at the end of the query translates in TSQL to `SELECT TOP (0)`. If the purpose is to return zero rows, `WHERE 0=1` will accomplish the same more universally.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/integration_tests/tests/assert_get_query_results_as_dict_objects_equal.sql
+++ b/integration_tests/tests/assert_get_query_results_as_dict_objects_equal.sql
@@ -77,5 +77,5 @@ Instead, we'll manually check that the values of these dictionaries are equivale
 {% endif %}
 
 {{ log(ns.err_msg, info=True) }}
-select 1 as col_name {% if ns.pass %} where 0=1 {% endif %}
+select 1 as col_name {% if ns.pass %} {{ limit_zero }} {% endif %}
 {% endif %}
--- a/integration_tests/tests/assert_get_query_results_as_dict_objects_equal.sql
+++ b/integration_tests/tests/assert_get_query_results_as_dict_objects_equal.sql
@@ -77,5 +77,5 @@ Instead, we'll manually check that the values of these dictionaries are equivale
 {% endif %}
 
 {{ log(ns.err_msg, info=True) }}
-select 1 {% if ns.pass %} limit 0 {% endif %}
+select 1 as col_name {% if ns.pass %} where 0=1 {% endif %}
 {% endif %}
--- a/integration_tests/tests/assert_get_query_results_as_dict_objects_equal.sql
+++ b/integration_tests/tests/assert_get_query_results_as_dict_objects_equal.sql
@@ -77,5 +77,5 @@ Instead, we'll manually check that the values of these dictionaries are equivale
 {% endif %}
 
 {{ log(ns.err_msg, info=True) }}
-select 1 as col_name {% if ns.pass %} {{ limit_zero }} {% endif %}
+select 1 as col_name {% if ns.pass %} {{ limit_zero() }} {% endif %}
 {% endif %}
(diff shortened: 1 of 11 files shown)


clrcrl:
@jtcohen6 — what do you think of this PR? IIRC we incorporated some similar logic into dbt core since it's non-breaking, but I also know we say "create a shim package" for other adapter functionality

jtcohen6 (commented):
For precedent's sake, what I'd rather do here is add `dispatch` functionality to each of these macros so that they can be rewritten/overridden in a shim package. I.e. replace
```
{% macro test_at_least_one(model) %}
```
with
```
{% macro test_at_least_one(model) %}
  {{ adapter.dispatch('test_at_least_one', packages = dbt_utils._get_utils_namespaces())(model) }}
{% endmacro %}

{% macro default__test_at_least_one(model) %}
```

For that matter, we should support dispatching for every macro defined in the package, even if all of Postgres/Redshift/Snowflake/BigQuery are happy with the `default__` implementation.

Edit: as in #312, which I see @swanderz also opened!

integration_tests/tests/assert_get_query_results_as_dict_objects_equal.sql
```diff
@@ -77,5 +77,5 @@ Instead, we'll manually check that the values of these dictionaries are equivale
 {% endif %}
 
 {{ log(ns.err_msg, info=True) }}
-select 1 {% if ns.pass %} limit 0 {% endif %}
+select 1 as col_name {% if ns.pass %} where 0=1 {% endif %}
```
swanderz, line 80: ouch. looks at if TSQL and BQ don't play nice together. From CircleCI

> Query without FROM clause cannot have a WHERE clause

jtcohen6: Gross. Perhaps we do need a dispatched macro `{{ limit_zero() }}` after all, just for the sake of integration tests

swanderz: where would `limit_zero()` live? as a macro just for integration tests, `integration_tests/macros`? could it be properly dispatched there?

jtcohen6: Yes, it could live in `integration_tests/macros`, and look like:
```
{% macro limit_zero() %}
    {{ return(adapter.dispatch('limit_zero', dbt_utils._get_utils_namespaces())()) }}
{% endmacro %}

{% macro default__limit_zero() %}
    {{ return('limit 10') }}
{% endmacro %}
```
i.e. It dispatches via the `dbt_utils_dispatch_list` var for namespace purposes, but it doesn't actually "live" in / ship with the `dbt_utils` package.

Then, you should be able to define:
```
{% macro sqlserver__limit_zero() %}
  {{ return('where 0=1') }} 
{% endmacro %}
```

This is the same thing as, or very similar to, what we did with [`prep_external`](https://github.com/fishtown-analytics/dbt-external-tables/blob/master/integration_tests/macros/common/prep_external.sql) over in the external tables package

jtcohen6:
@swanderz As for the current error, I think you just need to add the following line [here](https://github.com/fishtown-analytics/dbt-utils/blob/master/integration_tests/dbt_project.yml):
```yml
vars:
  dbt_utils_dispatch_list: ['dbt_utils_integration_tests']
```

jtcohen6 (approved):
looks good! could you add a changelog entry to this one as well?

swanderz:
> looks good! could you add a changelog entry to this one as well?

done! stoked to have both these merged! game changer!

Issue #311, opened 2020-12-23
'dispatch'-ize the schema test matcros


swanderz:
### Describe the bug
per [this discussion](https://github.com/dbt-msft/tsql-utils/pull/6#discussion_r547601190) with @jtcohen6, we'll need make the schema test macros make use of `adapter.dispatch()` so that `tsql-utils` can customize the schema tests.

### Are you interested in contributing the fix?

Yep, but might not get to it for a few weeks due to holidays & PTO

jtcohen6:
Ultimately, I think we should add `dispatch` to every macro in the package (:

swanderz:
#312 closes this out!

PR #279, opened 2020-09-16
Remove errant default args from get_tables_by_pattern_sql

clrcrl:
This is a:
- [x] bug fix PR with no breaking changes (please change the base branch to `main`)
- [ ] new functionality
- [ ] a breaking change

## Description & motivation
Turns out that this line forces the argument to be the default value.

Closes [this issue on codegen](https://github.com/fishtown-analytics/dbt-codegen/issues/33)

## Testing
I updated my local version of codegen to also use my local version of utils, and tested locally.

I also did a quick `cmf + f` to find any other instances of this, and couldn't see anything — would definitely appreciate a second set of eyes there

## Checklist
- [ ] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
- [ ] I have updated the README.md (if applicable)
- [ ] I have added tests & descriptions to my models (and macros if applicable)
- [ ] I have added an entry to the changelog

PR #312, opened 2020-12-23
dispatch-ify ALL THE MACROS!

swanderz:
This is a:

## Description & motivation
see #311 
cc: @alittlesliceoftom

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Speed up CI via threads, workflows ([#315](https://github.com/fishtown-analytics/dbt-utils/pull/315), [#316](https://github.com/fishtown-analytics/dbt-utils/pull/316))
 - Fix `equality` test when used with ephemeral models + explicit column set ([#321](https://github.com/fishtown-analytics/dbt-utils/pull/321))
 - Fix `get_query_results_as_dict` integration test with consistent ordering ([#322](https://github.com/fishtown-analytics/dbt-utils/pull/322))
+- All macros are now properly dispatched, making it possible for non-core adapters to implement a shim package for dbt-utils ([#312](https://github.com/fishtown-analytics/dbt-utils/pull/312)) Thanks [@chaerinlee1](https://github.com/chaerinlee1) and [@swanderz](https://github.com/swanderz)
 
 # dbt-utils v0.6.3
 
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -19,6 +19,9 @@ clean-targets:         # directories to be removed by `dbt clean`
     - "target"
     - "dbt_modules"
 
+vars:
+  dbt_utils_dispatch_list: ['dbt_utils']
+
 seeds:
 
   +quote_columns: false
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -19,9 +19,6 @@ clean-targets:         # directories to be removed by `dbt clean`
     - "target"
     - "dbt_modules"
 
-vars:
-  dbt_utils_dispatch_list: ['dbt_utils']
-
 seeds:
 
   +quote_columns: false
--- a/integration_tests/tests/logger/assert_pretty_output_msg_is_string.sql
+++ b/integration_tests/tests/logger/assert_pretty_output_msg_is_string.sql
@@ -1,4 +1,4 @@
-{% if dbt_utils.pretty_log_format() is string %}
+{% if pretty_log_format() is string %}
     {# Return 0 rows for the test to pass #}
     select 1 limit 0
 {% else %}
--- a/integration_tests/tests/logger/assert_pretty_output_msg_is_string.sql
+++ b/integration_tests/tests/logger/assert_pretty_output_msg_is_string.sql
@@ -1,4 +1,4 @@
-{% if pretty_log_format() is string %}
+{% if dbt_utils.pretty_log_format() is string %}
     {# Return 0 rows for the test to pass #}
     select 1 limit 0
 {% else %}
--- a/macros/schema_tests/at_least_one.sql
+++ b/macros/schema_tests/at_least_one.sql
@@ -1,4 +1,8 @@
 {% macro test_at_least_one(model) %}
+  {{ adapter.dispatch('test_at_least_one', packages = dbt_utils._get_utils_namespaces())(model) }}
+{% endmacro %}
+
+{% macro default__test_at_least_one(model) %}
 
 {% set column_name = kwargs.get('column_name', kwargs.get('arg')) %}
 
--- a/macros/schema_tests/at_least_one.sql
+++ b/macros/schema_tests/at_least_one.sql
@@ -1,5 +1,5 @@
 {% macro test_at_least_one(model) %}
-  {{ adapter.dispatch('test_at_least_one', packages = dbt_utils._get_utils_namespaces())(model, **kwargs) }}
+  {{ return(adapter.dispatch('test_at_least_one', packages = dbt_utils._get_utils_namespaces())(model, **kwargs)) }}
 {% endmacro %}
 
 {% macro default__test_at_least_one(model) %}
--- a/macros/schema_tests/at_least_one.sql
+++ b/macros/schema_tests/at_least_one.sql
@@ -1,5 +1,5 @@
 {% macro test_at_least_one(model) %}
-  {{ adapter.dispatch('test_at_least_one', packages = dbt_utils._get_utils_namespaces())(model) }}
+  {{ adapter.dispatch('test_at_least_one', packages = dbt_utils._get_utils_namespaces())(model, **kwargs) }}
 {% endmacro %}
 
 {% macro default__test_at_least_one(model) %}
--- a/macros/schema_tests/expression_is_true.sql
+++ b/macros/schema_tests/expression_is_true.sql
@@ -1,5 +1,5 @@
 {% macro test_expression_is_true(model, condition='true') %}
-  {{ adapter.dispatch('test_expression_is_true', packages = dbt_utils._get_utils_namespaces())(model, condition='true') }}
+  {{ adapter.dispatch('test_expression_is_true', packages = dbt_utils._get_utils_namespaces())(model, condition='true', **kwargs) }}
 {% endmacro %}
 
 {% macro default__test_expression_is_true(model, condition='true') %}
--- a/macros/schema_tests/expression_is_true.sql
+++ b/macros/schema_tests/expression_is_true.sql
@@ -1,5 +1,5 @@
 {% macro test_expression_is_true(model, condition='true') %}
-  {{ adapter.dispatch('test_expression_is_true', packages = dbt_utils._get_utils_namespaces())(model, condition='true', **kwargs) }}
+  {{ return(adapter.dispatch('test_expression_is_true', packages = dbt_utils._get_utils_namespaces())(model, condition='true', **kwargs)) }}
 {% endmacro %}
 
 {% macro default__test_expression_is_true(model, condition='true') %}
--- a/macros/schema_tests/expression_is_true.sql
+++ b/macros/schema_tests/expression_is_true.sql
@@ -1,5 +1,5 @@
 {% macro test_expression_is_true(model, condition='true') %}
-  {{ return(adapter.dispatch('test_expression_is_true', packages = dbt_utils._get_utils_namespaces())(model, condition='true', **kwargs)) }}
+  {{ return(adapter.dispatch('test_expression_is_true', packages = dbt_utils._get_utils_namespaces())(model, condition, **kwargs)) }}
 {% endmacro %}
 
 {% macro default__test_expression_is_true(model, condition='true') %}
--- a/macros/schema_tests/expression_is_true.sql
+++ b/macros/schema_tests/expression_is_true.sql
@@ -1,4 +1,8 @@
 {% macro test_expression_is_true(model, condition='true') %}
+  {{ adapter.dispatch('test_expression_is_true', packages = dbt_utils._get_utils_namespaces())(model, condition='true') }}
+{% endmacro %}
+
+{% macro default__test_expression_is_true(model, condition='true') %}
 
 {% set expression = kwargs.get('expression', kwargs.get('arg')) %}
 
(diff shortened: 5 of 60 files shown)


swanderz:
@jtcohen6 not sure what I've done here...
```log
Encountered an error:
Compilation Error in test dbt_utils_equality_test_generate_series_ref_data_generate_series_ (models/sql/schema.yml)
  macro 'dbt_macro__test_equality' takes no keyword argument 'arg'
  
  > in macro test_equality (macros/schema_tests/equality.sql)
  > called by test dbt_utils_equality_test_generate_series_ref_data_generate_series_ (models/sql/schema.yml)
```

macros/schema_tests/at_least_one.sql
```diff
@@ -1,4 +1,8 @@
 {% macro test_at_least_one(model) %}
+  {{ adapter.dispatch('test_at_least_one', packages = dbt_utils._get_utils_namespaces())(model) }}
```
jtcohen6, line 2: This or something like it:
```suggestion
  {{ adapter.dispatch('test_at_least_one', packages = dbt_utils._get_utils_namespaces())(model, **kwargs) }}
```

jtcohen6 (commented):
Nice PR! I think this is the way to go.

You need to allow for passing `kwargs` through the macro dispatcher. It's a bit tricky.

clrcrl:
Removing myself as reviewer, as Jeremy has it covered :)

integration_tests/tests/logger/assert_pretty_output_msg_is_string.sql
```diff
@@ -1,4 +1,4 @@
-{% if dbt_utils.pretty_log_format() is string %}
+{% if pretty_log_format() is string %}
     {# Return 0 rows for the test to pass #}
     select 1 limit 0
 {% else %}
```
swanderz, line 4: @jtcohen6 thought we'd be extra and help make ALL macros use dispatch... but will this break any references to `dbt_utils.some_macro()` in other dbt packages? for example:
- [dbt-external-tables](https://github.com/fishtown-analytics/dbt-external-tables/blob/a3c78be7bceaf76329fedc149ce8d34d734eebfb/macros/common/stage_external_sources.sql#L44)
- [dbt's integration tests ](https://github.com/fishtown-analytics/dbt/blob/28c36cc5e211f7c41bd5ca177c9253183c697097/test/integration/100_rpc_test/deps_models/main.sql#L1)

jtcohen6: You want to keep this as `dbt_utils.pretty_log_format()`!

If the macro call is just `pretty_log_format`, dbt will only look for a macro called `pretty_log_format` in (a) the user's own project or (b) the default (`dbt`) namespace. Macros from installed packages need to be fully qualified with their package name. With the addition of `dispatch` + the `dbt_utils_dispatch_list` var, _another_ package can interject when dbt goes to implement the fully-qualified macro `dbt_utils.pretty_log_format()`.

It's a bit tricky!

swanderz: I _think_ I understand. Perhaps I've been spared having to learn this from primarily working with adapters instead of packages. very much appreciate you coaching me through all of this, I'll revert now.

jtcohen6: Yup, definitely not saying it's a good thing :( FWIW `dbt-utils` is where it feels trickiest

swanderz:
onto the next thing... I don't have a postgres target handy (but can get one running without too much trouble probably). @jtcohen6 any shoves in the right direction here? I haven't the slightest how the dispatch addition would make a test fail...
```log
Failure in test dbt_utils_expression_is_true_data_test_expression_is_true_col_b_0_5__col_a_0_5 (models/schema_tests/schema.yml)
  Got 2 results, expected 0.

  compiled SQL at target/compiled/dbt_utils_integration_tests/models/schema_tests/schema.yml/schema_test/dbt_utils_expression_is_true_data_test_expression_is_true_col_b_0_5__col_a_0_5.sql
```

macros/sql/generate_series.sql
```diff
@@ -1,4 +1,8 @@
 {% macro get_powers_of_two(upper_bound) %}
+    {{ adapter.dispatch('get_powers_of_two', packages = dbt_utils._get_utils_namespaces())(upper_bound) }}
```
jtcohen6, line 2: This is causing the error in CI right now:
```suggestion
    {{ return(adapter.dispatch('get_powers_of_two', packages = dbt_utils._get_utils_namespaces())(upper_bound)) }}
```
We need to wrap this in `return()` to preserve the datatype (int) returned by `default__get_powers_of_two`. Otherwise, it's taken as a string.

You can think of the `dispatch` macro as one more gateway through which the macro's call + args need to pass one way, and through which the returned value passes back.

It's not a bad idea to wrap _all_ `adapter.dispatch` calls in `return()`, to be honest.

chaerinlee1: Hi Jeremy,

After wrapping all `adapter.dispatch` with `return()`, we are encountering the error below

image

Do you have an idea for how to fix this? Thank you!

macros/sql/unpivot.sql
```diff
@@ -13,6 +13,10 @@ Arguments:
 #}
 
 {% macro unpivot(relation=none, cast_to='varchar', exclude=none, remove=none, field_name='field_name', value_name='value', table=none) -%}
+    {{ return(adapter.dispatch('unpivot', packages = dbt_utils._get_utils_namespaces())(relation=none, cast_to='varchar', exclude=none, remove=none, field_name='field_name', value_name='value', table=none)) }}
```
jtcohen6, line 16: We need this to be either:
```suggestion
    {{ return(adapter.dispatch('unpivot', packages = dbt_utils._get_utils_namespaces())(relation, cast_to, exclude, remove, field_name, value_name, table)) }}
```
Or:
```suggestion
    {{ return(adapter.dispatch('unpivot', packages = dbt_utils._get_utils_namespaces())(relation=relation, cast_to=cast_to, exclude=exclude, remove=remove, field_name=field_name, value_name=value_name, table=table)) }}
```
This `kwarg=value` syntax operates quite differently in either place:
- In the macro signature, it sets default values if not otherwise passed by the user
- Whereas in the `dispatch` call, it actually _overrides_ both the default and the user-passed values!

macros/sql/surrogate_key.sql
```diff
@@ -1,4 +1,8 @@
 {%- macro surrogate_key(field_list) -%}
+    {{ return(adapter.dispatch('surrogate_key', packages = dbt_utils._get_utils_namespaces())(field_list, *args)) }}
+{% endmacro %}
```
swanderz, line 3: @jtcohen6 one more... thanks for bearing with us as we muddle through this!
I added an `*args` arg, but it still doesn't like it...
```
Compilation Error in model test_surrogate_key (models/sql/test_surrogate_key.sql)
  macro 'dbt_macro__surrogate_key' takes not more than 1 argument(s)
  
  > in macro surrogate_key (macros/sql/surrogate_key.sql)
  > called by model test_surrogate_key (models/sql/test_surrogate_key.sql)
```

jtcohen6: Ugh. Yeah I think this may need to be something like
```sql
{%- macro surrogate_key(field_list) -%}
    {% set placeholder = varargs %}
    {{ return(adapter.dispatch('surrogate_key', packages = dbt_utils._get_utils_namespaces())(field_list, *args)) }}
{% endmacro %}
```

integration_tests/dbt_project.yml
```diff
@@ -19,6 +19,9 @@ clean-targets:         # directories to be removed by `dbt clean`
     - "target"
     - "dbt_modules"
 
+vars:
+  dbt_utils_dispatch_list: ['dbt_utils']
+
```
swanderz, line 24: is this needed? I added it as part of my siege on Circle CI

macros/sql/pivot.sql
```diff
@@ -51,6 +51,20 @@ Arguments:
...
+               values,
+               alias=True,
+               agg='sum',
+               cmp='=',
+               prefix='',
+               suffix='',
+               then_value=1,
+               else_value=0,
+               quote_identifiers=True,
+               distinct=False) %}
```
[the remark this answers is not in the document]
jtcohen6: Providing `=` arguments in a macro's signature works the way you'd expect: These are the default values, if other values are not passed by the user / wrapper macro.

Providing `=` arguments when _calling_ another macro from within a macro passes those are _the_ arguments.

Given that the defaults are already set in `pivot` and passed down to `default__`, etc, it's technically redundant to set the same ones here, but it should work the way we expect.

integration_tests/dbt_project.yml
jtcohen6: I don't think so.

This is only needed if `dbt_utils_integration_tests ` needed to override a `dbt_utils` macro, or inject some new macros into the `dbt_utils` namespace. That said, the integration test project does define [two schema tests](https://github.com/fishtown-analytics/dbt-utils/blob/master/integration_tests/macros/tests.sql),  `test_assert_equal` and `test_not_empty_string`; do you see any reason why we'd need to override those?

macros/schema_tests/at_least_one.sql
```diff
@@ -1,4 +1,8 @@
 {% macro test_at_least_one(model) %}
+  {{ return(adapter.dispatch('test_at_least_one', packages = dbt_utils._get_utils_namespaces())(model, **kwargs)) }}
```
jtcohen6, line 2: Pleasantly surprised this doesn't need the same hack as `varargs` did. Tested locally just to be extra sure.

jtcohen6 (approved):
Thanks for all the hard work here @swanderz @chaerinlee1! Truly a public service on behalf of all adapter plugin users

integration_tests/dbt_project.yml
swanderz: the genesis for this PR was because we couldn't properly run the `dbt-utils`'s integration tests without being able to override the schema tests. my self-interested answer is: if t-sql can't run the integration tests's data tests, then we'll need a way to override them At a glance, I can't see how they could be an issue in TSQL, but I'm about to get my hands dirty to find out for sure

macros/schema_tests/at_least_one.sql
swanderz: yep the big trip ups were:
- passing non-kw args to a function without regular arguments, and
- assigning default values to args within a macro.
thanks for being here with me for the "Trough of Disillusionment" phase of the my personal Jinja hype cycle. it's grim!

jtcohen6: Jinja is pythonic, in good times & in bad

integration_tests/dbt_project.yml
jtcohen6: Sounds good. If needed, we can rewrite those tests to be in the `dbt_utils` namespace (i.e. dispatchable from the integration tests) but not _actually_ ship with `dbt_utils` the package. That's essentially what I did for [`prep_external`](https://github.com/fishtown-analytics/dbt-external-tables/blob/master/integration_tests/macros/common/prep_external.sql) over in the `dbt-external-tables` integration tests.

macros/schema_tests/expression_is_true.sql
```diff
@@ -1,4 +1,8 @@
 {% macro test_expression_is_true(model, condition='true') %}
```
swanderz, line 1: another quirk: in #310, I propose changing the default value for condition to be `1=1` instead of `'true'`. How might I set that default value to be different _only_ for tsql? isn't that something we can't do? In a normal scenario you could edit the `schema.yml`, but in this case tsql can't touch it as it's part of another repo.

jtcohen6: I wonder if you could do something like:
```
{% macro sqlserver__test_expression_is_true(model, condition) %}

  {# override generic default #}
  {% if condition = 'true' %}
      {% set condition = '1=1' %}
  {% endif %}
  
  {{ return(default__test_expression_is_true(model, condition) }}

{% endmacro %}
```

(although I'm not opposed to the proposed change in #310, making this `1=1` across the board)

swanderz: good idea. i'm going to break out more of these questions into a different PR, so that this PR's focus can just be about dispatching

clrcrl:
If my memory serves me correctly, we don't want to pass through the defaults twice. e.g. instead of:
```
{% macro test_expression_is_true(model, condition='true') %}
This conversation was marked as resolved by swanderz
  {{ return(adapter.dispatch('test_expression_is_true', packages = dbt_utils._get_utils_namespaces())(model, condition, **kwargs)) }}
{% endmacro %}

{% macro default__test_expression_is_true(model, condition='true') %}
...
```
We only want:
```
...
{% macro default__test_expression_is_true(model, condition) %}
...
```

I think this caused us issues in a Snowplow macro somwhere 🤔

jtcohen6:
@clrcrl It really could be an issue! I was trying to remember earlier today.

The problem in #279 is that we passed the default args into a macro _call_—not explicitly an issue with repeating them across two macro signatures.

clrcrl:
oh right, yes I see the difference now!

swanderz:
let this comment stand in for what should have been marking my draft PR as ready for review!

clrcrl:
@jtcohen6 — I'm going to let you do the honors here (unless you'd prefer me to!)

0.6.4 seems imminent!

jtcohen6 (approved):
Let's ship it!!

@swanderz Could you add a changelog entry (I think "under the hood"), and add yourself + @chaerinlee1 as contributors?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Speed up CI via threads, workflows ([#315](https://github.com/fishtown-analytics/dbt-utils/pull/315), [#316](https://github.com/fishtown-analytics/dbt-utils/pull/316))
 - Fix `equality` test when used with ephemeral models + explicit column set ([#321](https://github.com/fishtown-analytics/dbt-utils/pull/321))
 - Fix `get_query_results_as_dict` integration test with consistent ordering ([#322](https://github.com/fishtown-analytics/dbt-utils/pull/322))
+- All macros are now properly dispatched, making it possible for non-core adapters to implement a shim package for dbt-utils ([#312](https://github.com/fishtown-analytics/dbt-utils/pull/312)) Thanks [@chaerinlee1](https://github.com/chaerinlee1) and [@swanderz](https://github.com/swanderz)
```
swanderz, line 10: comme ça?

jtcohen6: merci !

PR #316, opened 2021-01-04
Switch to CircleCI workflow

### LINKED ISSUE #315: Increase threads for Integration Test suite
### Describe the feature
Currently integration tests run on 1 thread; this causes the integration test models and schema tests to run very slowly, particularly on Snowflake and BigQuery where a higher thread count would speed things up (and potentially save $ on Snowflake compute). 
See:

image

I propose increasing the `threads` to 4 or even higher, e.g. 8, for Snowflake and BigQuery integration tests.

### Who will this benefit?
Anyone submitting PRs and the FA team if this ends up saving Snowflake compute cost.

Yes

jtcohen6:
resolves #315

- Use 5 threads (Postgres, Redshift) or 10 (Snowflake, BigQuery)—an arbitrary number!
- Allows us to run tests on different warehouses (Redshift, Snowflake, BigQuery) simultaneously rather than in sequence
    - Removes the caching of dependencies. (a) I think it's pretty fast to install dbt. (b) Helpful if we need to test with a new (prerelease) version of dbt, without needing to manually invalidate cache. (c) I'm not entirely sure how to do these with workflows, and I know it can be tricky to share cache across concurrent jobs—[there are plenty of docs](https://circleci.com/docs/2.0/caching/) to read

This is a:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,23 +1,15 @@
 
-version: 2
+version: 2.1
 
 jobs:
-  build:
+  
+  integration-postgres:
     docker:
       - image: circleci/python:3.6.3-stretch
       - image: circleci/postgres:9.6.5-alpine-ram
 
     steps:
       - checkout
-
-      - run:
-          run: setup_creds
-          command: |
-            echo $BIGQUERY_SERVICE_ACCOUNT_JSON > ${HOME}/bigquery-service-key.json
-
-      - restore_cache:
-          key: deps2-{{ .Branch }}
-
       - run:
           name: "Run Tests - Postgres"
           environment:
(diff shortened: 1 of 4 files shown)


jtcohen6:
Screen Shot 2021-01-04 at 4 07 38 PM

clrcrl (approved):
This looks really great! Curious why we decided to have the three clouds depend on postgres, I'm guessing to catch errors early?

clrcrl:
While we're here, I wouldn't hate adding the `-x` flag to the commands!

jtcohen6:
> I'm guessing to catch errors early?

Yeah, that was my initial thought, but we don't need to do it that way—the tests are so fast, it's not like it saves us very much.

> While we're here, I wouldn't hate adding the `-x` flag to the commands!

Good call!

jtcohen6:
This is unexpected:
```
Compilation Error in macro materialization_insert_by_period_default (macros/materializations/insert_by_period_materialization.sql)
  'dbt.utils.AttrDict object' has no attribute 'status'
Encountered an error:
FailFast Error in model test_insert_by_period (models/materializations/test_insert_by_period.sql)
```
I wonder if this is the result of a weird interaction with `--fail-fast`?

clrcrl:
(I kind of liked the "postgres first" DAG)

clrcrl:
oh _of course_ `insert by period` fails. UGH! let me check it out

jtcohen6:
Hm I don't think this is a matter of the concurrent run (though that is a potential problem!). I ran `drop schema dbt_utils_integration_tests_redshift cascade`, reran from failed, and still see:
```
Compilation Error in macro materialization_insert_by_period_default (macros/materializations/insert_by_period_materialization.sql)
  'dbt.utils.AttrDict object' has no attribute 'status'
Encountered an error:
FailFast Error in model test_insert_by_period (models/materializations/test_insert_by_period.sql)
  ```

PR #317, opened 2021-01-04
Make queue first step of workflow

clrcrl:
Building  on #316 

This might help us out when we have the issue of two branches being updated at once, causing overlapping runs on a particular database.

The current failure message indicates that some work needs to be done on the CircleCI side. We might also need to add an API key to get their working, as per the [docs](https://circleci.com/developer/orbs/orb/eddiewebb/queue)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,9 @@
 
 version: 2.1
 
+orbs:
+  queue: eddiewebb/queue@1.5.0
+
 jobs:
   
   integration-postgres:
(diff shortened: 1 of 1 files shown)


jtcohen6:
```
Orb eddiewebb/queue@1.5.0 not loaded. To use this orb, an organization admin must opt-in to using third party orbs in Organization Security settings.
```
I think I have the permissions to turn this on, but I'd want to double check first.

Alternative thought:
- Make the schema `dbt_utils_integration_tests_redshift_{{ env_var('CIRCLE_PR_NUMBER') }}`
- Drop the schema as the final step of `run_test.sh`. (Or: Every few months I log into each database and drop a ton of schemata.)

clrcrl:
I had that thought too, but I think if the orb works nicely, we should just use that. It feels a little less error-prone for some reason

clrcrl:
Closing due to lack of interest :)

PR #309, opened 2020-12-19
Allow dbt v0.19.0

jtcohen6:
## Description & motivation
- Resolves #308
- Bumps upper bound to allow dbt v0.19.0, which has no breaking changes
- Since there are no breaking changes, I'd love to cut a patch release of dbt-utils that bumps its upper bounds _before_ the final release of v0.19.0. This would make it easier for folks to test v0.19.0-rc1, since it obviates the need to explicitly pass `--no-version-check`.

## TODO
- [x] Wait until v0.19.0-rc1 is released on pypi

## Checklist
- [x] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [x] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake
- ~I have updated the README.md (if applicable)~
- ~I have added tests & descriptions to my models (and macros if applicable)~
- [x] I have added an entry to CHANGELOG.md

PR #319, opened 2021-01-06
broken commit: update internals of insert by period

clrcrl:
@jtcohen6 — this is what is broken, but I can't quite get it to work! not sure how to access the internals of the AdapterResponse class. Figured you would have a better idea than me :) 

Also — this would be classed as a breaking change, so we might want to write some logic about handling this on different versions of dbt to avoid the break.

I was curious as to how this snuck in, and went back to #309 — turns out it [silently failed](https://app.circleci.com/pipelines/github/fishtown-analytics/dbt-utils/199/workflows/4426f6ef-333b-4336-b3e3-31cff022dce4/jobs/700/parallel-runs/0/steps/0-105) back then! I think because the "test" step passed, even though the "run" step failed. Now that we are using `-x`, the job exits, causing other tests to fail, and the resultant error message.

I think the problem here is that we run the steps via the `run_test.sh` file, when potentially we want them as separate (parameterized?) steps in the CircleCI workflow.

### DIFF
--- a/macros/materializations/insert_by_period_materialization.sql
+++ b/macros/materializations/insert_by_period_materialization.sql
@@ -142,7 +142,9 @@
           from {{tmp_relation.include(schema=False)}}
       );
     {%- endcall %}
-    {%- set rows_inserted = (load_result('main-' ~ i)['status'].split(" "))[2] | int -%}
+    {% set result = load_result('main-' ~ i) %}
+    {% set rows_inserted = result['response']['rows_affected'] %}
+    {% endif %}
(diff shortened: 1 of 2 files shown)


jtcohen6:
Nice work on this one! Check out #320. I think we can still manage this as a patch release.

> Now that we are using `-x`, the job exits, causing other tests to fail, and the resultant error message.

Which is a good thing!

> I think the problem here is that we run the steps via the `run_test.sh` file, when potentially we want them as separate (parameterized?) steps in the CircleCI workflow.

Ideally the bash script would stop on the first error. We could maybe get this for free by adding `-e` to the script invocation... though there are [known caveats](http://mywiki.wooledge.org/BashFAQ/105).

PR #321, opened 2021-01-07
Fix equality test with ephemeral model + explicit column set

jtcohen6:
See: https://github.com/fishtown-analytics/spark-utils/issues/7

This is a:

## Description & motivation

If `equality` is comparing an ephemeral model with the `compare_columns` keyword argument, we want to avoid running `adapter.get_columns_in_relation(model)` entirely.

Currently, it is the default/fallback arg to `kwargs.get()`. Instead, we should place it inside a conditional that only runs if `kwargs.get(compare_columns)` returns none. I was under the impression that these were effectively the same, but this change actually fixes an integration test (`test_equal_column_subset`) on `dbt-utils` + `spark-utils`.

Why did this only break with Spark? Because `dbt-spark` uses `describe table [table_name]` to get the columns from a given table, a command that fails if the table does not exist. Most "core" adapters query the information schema instead; if the table doesn't exist, the query returns no results, but it doesn't fail explicitly.

## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake
    - [x] Spark
- ~I have updated the README.md (if applicable)~
- ~I have added tests & descriptions to my models (and macros if applicable)~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# dbt-utils v0.6.4
+
+- Fix `insert_by_period` to support `dbt v0.19.0`, with backwards compatibility for earlier versions ([#319](https://github.com/fishtown-analytics/dbt-utils/pull/319), [#320](https://github.com/fishtown-analytics/dbt-utils/pull/320))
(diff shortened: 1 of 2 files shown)


clrcrl (approved):
Makes sense to me!

PR #332, opened 2021-02-08
WIP: Add cross db ceiling macro

emilieschario:
This is a:

## Description & motivation

This adds a default call and an alternate call for `ceil` or `ceiling` so that it can be used across more systems. Notably, this makes things Spark compatible. https://github.com/fishtown-analytics/spark-utils/issues/5

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake


emilieschario:
@clrcrl can you share what's failing? I can't see the pipeline details.

macros/cross_db_utils/ceil.sql
```diff
@@ -0,0 +1,11 @@
+{% macro ceil(field) -%}
+  {{ return(adapter.dispatch('ceil', packages = dbt_utils._get_utils_namespaces())(fields)) }}
+{%- endmacro %}
+
+{% macro default__ceil(field) -%}
+    ceil({{ field }})
+{%- endmacro %}
+
+{% macro alternative_ceil(field) %}
```
clrcrl, line 9: Which adapters need to use `ceiling` over `ceil`? We'll want to adjust this to something like:
```suggestion
{% macro redshift__ceil(field) %}
```

integration_tests/models/cross_db_utils/test_ceil.sql
```diff
@@ -0,0 +1,12 @@
+
+with data as (
+
+    select * from {{ ref('data_ceil') }}
+
+)
+
+select
+    {{ dbt_utils.ceil(input) }} as actual,
```
clrcrl, line 9: This needs to be passed as a string — otherwise Jinja is looking for a variable named `input`
```suggestion
    {{ dbt_utils.ceil('input') }} as actual,
```

clrcrl (changes requested):
😅 
sorry about the delay here!

macros/cross_db_utils/ceil.sql
emilieschario: it's spark specifically (so dbx, but I thought it took the default then the alternative then database specific ones? since there was only on alternative (redshift, bq, and snowflake are all ceil), I thought one catchall might make more sense? If you don't think so, I can move this PR to spark_utils.

clrcrl: I guess in both cases we'd need the default in dbt-utils, and the override in spark utils. Since none of our core adapters use `ceiling`, I think we'd want the spark-specific version over in spark-utils

macros/cross_db_utils/ceil.sql
```diff
@@ -0,0 +1,11 @@
+{% macro ceil(field) -%}
+  {{ return(adapter.dispatch('ceil', packages = dbt_utils._get_utils_namespaces())(fields)) }}
```
clrcrl, line 2: Ah, this argument as to match the arguments of the macro. This syntax is... not great 😬 

```suggestion
  {{ return(adapter.dispatch('ceil', packages = dbt_utils._get_utils_namespaces())(field)) }}
```

clrcrl (changes requested):

PR #341, opened 2021-03-03
Update README.md

haf:
Improving documentation for how to override macros.

### DIFF
--- a/README.md
+++ b/README.md
@@ -828,7 +828,7 @@ Set a variable in your `dbt_project.yml`:
 ```yml
 vars:
   dbt_utils_dispatch_list:
-    - my_analytics_proj          # likely the name of your root project (only the root folder)
+    - first_package_to_search    # likely the name of your root project (only the root folder)
     - second_package_to_search   # likely an "add-on" package, such as spark_utils
     # dbt_utils is always the last place searched
 ```
--- a/README.md
+++ b/README.md
@@ -833,7 +833,7 @@ vars:
     # dbt_utils is always the last place searched
 ```
 
-When overriding macros, you need to prefix their names with `default__` (note the two underscores).
+If overriding a dispatched macro with a custom implementation in your own project's `macros/` directory, you must name your custom macro with a prefix: either `default__` (note the two underscores), or the name of your adapter followed by two underscores. For example, if you're running on Postgres and wish to override the behavior of `dbt_utils.datediff` (such that `dbt_utils.date_spine` will use your version instead), you can do this by defining a macro called either `default__datediff` or `postgres__datediff`.
 
 When running on Spark, if dbt needs to dispatch `dbt_utils.datediff`, it will search for the following in order:
 ```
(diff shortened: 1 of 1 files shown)


clrcrl:
Reassigning to Jeremy as dispatch macros are his area of knowledge :)

README.md
```diff
@@ -824,15 +824,17 @@ variable in your project, when dbt searches for implementations of a dispatched
 `dbt_utils` macro, it will search through your listed packages _before_ using
 the implementations defined in `dbt_utils`.
 
-Set the variable:
+Set the property `var` in `dbt_project.yml`:
```
jtcohen6, line 827: You're right to specify that this should be in `dbt_project.yml`. I'd rather refer to this as a "variable" vs. a "property," since [the latter word](https://docs.getdbt.com/reference/declaring-properties) has a different contextual meaning.
```suggestion
Set a variable in your `dbt_project.yml`:
```

README.md
```diff
@@ -824,15 +824,17 @@ variable in your project, when dbt searches for implementations of a dispatched
...
 ```yml
 vars:
   dbt_utils_dispatch_list:
-    - first_package_to_search    # likely the name of your root project
+    - my_analytics_proj          # likely the name of your root project (only the root folder)
```
jtcohen6, line 831: Could we keep this as `first_package_to_search`, for the sake of the example further down? I think it's fair to clarify that this is the `name:` property from `dbt_project.yml` of your root project.

README.md
```diff
@@ -812,8 +812,8 @@ We welcome contributions to this repo! To contribute a new feature or a fix, ple
 ### Dispatch macros
 
 **Note:** This is primarily relevant to users and maintainers of community-supported
-database plugins. If you use Postgres, Redshift, Snowflake, or Bigquery, this likely
-does not apply to you.
+database plugins, or if you want to override a default function. If you use Postgres,
+Redshift, Snowflake, or Bigquery, this likely does not apply to you.
```
jtcohen6, line 816: Good call. Here's wording I might use to make the target audience really clear:
```suggestion
**Note:** This is primarily relevant to:

- Users and maintainers of community-supported [adapter plugins](https://docs.getdbt.com/docs/available-adapters)
- Users who wish to override a low-lying `dbt_utils` macro with a custom implementation, and have that implementation used by other `dbt_utils` macros

If you use Postgres, Redshift, Snowflake, or Bigquery, this likely does not apply to you.
```

jtcohen6 (commented):
Thanks for this @haf!

Am I right in seeing [this Slack thread](https://getdbt.slack.com/archives/CBSQTAPLG/p1614776083328100) as the context for the change here?

I left a few suggestions. It feels like there are two adjacent-yet-different use cases for leveraging dispatch:

1. Extending `dbt_utils` macros to new adapters, by reimplementing low-lying functionality
2. Overriding `dbt_utils` macros on existing adapters, by reimplementing low-lying functionality (and ensuring the custom version is used by other `dbt_utils` macros)

There's a lot of overlap between them in terms of the actual steps required. I'd like to keep the guidance and context as clear as possible; the reader is likely here because they want to do one or the other.

jtcohen6 (approved):
Thanks for this @haf!

PR #390, opened 2021-07-15
Update dbt_project.yml

Ray7X:
The version is incorrect in dbt_project.yml file, changing it to the correct version.

motivation:
The incorrect version in dbt_project.yml is causing bugs in codegen package since codegen 0.4.0 requires dbt_utils version [">=0.6.2", "<0.8.0"]

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -1,5 +1,5 @@
 name: 'dbt_utils'
-version: '0.1.0'
+version: '0.7.0'
 
 require-dbt-version: [">=0.20.0", "<0.21.0"]
 


jtcohen6:
Hey @Ray7X, the `version` in `dbt_project.yml` is cosmetic only. It doesn't serve any functional purpose. I agree it'd be good to keep it up to date, though!

> The incorrect version in dbt_project.yml is causing bugs in codegen package since codegen 0.4.0 requires dbt_utils version [">=0.6.2", "<0.8.0"]

Could you say more about the issue you're running into? What error(s) are you seeing?

alexiswo:
Jumping in here because we discovered this issue during a Group Training session. @Ray7X was the one who figured out the bug, so thank you Ray!

I have this staged in the Nasdaq dbt Learn account. I'm happy to screen share and show you the bug! (I don't want to commit the change and mess up all of our nice branches - you get it 😉 )

## Steps for repro

1. Paste snippet into `packages.yml` 
  ```packages:
    - package: dbt-labs/codegen
      version: 0.4.0
  ```
_Note: weirdly this bug only occurs when you only install codegen. If you include codegen AND dbt_utils in packages.yml, codegen works fine._
2. Run `dbt deps`
3. Paste snippet into a statement tab and click "Compile SQL" `{{ codegen.generate_source('ticket_tailor', 'raw') }}`
4. Get error in Results tab
  ```Server error: Database Error in rpc request (from remote system)
  001003 (42000): SQL compilation error:
  syntax error line 1 at position 0 unexpected 'version'.
  ```
5. Testing the proposed fix: Manually edit the dbt_modules/dbt_utils/packages.yml file to change the `version: 0.1.0` to `version: 0.7.0`
6. Go back to statement tab and run "Compile SQL" on the same codegen macro as above in step #3. This time it works!

jtcohen6 (approved):
@alexiswo Very weird!! I'd definitely be interested to see a screenshare / loom of this one, given how strange it is.

In any case, I'm totally on board with this change. LGTM

alexiswo:
I met with @jtcohen6 today to try to show him the bug but now I'm having trouble reproducing it! Perhaps we were clicking on "Preview SQL" instead of "Compile SQL" when using the codegen macro? Either way this change is still a good one and thank you @Ray7X!

Ray7X:
@alexiswo Glad to help! I was having trouble preproducing the bug as well, but I remember I had trouble not only when compiling SQL, but also just trying to install the codegen package with dbt deps (after encountered the bug).

PR #394, opened 2021-07-18
Feature: Add optional `lookback_interval` param to `insert_by_period` materialization

GClunies:
This is a:
- [ ] bug fix PR with no breaking changes — please ensure the base branch is `master`
- [x] new functionality — please ensure the base branch is the latest `dev/` branch
- [ ] a breaking change — please ensure the base branch is the latest `dev/` branch

## Description & motivation
The `insert_by_period` materialization does not work with sessionization models like [`segment_web_page_views__sessionized`](https://github.com/dbt-labs/segment/blob/master/models/sessionization/segment_web_page_views__sessionized.sql) from the [dbt Segment package](https://github.com/dbt-labs/segment) due to some [unusually complex](https://github.com/dbt-labs/segment/blob/f63ae8baa02236237faf83c3ba678265828e80c4/models/sessionization/segment_web_page_views__sessionized.sql#L21-L34) logic. Since these sessionization models are often _very large and complex_, I would like to be able to use the `insert_by_period` materialization with them.

This is my first PR to a dbt repo, let alone a dev branch, so please let me know if any additional work is needed for this PR.

Also, huge thanks to Scott Barber on the dbt-labs customer support engineering team to get me started on refactoring the macro! His input and guidance were a massive help and this was a great learning experience for me.

## Describe your changes, and why you're making them.
This PR adds an optional `lookback_interval` parameter to the `insert_by_period` materialization, allowing it to be used with sessionization models like [`segment_web_page_views__sessionized`](https://github.com/dbt-labs/segment/blob/master/models/sessionization/segment_web_page_views__sessionized.sql) from the [dbt Segment package](https://github.com/dbt-labs/segment).

This is achieved by allowing the end user to include `__PERIOD_FILTER_WITH_LOOKBACK__` in their model SQL, which the materialization replaces at runtime of each period using this modified macro logic in the materialization code.
```sql
{% macro default__get_period_sql(target_cols_csv, sql, timestamp_field, period, start_timestamp, stop_timestamp, offset, lookback_interval) -%}

  {%- set period_filter -%}
    ("{{timestamp_field}}" >  '{{start_timestamp}}'::timestamp + interval '{{offset}} {{period}}' and
     "{{timestamp_field}}" <= '{{start_timestamp}}'::timestamp + interval '{{offset}} {{period}}' + interval '1 {{period}}' and
     "{{timestamp_field}}" <  '{{stop_timestamp}}'::timestamp)
  {%- endset -%}

  {%- set period_filter_with_lookback -%}
[...]

PR #189, opened 2020-02-11
WIP: changed the macro to work with snowflake

nehiljain:

PR #410, opened 2021-09-03
Modernize legacy insert by period materialization

HorvathDanielMarton:
This is a:
- [X] bug fix PR with no breaking changes — please ensure the base branch is `master`

## Description & motivation

The goal of this change is to get the `insert_by_period` materialization compatible with Snowflake while using the currently latest dbt (0.21.0b2) and keeping the same functionality. 

The `insert_by_period` materialization was introduced in [this discourse post](https://discourse.getdbt.com/t/a-materialization-for-problematic-initial-builds/60) and the intention behind it is to help with “problematic” initial builds of data models. Here, problematic means that some tables are just so large and/or complex that a simple `--full-refresh` is not adequate because the build will be incredibly slow, inefficient, and can even get stuck.

Unfortunately, the currently available version does not support Snowflake and it’s [advised to refactor](https://github.com/dbt-labs/dbt-utils/pull/189) it by using the refreshed code for the [incremental](https://github.com/dbt-labs/dbt/blob/0.21.latest/core/dbt/include/global_project/macros/materializations/incremental/incremental.sql) materialization as a good starting point. As a result, this change resolves dbt-labs/dbt-labs-experimental-features#32.

The materialization itself was completely built from the current incremental materialization and the modifications done to it was inspired by the [original](https://github.com/dbt-labs/dbt-utils/blob/master/macros/materializations/insert_by_period_materialization.sql) `insert_by_period`.

The macros that were originally in the `insert_by_period.sql` was moved to a `helpers.sql` file where the Snowflake version of `get_period_boundaries()`, and `get_period_sql()` are introduced. Also, a new `check_for_period_filter()` macro has been added.

The materialization seems fully functional and so far works as expected, nevertheless, I'm still testing it against production models. 

### Ideas 

1. Currently, in the config we can only set a period of `day`, `week`, `month`, etc... as a unit. However, we might want to build our models fortnightly (ie. 2-week chunks). We could introduce an optional configuration that lets us use a configurable number of `period`s as a chunk. The fortnightly example could look like this:
```sql
{{
    config(
        materialized = "insert_by_period",
        timestamp_field = "created_at",
        period = 'week',
        interval_size = 2,
        start_date = "2021-01-01",
        stop_date = "2021-08-31"
    )
}}
```
2. In some of our more complex models with multiple sources, we might need to offset the `__PERIOD_FILTER__` by some time (for example a month lookback time, but it should be configurable). Currently, this materialization doesn't support this, but that is surely a shortcoming for some of our models where `insert_by_period` would come in handy because of their massive size & complexity. Also, it seems like our team is not the only ones who bumped into this: dbt-labs/dbt-utils#394

### A weird behavior

The materialization can produce a weird behavior if it’s executed on an already existing model because the “last incremental run” may have a _very_ short overlap with the specified date range. Let’s take the following config as an example:

```sql
{{
    config(
        materialized = "insert_by_period",
        timestamp_field = "created_at",
        period = 'day',
        start_date = "2021-08-28",
        stop_date = "2021-08-31"
    )
}}
```

This is going to build a model with events `2021-08-28 < created_at < 2021-08-31`, so it will have 3 days, thus in 3 steps it will be built, given it’s a fresh start and the run is uninterrupted. After building the 1st day, if the run is terminated, the next run will have 3 steps, again. It's not unlikely, that the last step will only insert a handful of events (many magnitudes smaller, than the previous ones). Like on this CLI output:

Screen Shot 2021-08-31 at 17 05 03

This may raise some eyebrows, but it makes perfect sense if we look at the query’s where clause - which is basically this:

```sql
where created_at > '2021-08-30 23:59:59.852'::timestamp 
   and  created_at <= '2021-08-31 23:59:59.852'::timestamp -- this isn't relevant at all
   and  created_at <  '2021-08-30 23:59:59.999'::timestamp)
```
So the last iteration is only looking for events that happened between `2021-08-30 23:59:59.852` and `2021-08-30 23:59:59.999`, which is far from even being a second. It would be nice to deal with this, so it won't cause suspicion that something is buggy.

This is most conspicuous for large models with a lot of events. 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [X] Snowflake
- [X] I have "dispatched" any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/star.sql))

### DIFF
--- a/macros/materializations/helpers.sql
+++ b/macros/materializations/helpers.sql
@@ -1,4 +1,8 @@
 {% macro check_for_period_filter(model_unique_id, sql) %}
+    {{ return(adapter.dispatch('check_for_period_filter', 'dbt_utils')(model_unique_id, sql)) }}
+{% endmacro %}
+
+{% macro default__check_for_period_filter(model_unique_id, sql) %}
     {%- if sql.find('__PERIOD_FILTER__') == -1 -%}
         {%- set error_message -%}
         Model '{{ model_unique_id }}' does not include the required string '__PERIOD_FILTER__' in its sql
--- a/macros/materializations/helpers.sql
+++ b/macros/materializations/helpers.sql
@@ -18,10 +18,15 @@
 {% macro default__get_period_boundaries(target_schema, target_table, timestamp_field, start_date, stop_date, period) -%}
 
   {% call statement('period_boundaries', fetch_result=True) -%}
+    {{ dbt_utils.log_info(model) }}
+    {%- set model_name = model['name'] -%}
+    
     {%- set target_schema_msg = " [DEBUG] (get_period_boundaries) target schema is: " ~ target_schema -%}
     {%- set target_table_msg = " [DEBUG] (get_period_boundaries) target table is: " ~ target_table -%}
+    {%- set model_name_msg = " [DEBUG] (get_period_boundaries) model[name] is: " ~ model_name -%}
     {{ dbt_utils.log_info(target_schema_msg) }}
     {{ dbt_utils.log_info(target_table_msg) }}
+    {{ dbt_utils.log_info(model_name_msg) }}
     
     with data as (
       select
@@ -32,7 +37,7 @@
                                 "nullif('" ~ stop_date ~ "','')::timestamp")}},
             {{dbt_utils.current_timestamp()}}
           ) as stop_timestamp
-      from "{{target_schema}}"."{{target_table}}"
+      from "{{target_schema}}"."{{model_name}}"
     )
 
     select
--- a/macros/materializations/helpers.sql
+++ b/macros/materializations/helpers.sql
@@ -28,6 +28,35 @@
 
 {%- endmacro %}
 
+{% macro snowflake__get_period_boundaries(target_schema, target_table, timestamp_field, start_date, stop_date, period) -%}
+
+  {% call statement('period_boundaries', fetch_result=True) -%}
+    {%- set model_name = model['name'] -%}
+    
+    with data as (
+      select
+          coalesce(max({{timestamp_field}}), '{{start_date}}')::timestamp as start_timestamp,
+          coalesce(
+            {{dbt_utils.dateadd('millisecond',
+                                -1,
+                                "nullif('" ~ stop_date ~ "','')::timestamp")}},
+            {{dbt_utils.current_timestamp()}}
+          ) as stop_timestamp
+      from {{target.schema}}.{{model_name}}
+    )
+
+    select
+      start_timestamp,
+      stop_timestamp,
+      {{dbt_utils.datediff('start_timestamp',
+                           'stop_timestamp',
+                           period)}}  + 1 as num_periods
+    from data
+  {%- endcall %}
+
+{%- endmacro %}
+
+
 {% macro get_period_sql(target_cols_csv, sql, timestamp_field, period, start_timestamp, stop_timestamp, offset) -%}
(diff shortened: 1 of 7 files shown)


joellabes:
Worth noting: anyone else who comes along in the meantime is of course welcome to make a copy of the code in this PR and apply it to their own project!

macros/materializations/helpers.sql
```diff
@@ -0,0 +1,113 @@
...
+
+{% macro get_period_sql(target_cols_csv, sql, timestamp_field, period, start_timestamp, stop_timestamp, offset) -%}
+    {{ return(adapter.dispatch('get_period_sql', 'dbt_utils')(target_cols_csv, sql, timestamp_field, period, start_timestamp, stop_timestamp, offset)) }}
+{% endmacro %}
+
+{% macro default__get_period_sql(target_cols_csv, sql, timestamp_field, period, start_timestamp, stop_timestamp, offset) -%}
+
+  {%- set period_filter -%}
+    ("{{timestamp_field}}" >  '{{start_timestamp}}'::timestamp + interval '{{offset}} {{period}}' and
+     "{{timestamp_field}}" <= '{{start_timestamp}}'::timestamp + interval '{{offset}} {{period}}' + interval '1 {{period}}' and
```
nhrebinka, line 83: This line will not work properly if `period = "month"` 
example: 
```
select  '2018-12-31 00:00:00'::timestamp + interval '4 month' + interval '1 month', 
        '2018-12-31 00:00:00'::timestamp + interval '5 month'
```
2019-05-30 00:00:00 & 2019-05-31 00:00:00 => one day will not be processed.
Redshift can deal with the brackets, like `'{{start_timestamp}}'::timestamp + (interval '{{offset}} {{period}}' + interval '1 {{period}}')`, Snowflake will not like it. 
Suggestion (that I cannot test right away): calculate `offset_plus_one` as `offset + 1` and use it in setting the filter, like
```
  {%- set period_filter -%}
    ("{{timestamp_field}}" >  '{{start_timestamp}}'::timestamp + interval '{{offset}} {{period}}' and
     "{{timestamp_field}}" <= '{{start_timestamp}}'::timestamp + interval '{{offset_plus_one}} {{period}}' and
     "{{timestamp_field}}" <  '{{stop_timestamp}}'::timestamp)
  {%- endset -%}
```

nhrebinka: I tested this block and it works ok. Feel free to use it.
```
{% macro get_period_sql(target_cols_csv, sql, timestamp_field, period, start_timestamp, stop_timestamp, offset) -%}

  {%- set offset_plus_one = offset + 1 -%}

  {%- set period_filter -%}
    ("{{timestamp_field}}" >  '{{start_timestamp}}'::timestamp + interval '{{offset}} {{period}}' and
     "{{timestamp_field}}" <= '{{start_timestamp}}'::timestamp + interval '{{offset_plus_one}} {{period}}' and
     "{{timestamp_field}}" <  '{{stop_timestamp}}'::timestamp)
  {%- endset -%}

  {%- set filtered_sql = sql | replace("__PERIOD_FILTER__", period_filter) -%}

  select
    {{target_cols_csv}}
  from (
    {{filtered_sql}}
  )

{%- endmacro %}
```

joellabes:
Folks on this thread will probably be interested in this Discussion: https://github.com/dbt-labs/dbt-utils/discussions/487 

I'd appreciate your input!

joellabes:
Unsurprisingly, I can't transfer a PR from one repo to another, but if you want to reopen the PR over there I'll be happy to merge it 👍

Issue #417, opened 2021-09-16
In 0.7.2, dbt_utils.star does not select all of the columns that it should select


EPNickBosma:
### Describe the bug

We have found that with dbt_utils 0.7.2, `star` does not select all the columns that it should. Some of our models in production fail to build because `star` is not pulling out an ID column that it should be pulling out.

This does not happen with `star` on dbt_utils 0.7.1, so we've put our main branch back on that version for the time being.

### Steps to reproduce

Unsure on _exact_ repro steps, except to say that once `dbt-labs/dbt_utils 0.7.2` is installed, `dbt_utils.star` is liable to leave columns out of its selection even though they're not in the `except` list. In our case it's an "id" column that's not being selected, so the fact that its name is `id` may or may not be technically relevant.

In our case, we're using `star` to pull things out of a source as opposed to a model - again, this may or may not be technically relevant.

### Expected results

When using `dbt_utils.star`, we expect to see all of the columns in the source that we're referring to pulled out, except any columns we've used the `except` list to exclude.

### Actual results

There is a column excluded that we did not specify as a column that should be excluded - in a sense, `dbt_utils.star` simply misses the column.

### Screenshots and log output

We have model code as follows -- note that we do use an `except` list but we do not specify that 'id' should be excluded:

```
with countries as (
    select {{dbt_utils.star(source('organisations', 'countries'), except=['currencyid','phoneprefix','outboundemail', 'inboundemail'])}}
    from {{source('organisations', 'countries')}}
)

select 
    id as country_id,
    ....
from countries
```

The compiled SQL as per dbt_utils 0.7.2 begins as follows - note that there is no `id` column:

```
with countries as (
    select "name",
  "abbreviation",
  "countrycode2char",
  "smallflagurl",
  "defaultdatasetgroupid",
  "defaulttimezoneid"
    from "reporting"."schools"."countries"
)

select 
    id as country_id,
....
```

The compiled SQL as per dbt_utils 0.7.1 begins as follows - note that there **is** an `id` column as there should be:

```
with countries as (
    select "id",
  "name",
  "abbreviation",
  "countrycode2char",
  "smallflagurl",
  "defaultdatasetgroupid",
  "defaulttimezoneid"
    from "reporting"."schools"."countries"
)

select 
    id as country_id,
    ....
```

A screenshot of the columns in the relevant source table:

[...]

PR #418, opened 2021-09-16
fix: in star make excludes a list

jasnonaz:
This is a:
- [ X] bug fix PR with no breaking changes — please ensure the base branch is `master`

## Description & motivation
The feature added in 0.7.2 which allowed for star excludes to be case insensitive appears to have brought in an issue causing columns to get incorrectly dropped. https://github.com/dbt-labs/dbt-utils/issues/417

It looks like this is due to the |lower command turning a list into a string - this fixes that by looping over the columns to exclude and lowercasing them individually. 

## Checklist
    - [ X] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
- [ ] I have "dispatched" any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/star.sql))

### DIFF
--- a/macros/sql/star.sql
+++ b/macros/sql/star.sql
@@ -12,14 +12,11 @@
     {% endif %}
 
     {%- set include_cols = [] %}
-    {%- set exclude_cols_lower_cased = [] %}
-    {%- for col in except -%}
-      {% do exclude_cols_lower_cased.append(col|lower) %}
-    {%- endfor %}
     {%- set cols = adapter.get_columns_in_relation(from) -%}
+    {%- set except = except | map("lower") | list %}
     {%- for col in cols -%}
 
-        {%- if col.column|lower not in exclude_cols_lower_cased -%}
+        {%- if col.column|lower not in except -%}
             {% do include_cols.append(col.column) %}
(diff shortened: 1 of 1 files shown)


jtcohen6 (approved):

clausherther:
FYI, I think another way to do this with looping and explicitly creating a new list would be something like 
```python
set cols = cols | map("lower") | list
set except = except | map("lower") | list
```

wuservices:
> FYI, I think another way to do this with looping and explicitly creating a new list would be something like
> 
> ```python
> set cols = cols | map("lower") | list
> set except = except | map("lower") | list
> ```

@clausherther Is `| list` necessary?

Could you just change from `|lower` to `|map('lower') i.e. from
```
{%- if col.column|lower not in except|lower -%}
```
to
```
{%- if col.column|lower not in except|map('lower') -%}
```

reidwil-jb:
I appreciate the quick solutions on my first commit and subsequently first bug. 🐛

PR #426, opened 2021-10-01
Add handling for `android-app://` in `get_url_host()` macro

foundinblank:
This is a:

## Description & motivation
In my Segment page view data, we see `referrer` values such as `android-app://m.facebook.com/`. These seem to be specific for Android apps and take the place of `http://`. The current `get_url_host()` macro will return an `android-app://` value instead of the hostname for those referrer URLs. Be great if the `get_url_host()` macro handled those too!

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
- [ ] I have "dispatched" any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/star.sql))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,12 @@ This is a compatibility release in preparation for `dbt-core` v1.0.0 (🎉). Whe
 - Bump `require-dbt-version` to have an upper bound of `'<=1.0.0'`.
 - Url link fixes within the README for `not_constant`, `dateadd`, `datediff` and updated the header `Logger` to `Jinja Helpers`. ([#431](https://github.com/dbt-labs/dbt-utils/pull/431))
 - Fully qualified a `cte_name.*` in the `equality` test to avoid an Exasol error ([#420](https://github.com/dbt-labs/dbt-utils/pull/420))
+- `get_url_host()` macro now correctly handles URLs beginning with `android-app://` ([#426](https://github.com/dbt-labs/dbt-utils/pull/426))
 
 ## Contributors:
 - [joemarkiewicz](https://github.com/fivetran-joemarkiewicz)
 - [TimoKruth](https://github.com/TimoKruth)
+- [foundinblank](https://github.com/foundinblank)
 
 # dbt-utils v0.7.3
 
--- a/integration_tests/data/web/data_url_host.csv
+++ b/integration_tests/data/web/data_url_host.csv
@@ -2,6 +2,6 @@ original_url,parsed_url
 www.google.co.uk?utm_source=google&utm_medium=cpc&utm_campaign=spring-summer,www.google.co.uk
 http://witanddelight.com/2018/01/tips-tricks-how-run-half-marathon-first-time/,witanddelight.com
 https://www.nytimes.com/2018/01/01/blog,www.nytimes.com
-android-app://m.facebook.com/,m.facebook.com,
-docs.nytimes.com/2021/01/01/index.js?utm_source=google,docs.nytimes.com,
+android-app://m.facebook.com/,m.facebook.com
+docs.nytimes.com/2021/01/01/index.js?utm_source=google,docs.nytimes.com
 https://m.facebook.com/,m.facebook.com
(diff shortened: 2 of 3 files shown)


foundinblank:
1. I'd add an entry to CHANGELOG but it seems it's on 0.7.1 while the latest release is 0.7.3? Not sure what to do here. 
2. Is this a bug fix or new functionality? I tried first as new functionality but Github suggested I couldn't merge to the latest `dev` branch, so I'm trying this as a bug fix and merging to `master`.

joellabes:
Hey @foundinblank, thanks for the PR! It looks reasonable to me. 

Could I also ask you to flesh out the [integration test](https://github.com/dbt-labs/dbt-utils/blob/master/integration_tests/models/web/test_url_host.sql) ([source seed here](https://github.com/dbt-labs/dbt-utils/blob/master/integration_tests/data/web/data_url_host.csv)) to reflect your newly-supported URI scheme? This will mean that we get tests across all of the adapters.

The only thing that strikes me is that we're now at 3 replacements deep (`http://`, `https://`, `android-app://`). If it went any further, I'd be thinking that a `regex_replace` for something along the lines of `^[-A-Za-z]+:\/\/` would be more appropriate ([in English](https://regex101.com/r/ZGCq19/1): from the start of the string, 1+ alpha chars or hyphens, followed by `://`). What do you think? Are there likely to be other URI schemes that we need to handle?

Regarding your target branch question, it's on the borderline but I'd probably call it new functionality. Let me look into the branch issue you mentioned!

With dbt 1.0 coming later this year, we're thinking about exactly how we handle the migration, but I'm expecting another minor release of dbt-utils with new functionality like this PR before then!

foundinblank:
Hey @joellabes, that's a great idea to use `regexp_replace`, so I've swapped that in and updated the integration test seed file. I keep running into this error, though: 

```
Compilation Error in model test_url_host (models/web/test_url_host.sql)
  'regexp_replace' is undefined
```

I'm pretty confused by it and have checked that the `regexp_replace()` function exists on all 4 databases. Googling it isn't much help either.

joellabes:
Ah! I think your problem is that you'll need to wrap the whole command in quotes, because the entire command is being passed into dbt_utils.split_part(). Right now, the dbt compiler is looking for a Jinja macro called regexp_replace.

`"regexp_replace(field, '^[-A-Za-z]+:\/\/', "''""`

Unfortunately, there's already a lot of quotes in the mix so I doubt that will work as written. Does Jinja let you escape with backslashes etc? I don't know off the top of my head. 

If it doesn't, then maybe we want to go down the `regexp_substr` route instead of using split_part. 

Or, we could set the replace command outside of the parsed block. In handwavy pseudocode, something like 
```
{% set replaced = regexp_replace() %}

{% set parsed = split_part(split_part(replaced)) %}

{{ dbt_utils.safe_cast() }} --as it currently is
```
would probably work. 

Sorry that this is blowing up a bit!

foundinblank:
@joellabes I couldn't make it work so I reverted back to the original approach with 3 nested replacements, which does work. I don't think we'll have more URI schemas other than `http` `https` and `android-app` in the future, but if we do maybe whoever updates this macro next will be better at Jinja parsing than I am!

joellabes:
I am 100% on board with this as an approach. Godspeed the next person. 

I feel guilty that I didn't have a stable dbt installation until last week (😬 😳 😱 ) so couldn't properly dive into the Jinja-weeds with you. 

I'll do a quick check of this in the next couple of days and get it merged!

foundinblank:
Haha - no problem! I didn't have a great dbt installation on my machine either, was relying mostly on CI/CD to do the lifting. We shall be better at this next time! 😄 Thanks for your help in making this happen!

joellabes:
@foundinblank Could I get you to add yourself to the changelog please?

foundinblank:
@joellabes done!

joellabes (approved):
Thank you for the contribution! 🚀

PR #436, opened 2021-11-08
add prefix and suffix arguments to star macro

fivetran-jamie:
This is a:

## Description & motivation

Just added optional `prefix` and `suffix` arguments to the star macro, so that the output fields will all be renamed as prefix ~ name ~ suffix (very similar to the `pivot` macro). 

This is helpful to me in that I can more easily know where columns came from. Morewover, I have a particular case where I am using `star` twice in the same CTE, and the columns included in each source relation that I'm joining/selecting from can have custom names. If there's any overlap between the two sources, we'll see an `ambiguous column` error without any prefix or suffix to distinguish them

## Checklist
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
- [ ] I have "dispatched" any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/star.sql)) **is this just adding the arguments to the dispatch line?**

### DIFF
--- a/README.md
+++ b/README.md
@@ -725,7 +725,7 @@ group by 1,2,3
 ```
 
 #### star ([source](macros/sql/star.sql))
-This macro generates a list of all fields that exist in the `from` relation, excluding any fields listed in the `except` argument. The construction is identical to `select * from {{ref('my_model')}}`, replacing star (`*`) with the star macro. This macro also has an optional `relation_alias` argument that will prefix all generated fields with an alias.
+This macro generates a list of all fields that exist in the `from` relation, excluding any fields listed in the `except` argument. The construction is identical to `select * from {{ref('my_model')}}`, replacing star (`*`) with the star macro. This macro also has an optional `relation_alias` argument that will prefix all generated fields with an alias (`relation_alias`.`field_name`). The macro also has optional `prefix` and `suffix` arguments, which will be appropriately concatenated to each field name in the output (`prefix` ~ `field_name` ~ `suffix`).
 
 **Usage:**
 ```sql
--- a/integration_tests/data/sql/data_star_prefix_suffix_expected.csv
+++ b/integration_tests/data/sql/data_star_prefix_suffix_expected.csv
@@ -1,0 +1,4 @@
+prefix_field_1_suffix,prefix_field_2_suffix,prefix_field_3_suffix
+a,b,c
+d,e,f
+g,h,i
--- a/integration_tests/models/sql/schema.yml
+++ b/integration_tests/models/sql/schema.yml
@@ -106,6 +106,11 @@ models:
       - dbt_utils.equality:
           compare_model: ref('data_star_expected')
 
+  - name: test_star_prefix_suffix
+    tests:
+      - dbt_utils.equality:
+          compare_model: ref('data_star_prefix_suffix_expected')
+
   - name: test_surrogate_key
     tests:
       - assert_equal:
--- a/integration_tests/models/sql/test_star_prefix_suffix.sql
+++ b/integration_tests/models/sql/test_star_prefix_suffix.sql
@@ -1,5 +1,5 @@
-{% set prefix_with = 'prefix_' %}
-{% set suffix_with = '_suffix' %}
+{% set prefix_with = 'prefix_' if target.type != 'snowflake' else 'PREFIX_' %}
+{% set suffix_with = '_suffix' if target.type != 'snowflake' else '_SUFFIX' %}
 
 with data as (
 
--- a/integration_tests/models/sql/test_star_prefix_suffix.sql
+++ b/integration_tests/models/sql/test_star_prefix_suffix.sql
@@ -1,0 +1,13 @@
+{% set prefix_with = 'prefix_' %}
+{% set suffix_with = '_suffix' %}
+
+with data as (
+
+    select
+        {{ dbt_utils.star(from=ref('data_star'), prefix=prefix_with, suffix=suffix_with) }}
+
+    from {{ ref('data_star') }}
+
+)
+
+select * from data
--- a/macros/sql/star.sql
+++ b/macros/sql/star.sql
@@ -1,8 +1,8 @@
-{% macro star(from, relation_alias=False, except=[]) -%}
-    {{ return(adapter.dispatch('star', 'dbt_utils')(from, relation_alias, except)) }}
+{% macro star(from, relation_alias=False, except=[], prefix='', suffix='') -%}
+    {{ return(adapter.dispatch('star', 'dbt_utils')(from, relation_alias, except, prefix, suffix)) }}
 {% endmacro %}
 
-{% macro default__star(from, relation_alias=False, except=[]) -%}
+{% macro default__star(from, relation_alias=False, except=[], prefix='', suffix='') -%}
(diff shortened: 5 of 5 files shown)


joellabes:
Hey @fivetran-jamie, this makes sense to me! Happy to welcome it into the fold. 

Could you add another test in the same vein as the existing one ([model.yml](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/schema.yml#L104-L107), [model](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/test_star.sql), [seed](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/data/sql/data_star.csv)) to verify that they are behaving the way you expect? 

dbt 1.0.0-rc1 is imminent, and so we'll be cutting utils 0.7.4 and 0.8.0 in the next week or so. Would love to sneak this in as well!

fivetran-jamie:
hey @joellabes happy to do so! i'm a little confused by the test files though (this is my first dbt-utils PR 😅 ) -- are there guidelines somewhere on what to add?

joellabes:
@fivetran-jamie There aren't (yet) sorry! In short, there's another dbt project inside of this repo called `integration_tests`. It has a [dependency on this folder](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/packages.yml) and Circle CI runs the integration tests project on each of the core 4 DWHs to check everything's OK. 

In this case, you'd want to do something along these lines:
- Make a model called `test_star_prefix_suffix.sql` which also selects from `{{ ref('data_star') }}` - basically the same as [test_star](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/test_star.sql)
- Instead of doing an `exclude` as the existing one does, it would pass prefix/suffix args in. 
- Then add a seed file called `data_star_prefix_suffix_expected` along the same lines as [data_star_expected](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/data/sql/data_star_expected.csv)
- And add another test in the [schema.yml file](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/schema.yml#L104-L107):
```yml
  - name: test_star_prefix_suffix
    tests:
      - dbt_utils.equality:
          compare_model: ref('data_star_prefix_suffix_expected')
```

Then when you push it up, the CI will check everything for you. if it fails, you can't actually see why 😢  just @ me and I'll check it out for you. 

Hope that gets you started!

joellabes:
@fivetran-jamie three outta four ain't bad! 

The CI error is: 
```
Database Error in test dbt_utils_equality_test_star_prefix_suffix_ref_data_star_prefix_suffix_expected_ (models/sql/schema.yml)
  000904 (42000): SQL compilation error: error line 47 at position 11
  invalid identifier '"prefix_FIELD_1_suffix"'
  compiled SQL at target/run/dbt_utils_integration_tests/models/sql/schema.yml/dbt_utils_equality_test_star_p_451b32d440e658c73dd0373a328e9b28.sql
Encountered an error:
FailFast Error in test dbt_utils_equality_test_star_prefix_suffix_ref_data_star_prefix_suffix_expected_ (models/sql/schema.yml)
```

Looks like a casing issue to me? (*shakes fist at Snowflake*)

fivetran-jamie:
darn you snowflake...trying out the target-conditional casing used in https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/test_pivot.sql

fivetran-jamie:
huzzah!

joellabes:
> trying out the target-conditional casing used in https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/test_pivot.sql

I hate it but I love it 💀  it feels very backwards to change the test to pass the reality, but the fact that there's prior art puts my mind a bit more at ease. 

I'll give it a test locally before I properly merge it in, but this is looking good! thank you 🌟

fivetran-jamie:
yah it didn't feel good to add that lol -- happy to investigate other solutions if needed, just came across this one first 🤷‍♂️ 

also thanks for laying out everything so clearly!

fivetran-jamie:
hm would it be best to automatically capitalize the prefix + suffix on snowflake warehouses? ie 

```sql
{% macro star(from, relation_alias=False, except=[], prefix='', suffix='') -%}
    {{ return(adapter.dispatch('star', 'dbt_utils')(from, relation_alias, except, prefix, suffix)) }}
{% endmacro %}

{% macro default__star(from, relation_alias=False, except=[], prefix='', suffix='') -%}
    {%- do dbt_utils._is_relation(from, 'star') -%}
    {%- do dbt_utils._is_ephemeral(from, 'star') -%}

    {%- if not execute -%}
        {{ return('') }}
    {% endif %}

    {%- set include_cols = [] %}
    {%- set cols = adapter.get_columns_in_relation(from) -%}
    {%- for col in cols -%}

            {% do include_cols.append(col.column) %}

        {%- endif %}
    {%- endfor %}

    {% set prefix_cased = prefix | upper if target.type == 'snowflake' else prefix %} -- <--- here
    {% set suffix_cased = suffix | upper if target.type == 'snowflake' else suffix %} -- <--- here

    {%- for col in include_cols %}

        {%- if relation_alias %}{{ relation_alias }}.{% else %}{%- endif -%}{{ adapter.quote(col)|trim }} as {{ adapter.quote(prefix_cased ~ col ~ suffix_cased)|trim }}
        {%- if not loop.last %},{{ '\n  ' }}{% endif %}

    {%- endfor -%}
{%- endmacro %}

```

joellabes:
> hm would it be best to automatically capitalize the prefix + suffix on snowflake warehouses
@fivetran-jamie I don't think so - in some circumstances (with explicit quoting for example) Snowflake can support lowercase. The hack in the tests is because the test suite doesn't behave that way, but the seed is lowercase. It doesn't feel good to change the behaviour in the macro itself 

One day I'd like to flesh out the test suite to cover more variations of quoting behaviour, but that would require another warehouse connection etc so I'm not in a rush to do that 😬

joellabes (approved):
🎉  Thanks for the contribution!

DataGuru2023:
Hi,
I used the macro as below and it generates a SQL that is not compatible with the database. The alias and the trim() are reversed behaviour -> BOOKINGCMB2 as trim(BOOKINGCMB2)
desired behaviour -> trim(BOOKINGCMB2) as BOOKINGCMB2 
The intent was to use trim function for all source columns to remove any spaces
------------------
//Macro usage
select
    {{ dbt_utils.star(from=ref('saviom_outboxbooking'),prefix='trim(', suffix=')',quote_identifiers=False) }}
from {{ ref('saviom_outboxbooking') }}
where DBT_VALID_TO is null
------------------
//output
SELECT
 BOOKINGCMB2 as trim(BOOKINGCMB2),
  BOOKINGCMB3 as trim(BOOKINGCMB3),
  BOOKINGCMB4 as trim(BOOKINGCMB4),
  BOOKINGCMB5 as trim(BOOKINGCMB5),
  USERCMB1 as trim(USERCMB1),
  PROJECTACTIVECLOSEIND as trim(PROJECTACTIVECLOSEIND)
from DEV_IMG.SAVIOM.OUTBOXBOOKING
where DBT_VALID_TO is null

PR #442, opened 2021-11-12
feat(utils): add object type in datatypes.sql

vicmattos:
This is a:

## Description & motivation
Offer object datatype in the package

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [x] Snowflake
    - [ ] using `dbt_utils.type_*` macros instead of explicit datatypes (e.g. `dbt_utils.type_timestamp()` instead of `TIMESTAMP`)

### DIFF
--- a/macros/cross_db_utils/datatypes.sql
+++ b/macros/cross_db_utils/datatypes.sql
@@ -93,3 +93,18 @@
 {% macro bigquery__type_int() %}
     int64
 {% endmacro %}
+
+
+{# object  -------------------------------------------------     #}
+
+{%- macro type_object() -%}
+  {{ return(adapter.dispatch('type_object', 'dbt_utils')()) }}
+{%- endmacro -%}
+
+{% macro default__type_object() %}
+    object
+{% endmacro %}
+
+{% macro postgres__type_object() %}
+    bjson
+{% endmacro %}


joellabes:
Hi @vicmattos, thanks for opening this! Curious as to your use case - are you working on a project that covers both Snowflake and PGSQL? Can you tell me more? 

As far as compatibility with the other two warehouses: 
- From a quick Google, it looks like Redshift doesn't support objects; so instead it would be a `VARCHAR(64K)`: https://docs.aws.amazon.com/redshift/latest/dg/federated-data-types.html
- BQ is unclear, but it also doesn't seem to have a complex object type so I assume it would return a `string` as well: https://cloud.google.com/bigquery/docs/reference/standard-sql/json_functions 

Finally, new contributions to dbt-utils should have integration tests. What could we do to check that this behaves correctly?

vicmattos:
Hi @joellabes. 

My use case is that I have my production environment in Snowflake and my local development environment in PGSQL.

I created a sandbox environment with docker and vscode to test behaviour hypothesis for DBT: https://github.com/vicmattos/dbt-sandbox

About the complex datatypes in other warehouses:
- Found a complex data type in Redshift called SUPER, which it's definition looks like with an object type (https://docs.aws.amazon.com/en_us/redshift/latest/dg/r_SUPER_type.html)
- Thought that BQ had a Struct type that would be closer to the object type (https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type)

Regarding integration tests, I'm not sure. Actually this is my first OSS contribution, so I might do a lot of things wrong.

Thanks for your response and sorry for my late message.

joellabes:
@vicmattos thanks for following up, and appreciate you doing the extra legwork to find SUPER and STRUCT types; agreed that they're closer to an object. Could add the mappings for the other two as well? That'd be really useful to ensure that dbt-utils' cross-database capabilities remain strong. 

This is what I'm thinking re testing, let me know if it makes sense:
Could we create a serialised object string and deserialise it to an object/jsonb/super/etc, then try to access a value from it? 

If that works for you, happy to work with you to take a basic proof of concept query and turn it into a full integration test. Let me know! Also open to other approaches.

PR #440, opened 2021-11-09
Use the relation object passed into get_column_values, instead of making our own

joellabes:
Resolves #424 - Mixed-quoting styles on Snowflake created an invalid `Relation` using the adapter

This is a:
- [x] bug fix PR with no breaking changes — please ensure the base branch is `master`

## Checklist
- [ ] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
- [ ] I followed guidelines to ensure that my changes will work on "non-core" adapters by:
    - [ ] dispatching any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/star.sql))
    - [ ] using the `limit_zero()` macro in place of the literal string: `limit 0`
    - [ ] using `dbt_utils.type_*` macros instead of explicit datatypes (e.g. `dbt_utils.type_timestamp()` instead of `TIMESTAMP`
- [ ] I have updated the README.md (if applicable)
- [x] I have added tests & descriptions to my models (and macros if applicable)
- [x] I have added an entry to CHANGELOG.md

Issue #424, opened 2021-09-27
dbt_utils.get_column_values should use the Relation object it's given instead of making its own


fivetran-joemarkiewicz:
### Describe the bug

The `get_column_values` and `star` macros do not seem to properly work when the source is passed through as an argument and the `src.yml` is leveraging the non-default [quoting](https://docs.getdbt.com/reference/project-configs/quoting). Specifically for Snowflake.

### Steps to reproduce

Use [quoting](https://docs.getdbt.com/reference/project-configs/quoting) in your `src.yml`:
```yml
version: 2

sources:
  - name: salesforce
    schema: 'salesforce'
    database: mydatabase
    quoting: 
          database: false
          schema: true
          identifier: true

    tables:
      - name: fivetran_formula
      - name: test_order_c
```
With the above `src.yml` config if I attempt to use the `get_column_values` and `star` macro like below:
```sql
{%- set old_formula_fields = dbt_utils.get_column_values(source('salesforce', 'fivetran_formula'),'field') | upper -%}

select
{{ dbt_utils.star(source('salesforce','test_order_c'), except=old_formula_fields) }}
from {{ source('salesforce', 'test_order_c') }}
```
The resulting compiled output is completely empty.
```sql
select

from DATABASE."SCHEMA"."test_order_c"
```

### Expected results

If I remove the quoting config in the `src.yml` then I see the output as expected. The expected results would be to allow users that have to use the quoting config to still be able to use the combination of these macros.
```sql
select
    "FIELDS_I_WANT",
    "ETC"
from DATABASE."SCHEMA"."test_order_c"
```

### Actual results

Compiled output below.
```sql
select

from DATABASE."SCHEMA"."test_order_c"
```

### System information
**The contents of your `packages.yml` file:**
```yml
packages:
  - package: fivetran/salesforce_formula_utils
    version: [">=0.4.0", "<0.5.0"]
  - package: fivetran/fivetran_utils
    version: [">=0.2.0", "<0.3.0"]
  - package: dbt-labs/dbt_utils
    version: [">=0.7.0", "<0.8.0"]
```
**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [X] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
installed version: 0.20.2
   latest version: 0.20.2

Up to date!

Plugins:
  - bigquery: 0.20.2
  - snowflake: 0.20.2
[...]

PR #152, opened 2019-07-23
Resolve "get_column_values returns an error if passed a `ref`, but the model has not been previously built"

emilieschario:
Closes #119

This PR updates the get column values macro to take an optional `default` argument that can be used when the ref'd relation does not already exist. This is notable in cases when running `dbt compile`. 

Passing tests are: https://circleci.com/gh/fishtown-analytics/dbt-utils/330?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Issue #119, opened 2019-02-11
get_column_values returns an error if passed a `ref`, but the model has not been previously built


clrcrl:
## Steps to reproduce
1. Call `get_column_values` with a `ref` passed to it:
```
-- survey_guest_facts.sql
{% set slug_list = dbt_utils.get_column_values(table=ref('survey_responses'), column='slug') %}
...
```
2. Ensure the upstream model has not yet been built 
3. Run `dbt compile`

## Expected results
An empty list is returned during compilation, with no error

## Actual results
A database error occurs:
```
$ dbt compile
Found 95 models, 251 tests, 0 archives, 1 analyses, 236 macros, 7 operations, 1 seed files

13:49:01 | Concurrency: 1 threads (target='dev2')
13:49:01 |
Encountered an error:
Runtime Error
  Database Error in model survey_guest_facts (models/components/survey_guest_facts.sql)
    relation "cc_base.survey_responses" does not exist
```
From the logs:
```
2019-02-11 13:50:08,257 (Thread-1): Compiling model.jaffle_shop.survey_guest_facts
2019-02-11 13:50:08,272 (Thread-1): Acquiring new redshift connection "survey_guest_facts".
2019-02-11 13:50:08,272 (Thread-1): Re-using an available connection from the pool.
2019-02-11 13:50:08,272 (Thread-1): Using redshift connection "survey_guest_facts".
2019-02-11 13:50:08,272 (Thread-1): On survey_guest_facts: BEGIN
2019-02-11 13:50:08,298 (Thread-1): SQL status: BEGIN in 0.03 seconds
2019-02-11 13:50:08,299 (Thread-1): Using redshift connection "survey_guest_facts".
2019-02-11 13:50:08,299 (Thread-1): On survey_guest_facts: 

        select
            slug as value

        from "cc_base"."survey_responses"
        group by 1
        order by count(*) desc

        
2019-02-11 13:50:08,382 (Thread-1): Postgres error: relation "cc_base.survey_responses" does not exist

2019-02-11 13:50:08,382 (Thread-1): On survey_guest_facts: ROLLBACK
```

## System info
```
$ dbt debug
dbt version: 0.12.2
python version: 3.7.2
python path: /usr/local/Cellar/dbt/0.12.2/libexec/bin/python3.7
```
version of utils: 0.1.20

## Notes
- I'm interested in why this fails on `compile` but not on `run`.
[...]

PR #448, opened 2021-11-26
Regression: Correctly handle missing relations in get_column_values

### LINKED ISSUE #444: "if not target_relation" is not true even if relation does not exist in get_column_values
### Describe the bug
In `dbt_utils<0.7.4`, while using `get_column_values` macro, `dbt compile` on a dummy schema that doesn't exist in the database yet, returns the `default` values specified in the macro.
But, In `dbt_utils==0.7.4`, `dbt compile` is trying to query the column values even though the table does not exist yet and throws `schema <dummy_schema_1111> does not exist` error.

### Steps to reproduce
Step 1:
This is a sample model that we have. 
```sql
{{ config(materialized='view', enabled=true) }}

{%- set activity_relation= ref('stg_sf__activities') -%}

with pivot_activity_type as (

  select
    sf_account_id,
    reporting_date,
    {{
      dbt_utils.pivot(
        'activity_type',
        dbt_utils.get_column_values(
          activity_relation,
          'activity_type',
          default=['email', 'call']
        ),
      prefix= 'activity_type_'
      )
    }}

    from {{activity_relation}}

    group by sf_account_id, reporting_date
)

select * from pivot_activity_type
```

Step 2:
Set `schema` in `profiles.yml` to a dummy value.

Step 3:
run `dbt compile`

### Expected results
Previously the macro has:
```
{%- set target_relation = adapter.get_relation(database=table.database,
                                          schema=table.schema,
                                         identifier=table.identifier) -%} 
```
which sets `target_relation` to null when it doesn't exist and the code return default values `['email', 'call']` and the compilation succeeds without querying the database.

### Actual results
Now since the code is changed to:
```
{%- set target_relation = table -%}
```
it returns `schema does not exist` error.

### System information
```yml
packages:
  - package: dbt-labs/dbt_utils
     version: 0.7.4
```

- [X] redshift

```
installed version: 0.21.0
  latest version: 0.21.0

Up to date!

Plugins:
  - redshift: 0.21.0
  - postgres: 0.21.0
```

### Additional context
The code doesn't go to the below block even if the relation doesn't exist in the database:
```
        {%- elif not target_relation and default is not none -%}

          {{ log("Relation " ~ table ~ " does not exist. Returning the default value: " ~ default) }}

          {{ return(default) }}
```

Yes, I would love to contribute.

joellabes:
Resolves #444 

This is a:

## Description & motivation
The change I made in #440 to solve #424, in turn broke the fix introduced in #152/#119 to have a default when a model doesn't exist.

This:
- adds a test to validate behaviour when a model doesn't exist, to prevent future regressions. 
- goes back to the old approach of explicitly trying to load the relation (but using the convenience method) to check existence, instead of taking the not-None-ness of the input object at face value. 

Concerns: 
- I don't know why the integration tests aren't running, and I want them to! 
- The convenience method `load_relation` doesn't do anything differently to the old behaviour. I _think_ this means that anyone who has both #424 and #119 problems will still be stuck. But I haven't run into any of them yet, so we'll burn that bridge when we get to it. 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake

### DIFF
--- a/integration_tests/data/sql/data_get_column_values_dropped.csv
+++ b/integration_tests/data/sql/data_get_column_values_dropped.csv
@@ -1,0 +1,12 @@
+field
+a
+b
+c
+d
+e
+f
+g
+g
+g
+g
+g
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -56,7 +56,9 @@ seeds:
         +schema: events
       
       data_get_column_values_dropped:
-        +post-hook: "{% do adapter.drop_relation(this) %}"
+        # TODO: Ideally this would use the adapter, but it gives a "Tried to drop relation, but its type is null" error.
+        #+post-hook: "{% do adapter.drop_relation(this) %}"
+        +post-hook: "drop table {{ this }}"
 
 
     schema_tests:
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -54,6 +54,10 @@ seeds:
     sql:
       data_events_20180103:
         +schema: events
+      
+      data_get_column_values_dropped:
+        +post-hook: "{% do adapter.drop_relation(this) %}"
+
 
     schema_tests:
       data_test_sequential_timestamps:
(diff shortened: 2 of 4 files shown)


jasnonaz (approved):
Looks good~

Issue #450, opened 2021-11-28
Add grouping/partitioning to relevant schema tests


emilyriederer:
### Describe the feature

Many of the data checks offered by `dbt-utils` could benefit by having an optional `group` argument.  This would allow checks to be done separately for each subgroup of data so that success/failure could be analyzed at the group level. 

Specifically, I believe that this could be relevant to the following checks:

- [ ] equal_rowcount()
- [ ] recency()
- [ ] at_least_one()
- [ ] not_constant()
- [ ] cardinality_equality()
- [ ] unique_where()
- [ ] non_null_proportion()
- [ ] sequential_values() (mentioned in #447 )

The motivation of this is two-fold:

- Some checks can only be expressed by group (e.g. subscription dates are only sequential *within* a user)
- Some checks are more rigorous by group (e.g. globally a field could be non-constant but it could still be errantly constant within a group)

This issue is explored at greater length in [my blog post](https://emilyriederer.netlify.app/post/grouping-data-quality/). The post includes an example (NYC subway turnstile data) where these checks add incremental value. This is a follow up issue encouraged by @joellabes after reading the post.

### Describe alternatives you've considered

I recognize there are downsides to this approach, but I believe these can be easily mitigated. These include:

- **Computational cost of grouped checks could be exceedingly high**. This could be annoying for DataOps-style testing purposes. However, this could be mitigated by long-running tests being labeled so they could be excluded by `dbt test` in certain settings
- **API bloat**. Any new feature makes an overall tool more complex and less navigable. However, I think any SQL user is very comfortable thinking in terms of groups/partitions so this does not seem overly burdensome.
- **Backwards compatibility**. The reasonable default for these checks would be "no grouping", so this should not cause any backwards compatibility issues for existing users of the test modules.

Alternatives to this approach, to me, seem unappealing:

- **Users could write their own custom macros.** However, the functionality of these custom macros would by 90% the same as existing `dbt-utils` functionality and not allow users to benefit from the superior testing and maintenance of `dbt-utils`
- **Non-grouped checks could be added at an "upstream" point in the data pipeline.** In theory, at some point in the data generation process, data may be already being created within a group and could be tested before being unioned to the whole. However, this is not true in many cases due either to the data-generating process or the analytics engineer's ability to move infinitely upstream.
[...]

PR #451, opened 2021-11-29
Implement schema tests by group/partition (WIP - not ready for review)

emilyriederer:
This is a:

## Description & motivation

This PR adds checks by groups as discussed in #450 . In short, the motivation is that some checks cannot be expressed at all without subgrouping, and other checks can be more rigorous at the group level. 

## Checklist

*I will check off checklist items before asking for formal approval of this PR. This PR will consist of many small, similar pieces, and this currently contains only one such iteration.*

    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -38,6 +38,14 @@ models:
           field: today
           interval: 1
 
+  - name: test_recency_group
+    tests:
+    - dbt_utils.recency:
+        datepart: day
+        field: today
+        interval: 1
+        group_vars: ['var1', 'var2']
+
   - name: test_equal_rowcount
     tests:
       - dbt_utils.equal_rowcount:
--- a/integration_tests/models/schema_tests/test_recency_group.sql
+++ b/integration_tests/models/schema_tests/test_recency_group.sql
@@ -6,7 +6,7 @@ select
     'c' as var2,
     {{ dbt_utils.date_trunc('day', dbt_utils.current_timestamp()) }} as today
 
-union 
+union all
 
 select
     'b' as var1,
@@ -20,7 +20,7 @@ select
     'c' as var2,
     cast({{ dbt_utils.date_trunc('day', dbt_utils.current_timestamp()) }} as datetime) as today
 
-union
+union all
 
 select
     'b' as var1,
--- a/integration_tests/models/schema_tests/test_recency_group.sql
+++ b/integration_tests/models/schema_tests/test_recency_group.sql
@@ -1,0 +1,30 @@
+
+{% if target.type == 'postgres' %}
+
+select
+    'a' as var1,
+    'c' as var2,
+    {{ dbt_utils.date_trunc('day', dbt_utils.current_timestamp()) }} as today
+
+union 
+
+select
+    'b' as var1,
+    'c' as var2,
+    {{ dbt_utils.date_trunc('day', dbt_utils.current_timestamp()) }} as today
+
+{% else %}
+
+select
+    'a' as var1,
+    'c' as var2,
+    cast({{ dbt_utils.date_trunc('day', dbt_utils.current_timestamp()) }} as datetime) as today
+
+union
+
+select
+    'b' as var1,
+    'c' as var2,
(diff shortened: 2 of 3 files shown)


emilyriederer:
@joellabes - I will leave this PR marked as a Draft for the foreseeable future until I've added the argument to all the in-scope tests, updated the docs, etc. In the meantime, I've pushed an example of the `recency` test. At your convenience, I'd appreciate any feedback on the general approach / style since the others will likely be similar.

joellabes:
Hi @emilyriederer, super late for me to come back to you but I have just spent some time looking at this. I love it! Big fan 🎉  I thought this was going to be much fiddlier than it has turned out to be, which is always a win. 

Some answers to your questions: 

> * Are there internal "style" rules somewhere that I should reference when making an decision that affect the API?

Not written down anywhere, but in general it's good to have common arguments named the same way and in the same order where possible. As an example, the metrics package has some [secondary calculations](https://github.com/dbt-labs/dbt_metrics#secondary-calculations) which have arguments in common - they're all named the same way and in the same order, optimising globally as opposed to picking the perfect word to get each one's local maximum. 

> * Is there a better name than `group_vars` or a better way to provide grouping columns than a list of names (`['col_a', 'col_b']`)?

I'm leaning towards something like `group_by` or `group_by_columns`, since `vars` has an existing meaning in dbt. If there's an existing pattern on other tests then feel free to adopt that though. I think providing the list of column names is the best way to handle it 👍 

> * I see some tests-of-the-tests are set up using `seeds` and others using `models`. Are there guidelines when I should use which? (The recency check requires a current timestamp, so this one has to be a model. This is more of a general question for the others.)

You've nailed the difference - seeds are nicer than hand-crafted select statements, **where the data is static**. But when you need the current timestamp, it has to be a proper model. 

> * I could have changed the `test_recency` _model_ instead of making a new `test_recency_group` _model_. Is there a preference between "avoid duplication in tests" versus "map each model to exactly one test"?

Good question! I'm generally OK with the same _input_ file being used for multiple output files, if it can be achieved without needing to contort the tests or the models too much. E.g. I was just looking at PR #507 which has multiple tests built on top of the same seed file, by using the where conditions. But the window functions vs standard functions made more sense to be in different seed files. No hard and fast rule, but if you're working too hard to keep it in one file, or doing huge amounts of copy-paste, you're probably leaning too far in one direction or the other. 

> * Looking at the tests for `test_recency`, I see integration tests for schema tests that _should pass_. Is it standard to only test the "positive" case or is there a framework to also test for correct detection of failures?

We don't currently have a way to test for detection of failures 😢  The Core team are looking into improving testability of adapters at the moment, which will hopefully bleed over to packages as well (cc @jtcohen6)

And a question from me: 
- Why is your new model restricted to just postgres? Did you only have access to a Postgres database?

Thanks for your work on this so far 🤩

joellabes:
Also! A heads up that #521 will change a bunch of file names, so it's worth waiting for that to be merged before moving this any further forward

emilyriederer:
Thanks @joellabes for all of the helpful comments! I'll wait until #521 is merged then work on your updates and expanding to other tests. 🤓 

To answer your question, this PR shouldn't be limited to any one database since the functionality is bound to be pretty basic changes to insert `group by` statements. I only checked BigQuery because that's what I currently have access to for testing.

joellabes:
> To answer your question, this PR shouldn't be limited to any one database since the functionality is bound to be pretty basic changes to insert `group by` statements. I only checked BigQuery because that's what I currently have access to for testing.

I was meaning in relation to [this line](https://github.com/dbt-labs/dbt-utils/pull/451/files#diff-b08733bbfd62afb661e1d29f9781f7bfa9638d2f58a742d15cc64908b0b0fb43R2) `{% if target.type == 'postgres' %}`, which would mean on any other target the model would be empty, right?

emilyriederer:
Ah yes, sorry! I see what you mean now. I'll clean that up before submitting. I initially set out to write a more complicated test (to have failure cases along with successes) and thought I might need to do it differently for Postgres so I'd started the if/else skeleton. However, as things stand, it's superfluous because the `if` and `else` parts of the code are identical.

PR #453, opened 2021-12-02
dbt v1 compat

joellabes:
This is a:
- [x] a breaking change — please ensure the base branch is the latest `dev/` branch

## Description & motivation
🥳 ONE POINT OH IS COMING BABYYYYYYYYYY

🧐 New minor version which supports any 1.x.x version of Core. 

I'm not pleased that I have to do `require-dbt-version: [">=1.0.0-rc", "<2.0.0"]`. My original attempt was `require-dbt-version: "~=1.0.0"` and that didn't work:
```
(dbt-preprod) joel@Joel-Labes internal-analytics % dbt deps
15:25:45 | [ info  ] | Running with dbt=1.0.0-rc3
15:25:47 | [ error ] | Encountered an error:
Runtime Error
  Could not parse version "~=1.0.0"
```

Nor did `require-dbt-version: [">=1.0.0", "<2.0.0"]` because it didn't pick up the release candidates.

Open to a different approach! 
## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake


joellabes:
@jtcohen6 not sure where the config is for the version of dbt used in integration tests, can you give me a pointer?

integration_tests/dbt_project.yml
```diff
@@ -54,6 +55,12 @@ seeds:
     sql:
       data_events_20180103:
         +schema: events
+      
+      data_get_column_values_dropped:
+        # TODO: Ideally this would use the adapter, but it gives a "Tried to drop relation, but its type is null" error.
+        #+post-hook: "{% do adapter.drop_relation(this) %}"
+        +post-hook: "drop table {{ this }}"
```
jtcohen6, line 62: If you know that this is a table, you could try something like (pseudo-code):
```sql
    +post-hook: "{% do adapter.drop_relation(this.incorporate(type='table') %}"
```

dbt_project.yml
```diff
@@ -1,11 +1,11 @@
 name: 'dbt_utils'
-version: '0.7.0'
+version: '0.1.0'
 
-require-dbt-version: [">=0.20.0", "<=1.0.0"]
+require-dbt-version: [">=1.0.0-rc", "<2.0.0"]
```
jtcohen6, line 4: What do you think about switching this to `[">=1.0.0", "<2.0.0"]` once we have a final release of dbt-core v1.0.0? In which case, this feels fine for now

jtcohen6 (commented):
It's already installing the latest compatible prerelease version of `dbt-<adapter>`:
https://github.com/dbt-labs/dbt-utils/blob/db8a047474ffdade6113cbdde7e19c45ff3aae77/run_test.sh#L9

The problem is, our Circle containers are running on python 3.6, and `dbt-core==1.0.0rc` dropped support for python 3.6 (it's reaching end of life on Dec 23). So I think you'll just want to upgrade these container versions from `3.6.13` to something newer, e.g. `cimg/python:3.9.9` ([docs](https://circleci.com/developer/images/image/cimg/python)):

https://github.com/dbt-labs/dbt-utils/blob/db8a047474ffdade6113cbdde7e19c45ff3aae77/.circleci/config.yml#L8

dbt_project.yml
joellabes: Yeah I don't really want to have to release 0.8.1 in a week that just drops the release candidates. Since
- 0.7.5 supports <= 1.0 as its upper bound, which _does_ include the RCs, and
- 1.0 final is on the cusp of release anyway,

 I think I'm happy to say that 0.8 is only going to cover the final released version of 1.0

joellabes: With that said, that means that when I merge this in, Circle CI will give it a ❌ until 1.0 final actually ships. Do we feel OK about that? (cc @jasnonaz)

jasnonaz (approved):
1.0!!!

Approving this with failing tests as the tests are failing due to the fact that this version of utils will not be compatible with release candidates of dbt v1.0.0, only with the final version. We have a successful CI run [here](https://app.circleci.com/pipelines/github/dbt-labs/dbt-utils/602/workflows/b01da5d8-a5c1-4037-8265-211a5e0339a0). The only change made subsequent to this run was bumping the bound to be >1.0.0, which, as expected, caused the tests to fail.

Issue #357, opened 2021-04-28
get_relations_by_pattern no longer works with foreign data wrappers in postgres


eculicny:
### Describe the bug
While upgrading to 0.6.4 from 0.6.3, we hit an issue that is similar to #350 but for a postgres target that had foreign data wrappers set up while using `get_relations_by_pattern`

### Steps to reproduce
Setup an imported schema `extdb_extschema` in postgres target database with table that matches `'%_'+table_suffix`.

Reference in model to `get_relations_by_pattern`:
```
{% set relations = dbt_utils.get_relations_by_pattern(schema_pattern='extdb__extschema',
                                                      table_pattern='%_'+table_suffix,
                                                      database=this.database) %}
```

Run `dbt compile` on model for target database

### Expected results
The successfully retrieves the set of relations to iterate over creating a single model out of the records in the matching tables (e.g. `union all` of matching relations).

This is the current result when running dbt utils `0.6.3`

### Actual results
Unhandled runtime errors occur during the `get_relations_by_pattern` call causing the model to not build.

### Screenshots and log output
```
Unhandled error while executing model.test_model.consolidation
Field "type" of type typing.Union[dbt.contracts.relation.RelationType, NoneType] in dbt.adapters.postgres.relation.PostgresRelation has invalid value 'foreign'
```

### System information
**The contents of your `packages.yml` file:**
```
packages:
  - package: fishtown-analytics/dbt_utils
    version: 0.6.4
```

dbt version: `0.19.1`

**Which database are you using dbt with?**
- [x] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
installed version: 0.19.1
   latest version: 0.19.1

Up to date!

Plugins:
  - snowflake: 0.19.1
  - redshift: 0.19.1
  - bigquery: 0.19.1
  - postgres: 0.19.1
```

### Additional context
[The addition](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/get_tables_by_pattern_sql.sql#L11-L14) of `table_type` into the `default__get_tables_by_pattern_sql` macro causes problems with foreign tables in postgres since they have a type of `FOREIGN` in `information_schema`, but `dbt` doesn't support that type by that name.

### Are you interested in contributing the fix?
[...]

PR #476, opened 2022-01-12
Added case for handling postgres foreign tables...

Aesthet:
(tables which are external to current database and are imported into current database from remote data stores by using Foreign Data Wrappers functionality).

This is a:

## Description & motivation
As described in [this issue](https://github.com/dbt-labs/dbt-utils/issues/357) we are facing errors when we are using Postgres Foreign Data Wrappers functionality - in particular tables which are imported as part of IMPORT FOREIGN SCHEMA command. Those tables are external to database in which they are imported and they have type FOREIGN, which is causing errors during execution of get_relations_by_pattern macros which is getting values from get_tables_by_pattern_sql macros.

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [x] Snowflake

### DIFF
--- a/macros/sql/get_tables_by_pattern_sql.sql
+++ b/macros/sql/get_tables_by_pattern_sql.sql
@@ -11,6 +11,7 @@
             case table_type
                 when 'BASE TABLE' then 'table'
                 when 'EXTERNAL TABLE' then 'external'
+                when 'FOREIGN' then 'external'
                 when 'MATERIALIZED VIEW' then 'materializedview'
                 else lower(table_type)
             end as "table_type"
--- a/macros/sql/get_tables_by_pattern_sql.sql
+++ b/macros/sql/get_tables_by_pattern_sql.sql
@@ -8,13 +8,7 @@
         select distinct
             table_schema as "table_schema",
             table_name as "table_name",
-            case table_type
-                when 'BASE TABLE' then 'table'
-                when 'EXTERNAL TABLE' then 'external'
-                when 'FOREIGN' then 'external'
-                when 'MATERIALIZED VIEW' then 'materializedview'
-                else lower(table_type)
-            end as "table_type"
+            {{ dbt_utils.get_table_types_sql() }}
         from {{ database }}.information_schema.tables
         where table_schema ilike '{{ schema_pattern }}'
         and table_name ilike '{{ table_pattern }}'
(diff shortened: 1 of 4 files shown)


macros/sql/get_tables_by_pattern_sql.sql
```diff
@@ -11,6 +11,7 @@
             case table_type
                 when 'BASE TABLE' then 'table'
                 when 'EXTERNAL TABLE' then 'external'
+                when 'FOREIGN' then 'external'
```
Aesthet, line 14: Of course I could add separate foreign type, which would require adding new relation type here:
```
class RelationType(StrEnum):
    Table = 'table'
    View = 'view'
    CTE = 'cte'
    MaterializedView = 'materializedview'
    External = 'external'
```
and/or add this case specifically to `postgres__get_tables_by_pattern_sql` instead of `default__get_tables_by_pattern_sql`, but I think that would be an overkill.

joellabes: I agree that making a separate foreign type would be overkill, but I don't hate the idea of getting a focused `postgres__get_tables_by_pattern_sql` version of the macro, or at least a subset. 

What about a new macro that generates the case statement for different table types? Something like 
```
{% macro map_table_types(table_type) %}
  {{ return adapter.dispatch(yada yada) }} 
{% endmacro  %}

{% macro default__map_table_types() %}
 case table_type
   when 'BASE TABLE' then 'table'
   ...

{% endmacro %}

{% macro postgres__map_table_types() %}
 case table_type
   when 'BASE TABLE' then 'table'
   ...
   when 'FOREIGN' then 'external'

{% endmacro %}
```

and then get_tables_by_pattern_sql swaps out the case statement for a call to `map_table_types()` (I don't like the macro name; feel free to pick something better). 
Handling an adapter-specific piece of functionality that other adapters either shouldn't know about or might have an adverse reaction to in the future (living in fear of Snowflake one day adding a FOREIGN type which is some other wild thing 😱) feels like _exactly_ the sort of thing that multiple dispatch lends itself to with minimal overhead. 

How does that sound to you?

joellabes (changes requested):
Love it! I've added some thoughts based on your comments about how far to take it - I'm not _wed_ to this, but I think it's a good way to balance the competing needs of minimal copy-paste vs not exposing adapter-specific logic to everything else.

macros/sql/get_tables_by_pattern_sql.sql
Aesthet: Sounds reasonable to me, will update this PR today.

joellabes:
Wonderful @Aesthet! Please add yourself to the 0.8.1 CHANGELOG, then we'll get this merged in 🎉

joellabes (dismissed):

joellabes (approved):

PR #478, opened 2022-01-17
Delete rows with matching unique key

matthewrj:
## Description & motivation
This introduces "merge" behaviour for insert by period to match the standard incremental materialisation. This allows insert by period to be used with models where rows change over time.

I simply copied the code from the standard incremental materialisation here: https://github.com/dbt-labs/dbt-core/blob/8442fb66a591c4f5353f8baaaaf62f846c5f5171/core/dbt/include/global_project/macros/materializations/models/incremental/merge.sql#L50.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [x] Redshift
    - [ ] Snowflake

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# dbt-utils v0.9.0
+## Features
+- Add `unique_key` configuration option to `insert_by_period` materialization ([#478](https://github.com/dbt-labs/dbt-utils/pull/478))
+
 # dbt-utils v0.8.0
 ## 🚨 Breaking changes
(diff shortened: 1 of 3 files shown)


joellabes:
@matthewrj Fancy! Thanks for opening this. It'd be nice to have this process support deletes 👍 

Two conflicting thoughts here: 
1. I'd love to see some tests around this to ensure that it behaves correctly. I think this would be a series of seeds with different stages of progress, but I don't know how to cause it to re-run without multiple invocations of the automated tests, which feels a bit fragile...
1. We're still thinking about the future of this macro (see #410), so I don't want to ask you to do a bunch of work that might go unmerged for a while. 

For now, please don't be offended if this sits on ice for a bit - we're about to dive into the future of dbt-utils with a vengeance, so I expect there to be some progress in the next month or two.

PR #479, opened 2022-01-20
Remove column name in generated column

armandduijn:
This is a:
- [X] bug fix PR with no breaking changes — please ensure the base branch is `master`

## Description & motivation
The sequential_values test internally generates a column name to keep track of the previous value. Generating this name causes errors when using the [quote property](https://docs.getdbt.com/reference/resource-properties/quote/).

This PR simplifies the column name that tracks the previous value. I've prefixed the name with an underscore "_" to reduce the risk of conflicts.

I'm not sure if this is a breaking change. Do users parse the `previous_{{ column_name }}` name when examining the test errors?

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [X] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/schema_tests/sequential_values.sql
+++ b/macros/schema_tests/sequential_values.sql
@@ -12,7 +12,7 @@ with windowed as (
         {{ column_name }},
         lag({{ column_name }}) over (
             order by {{ column_name }}
-        ) as previous_{{ column_name }}
+        ) as _previous
     from {{ model }}
 ),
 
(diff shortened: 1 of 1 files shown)


joellabes:
@armandduijn thanks for calling this out and providing a fix! 

> I've prefixed the name with an underscore "_" to reduce the risk of conflicts.

I _think_ this will work. I can't imagine that `_previous` is a common column name, but I'd feel more comfortable if we did something like using the [slugify](https://github.com/dbt-labs/dbt-utils#slugify-source) macro instead: `as previous_{{ dbt_utils.slugify(column_name) }}`. 

With that in mind, it might be nicer/more ergonomic to set that as a variable once and then pass it around, as opposed to having nested macros in the datediff macro for example. 

```
{% set previous_column_name = "previous_" ~ dbt_utils.slugify(column_name) %}

select ... as {{ previous_column_name }}
```

What do you think?

armandduijn:
That makes perfect sense. Let me test it tomorrow and I'll mark the PR as ready for review once I'm happy.

joellabes (approved):

PR #481, opened 2022-01-23
add support to rename columns inside of union_relations

chaimt:
This is a:
- [ X ] new functionality — please ensure the base branch is the latest `dev/` branch

## Description & motivation
There are times that two tables have the same column but under different names, and the union does not combine them. 
I want the option to rename a column and take it into consideration during the union.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ X] Databricks 

### DIFF
--- a/macros/sql/union.sql
+++ b/macros/sql/union.sql
@@ -1,10 +1,6 @@
-{%- macro union_relations(relations, column_override=none, include=[], exclude=[], source_column_name='_dbt_source_relation') -%}
-    {{ return(adapter.dispatch('union_relations', 'dbt_utils')(relations, column_override, include, exclude, source_column_name)) }}
-{% endmacro %}
+{%- macro union_relations_ex(relations, column_override=none, include=[], exclude=[], source_column_name='_dbt_source_relation', rename_columns=none) -%}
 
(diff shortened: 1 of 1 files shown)


joellabes:
Hi @chaimt, thanks for opening the PR! Could you tell me more about your use case? Do you have a _lot_ of relations with inconsistent names, or was this a one-off problem? 

I'm cautious about adding too much complexity to these functions' signatures without a compelling narrative, because of the flow-on effects to testing and ensuring that every permutation of behaviour works reliably.

chaimt:
we have a lot of tables (fact tables) that need to be in a union, and have common field but not the same name (like the timestamp of the metric).
I don't think it is wise to add another view just to fix the column names.

joellabes:
Thanks for clarifying @chaimt! I discussed this internally with our professional services team (who have exposure to the most dbt projects in the wild) and they haven't come across a broader use case for this, so I'm going to close this without merging it. 

Of course you're welcome to continue to use your version of the macro in your project, and if other community members come along in the future who would find value in this being part of utils then I'd be happy for an issue to be reopened!

Issue #482, opened 2022-01-24
BUG - datediff does not work if first_date and second_date are timestamptz


kpounder:
### Describe the bug
As seen here (https://github.com/fivetran/dbt_github/issues/25), `datediff` returns the following error on Redshift if `first_date` and `second_date` are `timestamptz`. 

```
Database Error in model github__issues (models/github__issues.sql)
  function pg_catalog.date_diff("unknown", timestamp with time zone, timestamp with time zone) does not exist
  HINT:  No function matches the given name and argument types. You may need to add explicit type casts.
  compiled SQL at target/run/github/models/github__issues.sql
```

### Steps to reproduce
1. Create a table with two columns, `col1` and `col2` of type `timestamptz`
2. Try adding a column that calculates the difference using `datediff`, e.g., `{{ dbt_utils.datediff('col1::timestamp', 'col2::timestamp', 'hour')}}`

### Expected results
The difference should be calculated. 

### Actual results
Redshift returns the above error message. 

### System information
**The contents of your `packages.yml` file:**
```
yml
packages:
  - package: dbt-labs/dbt_utils
    version: 0.8.0
  - package: fivetran/github
    version: [">=0.4.0", "<0.5.0"]
```

**Which database are you using dbt with?**
- [ ] postgres
- [x] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
1.0.1
```

### Additional context
I think the following would fix the issue. I'll try to raise a PR with this fix. 
```
{% macro redshift__datediff(first_date, second_date, datepart) %}

    datediff(
        {{ datepart }},
        {{ first_date }}::timestamp,
        {{ second_date }}::timestamp
        )

{% endmacro %}
```

### Are you interested in contributing the fix?
Yes, I am. I'll try to raise a PR.

PR #483, opened 2022-01-24
add casting

kpounder:
This is a:

## Description & motivation
Redshift's `datediff` function does not support the difference between two `timestamptz` (for example, see https://github.com/dbt-labs/dbt-utils/issues/482). In order to handle that case, I would like to add casting to `timestamp`. 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [x] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/cross_db_utils/datediff.sql
+++ b/macros/cross_db_utils/datediff.sql
@@ -61,6 +61,10 @@
 {# redshift should use default instead of postgres #}
 {% macro redshift__datediff(first_date, second_date, datepart) %}
 
-    {{ return(dbt_utils.default__datediff(first_date, second_date, datepart)) }}
+    datediff(
+        {{ datepart }},
+        {{ first_date }}::{{ dbt_utils.type_timestamp()}},
+        {{ second_date }}::{{ dbt_utils.type_timestamp()}}
+        )
 
 {% endmacro %}


kpounder:
This is my first time contributing to dbt, so please give this a thorough look. Don't want to break anything. :)

joellabes:
@kpounder thanks for opening this! I'm not _convinced_ that we want to silently strip timezone information from a column. My gut feeling is that this should be explicitly handled in the modelling layer (it looks like Fivetran has shipped a change to this effect), or it can also be handled in the macro call, as you did in your sample code: 

`{{ dbt_utils.datediff('col1::timestamp', 'col2::timestamp', 'hour')}}`

What do you think? In other places in the repo we've upcast columns, but not downcast as far as I can remember.

kpounder:
After discussion with @joellabes , abandoning this PR. Here's why: 

Redshift's `datediff` doesn't support either timestamp being `timestamptz` (per their docs). If both were `timestamptz`, then casting to `timestamp` before taking the difference would be essentially harmless. But if one is `timestamp` and the other is `timestamptz`, then casting to `timestamp` before taking the difference is not safe. 

Workaround: If you have both `timestamptz`, then you can just do the casting yourself first, e.g., `{{ dbt_utils.datediff('col1::timestamp', 'col2::timestamp', 'hour')}}`. Otherwise, be careful to do what is appropriate. :)

PR #494, opened 2022-02-10
fix pivot : in pivoted column value, single quote must be escaped (on…

mdutoo:
… postgresql)

else ex. syntax error near : when color = 'blue's'

This is a:
- [x ] bug fix PR with no breaking changes — please ensure the base branch is `master`

## Description & motivation
pivot script gets a syntax error when working with pivoted column values that contain quotes, because it quotes them but did not escape them. This fix quotes them.

## Checklist
- [ x] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [ x] Postgres
    - [ ] Redshift
    - [ ] Snowflake
- [ x] I have added tests & descriptions to my models (and macros if applicable)

### DIFF
--- a/integration_tests/data/sql/data_pivot.csv
+++ b/integration_tests/data/sql/data_pivot.csv
@@ -1,4 +1,5 @@
 size,color
 S,red
 S,blue
+S,blue's
 M,red
--- a/integration_tests/data/sql/data_pivot_expected.csv
+++ b/integration_tests/data/sql/data_pivot_expected.csv
@@ -1,3 +1,3 @@
 size,red,blue,blue's
-S,1,0,0
-M,1,1,1
+S,1,1,1
+M,1,0,0
--- a/integration_tests/data/sql/data_pivot_expected.csv
+++ b/integration_tests/data/sql/data_pivot_expected.csv
@@ -1,3 +1,3 @@
-size,red,blue
-S,1,1
-M,1,0
+size,red,blue,blue's
+S,1,0,0
+M,1,1,1
--- a/integration_tests/dbt_project.yml
+++ b/integration_tests/dbt_project.yml
@@ -59,6 +59,9 @@ seeds:
       data_get_column_values_dropped:
         # this.incorporate() to hardcode the node's type as otherwise dbt doesn't know it yet
         +post-hook: "{% do adapter.drop_relation(this.incorporate(type='table')) %}"
+      
+      data_pivot_expected:
+        quote_columns: true
 
 
     schema_tests:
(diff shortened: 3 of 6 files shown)


joellabes:
Hi @mdutoo, thanks for opening the PR! I'd be happy to merge this in once it's stable, but the CI tests are currently failing; you can review them by clicking the Details link above. 

Let me know if you need any help debugging!

mdutoo:
Hi @joellabes , my bad, I had tried out one too many things and ended up not putting it back the right way.
So now CI tests work on postgres (and also redshift seemingly). But alas still not on biggquery and snowflake.

That's because my fix is about escaping the single quote, and it solves it by doubling it, which probably don't work on bigquery and snowflake. For instance, a quick search shows that snowflake rather requires backslashing it.

Would you be able to adapt the fix to these other databases ?

Otherwise, I don't know as much about these other databases, but I could try to adapt the fix nonetheless. However, for that I need to know how you would integrate code specific to one of them. I've heard of dbt "adapters" but don't know much about it, but maybe you already have it plugged somewhere in dbt-utils ?

mdutoo:
@joellabes much thanks, I've learned a lot of things !

Well, seems now that the Markupsafe version used is now 2.1.0 rather than 2.0.1 and that this upgrade (not mine) broke dbt startup :

File "/home/********/project/venv/lib/python3.9/site-packages/jinja2/filters.py", line 13, in <module>
    from markupsafe import soft_unicode
ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/home/********/project/venv/lib/python3.9/site-packages/markupsafe/__init__.py)

joellabes:
@mdutoo thanks for updating that! The Markupsafe issue caught us out, but luckily dbt core 1.0.2 was already at release candidate status so we shipped the fix same day - sorry that you got trampled by it in the interim. 

When I re-ran the tests, a couple of things conspired against me with your new changes: 
- turns out you need two backslashes to escape the slash in Jinja and make it flow through to the SQL layer
- BigQuery doesn't allow apostrophes in column names, and Snowflake has finnicky case-sensitive behaviour when quoted

I made some changes based off your branch: specifically, I added an apostrophe-specific version of the tests which doesn't quote the tables, and also configured the test so that the slugify macro will strip out the apostrophe in the column alias. 

Normally I can push them directly to the branch as a maintainer of the project so that everything lives in the one place. GitHub couldn't find your forked version anymore though :( your work is now sitting at #503 so I'm going to close this PR and wrap up over there. 

Thanks for your patience along the way!

mdutoo:
Sorry that there was some work left. But well, that's the kind of stuff to be expected with escaping bug...
So great that you could put it all in order, thank you for your efforts ! And congrats for the great dbt-utils package.

PR #502, opened 2022-02-18
utils 0.8.1

joellabes:
## Description & motivation
Next patch version of dbt-utils, with backwards-compatible changes and new functionality. 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake


macros/cross_db_utils/escape_single_quotes.sql
```diff
@@ -0,0 +1,15 @@
+{% macro escape_single_quotes(expression) %}
+      {{ return(adapter.dispatch('escape_single_quotes', 'dbt_utils') (expression)) }}
+{% endmacro %}
+
+{% macro default__escape_single_quotes(expression) -%}
+{{ expression | replace("'","''") }}
```
jasnonaz, line 6: not blocking but we might want to explain these arcane escaping rules just a tad

macros/sql/pivot.sql
```diff
@@ -69,7 +69,7 @@ Arguments:
     {{ agg }}(
       {% if distinct %} distinct {% endif %}
       case
-      when {{ column }} {{ cmp }} '{{ v }}'
+      when {{ column }} {{ cmp }} '{{ dbt_utils.escape_single_quotes(v) }}'
```
jasnonaz, line 72: not blocking but while we're in here may want to rename 'v' to something more comprehensible

jasnonaz (approved):
Stellar work - thanks for getting this together! ⭐

Issue #490, opened 2022-02-07
Allow expression is true to use window functions (and maybe more)


jpmmcneill:
### Describe the feature
Currently, you can't have something like:

```
schema.yml

version:2
models:
  - name: my_model
    columns:
      - name: id
      - name: test_col_1
      - name: test_col_2
        tests:
        - dbt_utils.expression_is_true:
          - expression: "> sum(test_col_1) over (partition by id)"
```

due to how expression_is_true is implemented (namely, it evaluates the expression in the context of a where clause.

https://github.com/dbt-labs/dbt-utils/blob/51ed999a44fcc7f9f502be11e5f190f5bc84ba4b/macros/schema_tests/expression_is_true.sql#L16:L20

While I hate suggesting solutions in issues, this shouldn't be too hard to address. For example, a simple enough solution to this is to define a new column based on `expression` that would then get filtered out.

I haven't run into any other DB errors for this kind of problem, but this could certainly effect other functions that can't be evaluated inside where clauses.

### Describe alternatives you've considered
A workaround is obviously to define a new field in the model being tested as a window function and then use that field in the expression is true. However, depending on the number of columns being tested, this can be very painful.

### Who will this benefit?
People who don't want to have to write new fields in a table to use the expression is true test.

### Are you interested in contributing this feature?
Definitely, this would be very quick in principle.

jpmmcneill:
I'm aware that this isn't well due to the discussion over at https://github.com/dbt-labs/dbt-utils/discussions/487

joellabes:
@jpmmcneill I like your proposed solution! I'd welcome a PR to that effect 👍 

Bonus points if it also comes with a new integration test to check that it works correctly and doesn't break existing expressions. 

The one thing that I've got in the back of my mind is that it needs to work as both a column-level and a model-level test. I think that would still be possible, but in your case you I _think_ you might have to define your window function as a model-level test (not column-level) along these lines: 

```
schema.yml

version:2
models:
  - name: my_model
    tests:
      - dbt_utils.expression_is_true:
        - expression: "col_2 > sum(test_col_1) over (partition by id)"
    columns:
      - name: id
      - name: test_col_1
      - name: test_col_2
```

I'll leave the specifics to you to dig into, but let me know if you need to noodle on anything.

jpmmcneill:
Hey @joellabes. I got around to changing this today, pr above ☝️. There is one thing I'm a bit apprehensive about - i've detailed that in a coment.

[...]

PR #507, opened 2022-02-27
Jmcneill/expression is true tweak

jpmmcneill:
This is a:

## Description & motivation
Adjust the way that the `expression_is_true` test gets evaluated (basically move it outside of a where clause).
Issue context: https://github.com/dbt-labs/dbt-utils/issues/490
This allows for use of window functions at a column level, and maybe a bunch of other stuff.
Tests have been added window functionality.

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake


jpmmcneill:
I accidentally based on main rather than next/minor. Happy to change this if needed.

I guess that is the reason for https://github.com/dbt-labs/dbt-utils/pull/507/commits/a7f4f51fe1d9883150f01f623e83f1ab197efcbb being included in the diff.

integration_tests/models/schema_tests/schema.yml
```diff
@@ -31,6 +31,21 @@ models:
...
+          expression: max(col_c) over (partition by col_a) = min(col_c) over (partition by col_a)
+      - dbt_utils.expression_is_true:
+          expression: sum(col_b) over (partition by col_a) = 15
+          condition: col_a = 1
+    columns:
+      - name: col_c
+        tests:
+          - dbt_utils.expression_is_true:
+              expression: = sum(col_b) over (partition by col_a)
+
```
jpmmcneill, line 48: Test that window functions are behaving as expected

integration_tests/models/schema_tests/schema.yml
```diff
@@ -13,7 +13,7 @@ models:
         tests:
           - dbt_utils.at_least_one
 
-  - name: data_test_expression_is_true
+  - name: data_test_expression_is_true_1
```
jpmmcneill, line 16: Migrate old seed to new one

macros/schema_tests/expression_is_true.sql
```diff
@@ -7,16 +7,18 @@
...
 with meet_condition as (
-    select * from {{ model }} where {{ condition }}
+    select
+      *,
+      {% if column_name is none %}
+      {{ expression }}
+      {%- else %}
+      {{ column_name }} {{ expression }}
+      {%- endif %}
+      as _test_expression_passed
```
jpmmcneill, line 17: The one thing I'm a bit nervous about here is a field name collision.

A possible solution is to grab the column names from the given model and throw a compiler message if _test_expression_passed is one of the field names.

However, that seems a bit contrived to me. Another possibility is just make the field name more constructed (ie. maybe encode the expression in it etc). Would like to hear peoples thoughts :)

jpmmcneill (commented):
Adding notes for changes in comments

macros/schema_tests/expression_is_true.sql
jpmmcneill: I decided to catch this exception in: https://github.com/dbt-labs/dbt-utils/pull/507/commits/68dc34114e6636ee23495e72bed893c002a57ac4.

Let me know if its a bit expensive to use get columns in relation to catch this error, or if people should be left to fend for themselves when seeing "ambiguous column name" errors.

jpmmcneill:
Hey @joellabes - you may have missed this. Nudging as it's ready for review :)

integration_tests/models/schema_tests/schema.yml
joellabes: This is a great set of tests 😍

macros/schema_tests/expression_is_true.sql
joellabes: It's a fair concern! My gut feeling is that this is a reasonably obscure column name. Perhaps you could include the package name:
```suggestion
      as _dbt_utils_test_expression_passed
```

I don't think it's necessary to protect against the ambiguous column name, just selecting * should be OK.

macros/schema_tests/expression_is_true.sql
```diff
@@ -6,17 +6,28 @@
...
 
-select
-    *
-from meet_condition
-{% if column_name is none %}
-where not({{ expression }})
-{%- else %}
-where not({{ column_name }} {{ expression }})
-{%- endif %}
+select * from meet_condition where not(_test_expression_passed)
```
joellabes, line 31:
```suggestion
select * from meet_condition 
where not(_test_expression_passed)
```

integration_tests/models/schema_tests/schema.yml
joellabes: Could you name the seeds with a bit more clarity? perhaps `data_test_expression_is_true` (i.e staying as-is) and `data_test_expression_is_true_window_functions` or something?

joellabes (changes requested):
I like where this is going 😍 ! A couple of nitpicks and comments below.

A side note: Since this will change the shape of the results table, which _could_ be a breaking change for anyone who uses `--store-failures`, this will likely come out in 0.9.0 as opposed to a patch release. 

I'm OK with that though - we _could_ exclude the new column with `star`, but I think it's somewhere between `not harmful` and `helpful` to have the true/false result of the evaluation included.

integration_tests/models/schema_tests/schema.yml
jpmmcneill: Happy to. If this is something that you don't like, this pattern of seed (`*_1.csv`, `*_2.csv` etc...) is present in a few other test files FYI

macros/schema_tests/expression_is_true.sql
jpmmcneill: Will do & i'll remove the get column in relation check.

jpmmcneill:
@joellabes thanks for the review! I've updated with your comments in https://github.com/dbt-labs/dbt-utils/pull/507/commits/24ddca972e810fab8a041e01527de3adcbb4915c

> A side note: Since this will change the shape of the results table, which could be a breaking change for anyone who uses --store-failures, this will likely come out in 0.9.0 as opposed to a patch release.
I'm OK with that though - we could exclude the new column with star, but I think it's somewhere between not harmful and helpful to have the true/false result of the evaluation included.

On this, I'm certainly on team "it's helpful".

If i'm debugging a test, i'm always going to grab the compiled test code, run it and i'll want to see the result of some adjustments to pin down whats happening. In that case i'd like to see the result of the test to visualise how different things impact the result in something like a query editor - in my experience the fastest way to do this is with the test result as a column.

joellabes (approved):
Beautiful! Thanks @jpmmcneill - keep an eye out for this to come in utils 0.9.0 in the next wee while. If you need it before then, you can always install the package using git and pointing to the `next/minor` branch in the meantime.

jpmmcneill:
brill, thank you Joel! 🤜 🤛

PR #516, opened 2022-03-11
add macro to get columns

patkearns10:
This is a:

## Description & motivation

Similar to `dbt_utils.get_column_values`, except this pulls column names, rather than the values within a column. Like `dbt_utils.star`, but as a list, to be iterated over.

This is basically `dbt_utils.star`, but removing some of the last steps.

Unsure what the purpose of the top couple lines of star macro are:
```
{% macro star(from, relation_alias=False, except=[], prefix='', suffix='') -%}
    {{ return(adapter.dispatch('star', 'dbt_utils')(from, relation_alias, except, prefix, suffix)) }}
{% endmacro %}

{% macro default__star(from, relation_alias=False, except=[], prefix='', suffix='') -%}
    {%- do dbt_utils._is_relation(from, 'star') -%}
    {%- do dbt_utils._is_ephemeral(from, 'star') -%}
```
So I left those out. But I assume maybe that is important? 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake


README.md
```diff
@@ -585,6 +586,30 @@ Arguments:
 ...
 ```
 
+#### get_columns ([source](macros/sql/get_columns.sql))
+This macro returns the unique columns for a given [relation](https://docs.getdbt.com/docs/writing-code-in-dbt/class-reference/#relation) as a comma-separated list.
```
joellabes, line 590:
```suggestion
This macro returns the unique columns for a given [relation](https://docs.getdbt.com/docs/writing-code-in-dbt/class-reference/#relation) as a Jinja list
```

Right? From what I can see it's returning an iterable list of items, not a comma-separated values that you can drop directly into a query.

joellabes (changes requested):
Hey @patkearns10, thanks for opening this! 

>Unsure what the purpose of the top couple lines of star macro are:
```
{% macro star(from, relation_alias=False, except=[], prefix='', suffix='') -%}
    {{ return(adapter.dispatch('star', 'dbt_utils')(from, relation_alias, except, prefix, suffix)) }}
{% endmacro %}

{% macro default__star(from, relation_alias=False, except=[], prefix='', suffix='') -%}
    {%- do dbt_utils._is_relation(from, 'star') -%}
    {%- do dbt_utils._is_ephemeral(from, 'star') -%}
```

The first chunk is to enable multiple dispatch: https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch. In short, they allow other adapters to provide an override version of this macro, if they need to. 

The is_relation and is_ephemeral blocks then check that the thing you passed in is a real ref/source object, and that it's not ephemeral (because CTEs can't be used with this, because they don't go into the info schema). 

So you do need all parts of that 😬 

I'd be _very_ interested in whether the `star` macro could call this new macro directly, instead of having two copies of the exclude code etc. I haven’t looked at whether that's possible or not, but it feels reasonable - would love if you had a go as an extra for experts!

README.md
patkearns10: Yes, correct. I was thinking in python terms, it returns a list object.
I reworded that to make it more clear.

patkearns10:
Okay I moved some things around! 
the `star` macro now references the `get_columns` macro.

tested it works locally, by recreating the macros and removing the dispatch logic. How do I test the dispatch logic?

`get_columns` macro:
Screen Shot 2022-03-23 at 10 15 29 AM

call `get_columns`:
Screen Shot 2022-03-23 at 10 30 15 AM

call `get_columns` with except:
Screen Shot 2022-03-23 at 10 30 39 AM

replace within  local `star` macro:
Screen Shot 2022-03-23 at 10 15 20 AM

call local `star` macro:
Screen Shot 2022-03-23 at 10 15 07 AM

call local `star` macro with except:
Screen Shot 2022-03-23 at 10 38 15 AM

macros/sql/get_columns.sql
```diff
@@ -0,0 +1,25 @@
...
+    {%- do dbt_utils._is_relation(from, 'get_columns') -%}
+    {%- do dbt_utils._is_ephemeral(from, 'get_columns') -%}
+
+    {# -- Prevent querying of db in parsing mode. This works because this macro does not create any new refs. #}
+    {%- if not execute -%}
+        {{ return('') }}
+    {% endif %}
+
+    {%- set include_cols = [] %}
+    {%- set cols = adapter.get_columns_in_relation(from) -%}
```
patkearns10, line 15: @joellabes where does this `get_columns_in_relation` come from? I cannot find it.

joellabes (changes requested):
This is looking great :chefs_kiss: Just a couple of naming/doc tweaks for clarity. As for how to test the dispatched stuff, you'd want to add a dbt Singular test which validates the behaviour you expect. 

Something along similar lines to how codegen is tested: https://github.com/dbt-labs/dbt-codegen/blob/main/integration_tests/tests/test_generate_source.sql - I don't know if there's a way to compare two jinja lists' contents, or whether you'd have to turn it into a comma-separated string and compare that to an expected value 🤷

README.md
```diff
@@ -587,21 +587,15 @@ This macro returns the unique values for a column in a given [relation](https://
 ```
 
 #### get_columns ([source](macros/sql/get_columns.sql))
-This macro returns the unique columns for a given [relation](https://docs.getdbt.com/docs/writing-code-in-dbt/class-reference/#relation) as a comma-separated list.
+This macro returns an iterable Jinja list of columns for a given [relation](https://docs.getdbt.com/docs/writing-code-in-dbt/class-reference/#relation)
 
 **Args:**
 - `from` (required): a [Relation](https://docs.getdbt.com/reference/dbt-classes#relation) (a `ref` or `source`) that contains the list of columns you wish to select from
 - `except` (optional, default=`[]`): The name of the columns you wish to exclude
```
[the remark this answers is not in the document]
patkearns10: For posterity:

patkearns10:
## Added tests, documentation, and a new (opt-in) argument for lowercase output:

### star UPPER:
Screen Shot 2022-03-24 at 1 41 15 PM

### star lower:
Screen Shot 2022-03-24 at 1 40 55 PM

### get_filtered_columns UPPER:
Screen Shot 2022-03-24 at 1 44 37 PM

### get_filtered_columns lower:
Screen Shot 2022-03-24 at 1 42 11 PM

### integration tests:
Screen Shot 2022-03-24 at 1 39 45 PM

integration_tests/tests/sql/test_get_filtered_columns_in_relation.sql
```diff
@@ -0,0 +1,5 @@
+{% set actual_output = dbt_utils.get_filtered_columns_in_relation(from= ref('data_star_uppercase_columns'), except=['field_1'], output_lower=True) %}
+
+{% set expected_output = ['field_2', 'field_3'] %}
+
+{{ assert_equal_values (actual_output | trim, expected_output | trim) }}
```
joellabes, line 5: This is cool 🔥  out of curiosity, is object equivalency in Jinja/Python order-dependent? i.e. would `['field_3', 'field_2']` pass? I don't have strong feelings about whether it should or not (if you pushed me on it, I'd say they're _not_ equal), but feels important to know.

joellabes, line 5: This is cool 🔥  out of curiosity, is object equivalency in Jinja/Python order-dependent? i.e. would `['field_3', 'field_2']` pass? I don't have strong feelings about whether it should or not (if you pushed me on it, I'd say they're _not_ equal), but feels important to know.

integration_tests/tests/sql/test_star_output_lower.sql
```diff
@@ -0,0 +1,7 @@
+{% set actual_output = dbt_utils.star(from=ref('data_star_uppercase_columns'), except=['field_1', 'field_2'], prefix='test_', output_lower=True ) %}
+
+{% set expected_output %}
+"field_3" as "test_field_3"
+{% endset %}
+
+{{ assert_equal_values (actual_output | trim, expected_output | trim) }}
```
joellabes, line 7: For macros that return tables (e.g. star), we tend to build a seed that is the expected shape, and a model that should make that shape, and compare the two of them. It means that we're checking the whole thing works end to end and you don't accidentally wind up with an extra comma somewhere you shouldn't etc. 

_I suspect that if we get rid of the lowercase setting on this, then this becomes a moot point as we already have tests for star and can probably get rid of the whole thing_

joellabes (changes requested):
Big fan overall! My one complaint is the customisable lowercasing - since you can just pass it into the `lower` filter, I don't think adding two extra params adds enough value. I talk about that more on one of the specific comments - lmk what you think

integration_tests/macros/assert_equal_values.sql
```diff
@@ -0,0 +1,32 @@
...
+
+    {% endset %}
+
+    {{ log(msg, info=True) }}
+
+    select 'fail'
+
+{% else %}
+
+    select 'ok' limit 0
```
joellabes, line 29: This should be [`{{ limit_zero() }}`](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/macros/limit_zero.sql) - not all databases support limit clauses (notably [T-SQL](https://github.com/dbt-msft/tsql-utils/blob/main/integration_tests/dbt_utils/macros/limit_zero.sql))

README.md
```diff
@@ -585,6 +586,30 @@ Arguments:
...
+- optionally exclude columns
+- the input values are not case-sensitive (input uppercase or lowercase and it will work!)
+- declare whether you want lowercase output values with `output_lower=TRUE` (as Snowflake by default will output values in UPPERCASE)
+> Note: The native [`adapter.get_columns_in_relation` macro](https://docs.getdbt.com/reference/dbt-jinja-functions/adapter#get_columns_in_relation) allows you 
+to pull column names in a non-filtered fashion, also bringing along with it other (potentially unwanted) information, such as dtype, char_size, numeric_precision, etc.
+
+**Args:**
+- `from` (required): a [Relation](https://docs.getdbt.com/reference/dbt-classes#relation) (a `ref` or `source`) that contains the list of columns you wish to select from
+- `except` (optional, default=`[]`): The name of the columns you wish to exclude. (case-insensitive)
+- `output_lower` (optional, default=`False`): True if you want lowercase outputs. False if you want outputs to be database default (which may be upper or lower depending on database)
```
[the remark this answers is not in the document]
patkearns10: I have an unnatural aversion to capital letters, but, I didn't realize passing into lower was that easy, so I would be happy to remove that part. No harm done!

integration_tests/macros/assert_equal_values.sql
```diff
@@ -0,0 +1,32 @@
...
+
+    -----------
+    Actual:
+    -----------
+    --->{{ actual_object }}<---
+
+    -----------
+    Expected:
+    -----------
+    --->{{ expected_object }}<---
```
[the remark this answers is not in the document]
patkearns10: Your love is misplaced for whomever I stole this from!

integration_tests/tests/sql/test_get_filtered_columns_in_relation.sql
patkearns10: They are not equal. If you did want them to be treated that way, you would need to do something like: 
```
sorted(expected_output) == sorted(actual_output)
```
Screen Shot 2022-03-26 at 9 21 09 AM

integration_tests/macros/assert_equal_values.sql
patkearns10: Done!
@joellabes -- this is no longer required, as I switched to using the seed/model method. Should I remove this file then?

patkearns10:
Removed lowercase and swapped to different testing method. 

Screen Shot 2022-03-26 at 12 03 52 PM
Screen Shot 2022-03-26 at 12 03 38 PM


Curious some other tests were failing 🤔  but I didn't touch those and don't really want to dig into that.
Screen Shot 2022-03-26 at 12 03 23 PM

integration_tests/tests/sql/test_get_filtered_columns_in_relation.sql
joellabes: TIL! thank you 🙏

integration_tests/macros/assert_equal_values.sql
joellabes: I think that makes sense - the Core team are working on a wider testing suite that I hope will also cover packages - if that wasn't in the pipeline then I'd suggest we kept this around because "we’d need it one day" but I think we'll have a more long term solution soon.

joellabes:
OK I looked into this, and you're off the hook for the tests - they're checking `target.name` for snowflake instead of `target.type` so aren't lowercasing the properties 😬 
image

https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/test_unpivot.sql#L18-L21

And the `test_get_column_values_use_default` one I assume came from running `dbt build` or something - it has a post-hook which drops the seed after it's created which is a bit finnicky unless it's done as seed/run/test. 

These won't break on the CI jobs, (idk why that isn't running 🤔) so I think we're good to merge! Thanks for getting this done 🥳

joellabes (approved):
😍 😍

PR #524, opened 2022-03-21
feat: :sparkles: add order_by macro

pcorbel:
This is a:
- [X] new functionality — please ensure the base branch is the latest `dev/` branch

## Description & motivation
An order_by macro because all dbt_utils users who uses `group_by` should be also be allowed to use `order_by`.

## Checklist
    - [ ] BigQuery
    - [X] Postgres
    - [ ] Redshift
    - [X] Snowflake
- [X] I followed guidelines to ensure that my changes will work on "non-core" adapters by:
    - [X] dispatching any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/star.sql))
    - [X] using the `limit_zero()` macro in place of the literal string: `limit 0`
    - [X] using `dbt_utils.type_*` macros instead of explicit datatypes (e.g. `dbt_utils.type_timestamp()` instead of `TIMESTAMP`
- [X] I have added tests & descriptions to my models (and macros if applicable)
- [X] I have added an entry to CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# dbt-utils v0.8.3
+
+## New features
+- A cross-database implementation of `order_by()` ([#524](https://github.com/dbt-labs/dbt-utils/pull/524))
+
 # dbt-utils v0.8.2
 ## Fixes
 - Fix union_relations error from [#473](https://github.com/dbt-labs/dbt-utils/pull/473) when no include/exclude parameters are provided ([#505](https://github.com/dbt-labs/dbt-utils/issues/505), [#509](https://github.com/dbt-labs/dbt-utils/pull/509))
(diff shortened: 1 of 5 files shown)


joellabes:
Hi @pcorbel! Thanks for opening this PR 👋 

It's interesting that you mention the `group by` macro as the inspiration for this - if we were doing it over, we probably wouldn't have that macro in the project at all! (See [this Slack thread from back in the day](https://getdbt.slack.com/archives/CBSQTAPLG/p1621353503048500?thread_ts=1621298737.001700&cid=CBSQTAPLG)), and this article which talks about grouping by arbitrarily large numbers to be an antipattern: https://blog.getdbt.com/write-better-sql-a-defense-of-group-by-1/. I feel like creating an `order by` clause that's dozens of columns wide is also papering over a broader issue, so we probably shouldn't simplify it...

Wondering if you could share more details about your use case? I have occasionally used the `group_by` macro in a genericised context, e.g. `{{ dbt_utils.group_by(2 if something_is_true else 3) }}`, which can be useful but it seems less likely that that'd be necessary for ordering columns. 

As it currently stands, we probably wouldn't bring this macro into the project, but definitely keen to hear if there's a use case that's common to a lot of analysts that would depend on this.

PR #526, opened 2022-03-22
Add `alias` argument to `deduplicate` macro

judahrand:
This is a:

## Description & motivation

It occurred to me that quite often one might want to first filter a model and then deduplicate it. The requirement for `deduplicate` to take a relation (in order to be able to return the correct columns) makes this difficult. By adding an `alias` argument we allow the macro to find the columns from a relation but to select from a CTE.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,7 +717,7 @@ This macro returns the sql required to remove duplicate rows from a model or sou
     relation=source('my_source', 'my_table'),
     group_by="user_id, cast(timestamp as day)",
     order_by="timestamp desc",
-    alias="my_cte"
+    relation_alias="my_cte"
    )
 }}
 ```
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ For compatibility details between versions of dbt-core and dbt-utils, [see this 
 
 - [SQL generators](#sql-generators)
     - [date_spine](#date_spine-source)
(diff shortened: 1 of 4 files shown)


judahrand:
@joellabes I didn't add these changes to the original PR as it seemed like an additional enhancement to me. Would you be happy for these changes to go in too?

joellabes (changes requested):
I like it! My only request is that we call it `relation_alias`, so that its interface aligns with the star macro which has similar behaviour. Just makes it easier for people to guess the correct argument name without checking the docs, if they already know the concept from a different macro.

judahrand:
> I like it! My only request is that we call it `relation_alias`, so that its interface aligns with the star macro which has similar behaviour. Just makes it easier for people to guess the correct argument name without checking the docs, if they already know the concept from a different macro.

I had considered this initially but felt like the argument is quite different to the `relation_alias` arg of `star`, however, if you think this is better I'm happy to go with it!

joellabes:
Yeah it's not _exactly_ the same, but I think it still works! Thanks for your contribution ⭐

joellabes (approved):

PR #530, opened 2022-03-28
Feature/add listagg macro

graciegoheen:
This is a:

## Description & motivation

Create a cross database compatible `listagg` macro.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake


macros/cross_db_utils/listagg.sql
```diff
@@ -0,0 +1,50 @@
...
+        {{ delimiter_text }}
+        {% if order_by_clause -%}
+        {{ order_by_clause }}
+        {%- endif %}
+        )
+
+{%- endmacro %}
+
+{# redshift should use default instead of postgres #}
+{% macro redshift__listagg(measure, delimiter_text, order_by_clause, limit_clause) -%}
```
graciegoheen, line 46: Following the example in [datediff](https://github.com/dbt-labs/dbt-utils/blob/main/macros/cross_db_utils/datediff.sql#L62) macro - why is this necessary?

README.md
```diff
@@ -1061,6 +1062,19 @@ When an expression falls outside the range, the function returns:
...
+**Args**:
+- `measure` (required): The expression (typically a column name) that determines the values to be concatenated. To only include distinct values add keyword DISTINCT to beginning of expression (example: 'DISTINCT column_to_agg').
+- `delimiter_text` (required): Text representing the delimiter to separate concatenated values by.
+- `order_by_clause` (optional): An expression (typically a column name) that determines the order of the concatenated values.
+- `limit_clause` (optional): Only for bigquery. Specifies the maximum number of values to be concatenated.
+
+**Usage:**
+```
+{{ dbt_utils.listagg(measure='column_to_agg', delimiter_text="','", order_by_clause="order by order_by_column", limit_clause="limit 10") }}
+```
```
joellabes, line 1077: Right now, we don't (not necessarily for any good reason, in fairness) document the cross-db macros. I think it's because they're generally not intended for public consumption, so much as for usage by package maintainers themselves. 

Since they're moving to dbt core soon anyway (https://github.com/dbt-labs/dbt-core/issues/4813), I think we stick with that rule.

integration_tests/data/cross_db/data_listagg.csv
```diff
@@ -0,0 +1,5 @@
+group_col,string_text,order_col
+1,a,1
+1,b,2
+2,c,1
+2,d,2
```
joellabes, line 5: I like it when the tests can't just work by accident, e.g. if it happens to pull the data out in the default order, then we don't know whether our ordering is working or not. 

```suggestion
2,d,2
2,c,1
```

macros/cross_db_utils/listagg.sql
joellabes: Good question! Some background: 
- The Redshift adapter inherits from Postgres, so will try using Postgres-specific code before reverting to the default ([read more](https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch#adapter-inheritance))
- Postgres' datediff implementation [is very limited and needs some ridiculous hacks](https://github.com/dbt-labs/dbt-utils/blob/main/macros/cross_db_utils/datediff.sql#L27-L58). 
- Redshift actually has a good, sensible datediff implementation of its own instead of just behaving the same way as Postgres, so we're skipping the Postgres implementation in favour of the base implementation. 

This is the right thing to do here as well - Redshift uses `listagg`, not `string_agg`, so we should use the default implementation instead of Postgres'.

macros/cross_db_utils/listagg.sql
```diff
@@ -0,0 +1,50 @@
+{% macro listagg(measure, delimiter_text, order_by_clause='', limit_clause='') -%}
```
joellabes, line 1: Looks like support for DISTINCT is missing here

macros/cross_db_utils/listagg.sql
```diff
@@ -0,0 +1,50 @@
...
+  {{ return(adapter.dispatch('listagg', 'dbt_utils') (measure, delimiter_text, order_by_clause, limit_clause)) }}
```
joellabes, line 2: For optional arguments (`order_by`, `limit`), it'd be a good idea to set them to `none` so that the end user doesn't have to add that in themselves on every invocation. Otherwise I think that the macro will complain that it has four arguments and only two have been provided.

macros/cross_db_utils/listagg.sql
```diff
@@ -0,0 +1,50 @@
...
+{%- endmacro %}
+
+
+{% macro default__listagg(measure, delimiter_text, order_by_clause, limit_clause) -%}
```
joellabes, line 6: Not all of your implementations support `limit_clause` - for the cross-db macros it's important that they all do the same thing, because package maintainers use them to get consistent behaviour across all the different adapters. 

There's a couple of options, depending on how committed you're feeling (either option is OK! This is why open source is great):
- Remove the limit clause altogether - if and when someone else comes along who wants to add support, they can do it separately. 
- Look into a way to get equivalent behaviour by using other patterns. For example this sort of thing: https://stackoverflow.com/questions/27405150/postgresql-string-agg-with-limited-number-of-elements (you could stick with the native code unless a limit clause is provided, and then switch to the alternative approach).

joellabes (changes requested):
This is looking slick @graciegoheen! Thanks for getting the ball rolling so quickly. 

There's a handful of ergonomic comments (and, ironically, a request for less documentation 😬 ). 

I'd also love to see some more tests which exercise more of the permutations of arguments. For example, one that doesn't provide an order by or limit clause, one that uses distinct once it's added... you can really go down a rabbithole of permutations because they explode exponentially as you add more. 

But at a minimum I would say that each argument should be exercised at least once, along with a super basic version that passes no optional values. 

For the version without an order by clause, you _might_ have to use a different seed which has multiple copies of the same column value all the way through so that you don't get caught up in the nondeterministic sorting; I don't know. Fortunately the integration tests will let you know if that's a problem!

README.md
graciegoheen: Alright, I will remove this piece!

macros/cross_db_utils/listagg.sql
graciegoheen: I believe DISTINCT is just part of the measure input. So if you wanted to use this macro with DISTINCT, you would write: `{{ dbt_utils.listagg('DISTINCT string_text', "','", "order by order_col") }}`.

macros/cross_db_utils/listagg.sql
graciegoheen: alright! got this working (redshift was a beast) with the caveat that if there are instances of `delimiter_text` within your `measure`, you cannot include a `limit_num`

integration_tests/models/cross_db_utils/test_listagg.sql
```diff
@@ -0,0 +1,69 @@
...
+    select * from {{ ref('data_listagg_output') }}
+
+),
+
+calculate as (
+
+    select
+        group_col,
+        {{ dbt_utils.listagg('string_text', "','", "order by order_col") }} as actual,
+        1 as version
```
joellabes, line 18: I like that all of these are in the one model, but I'm wondering if we can move away from magic numbers, and instead show what the goal of the version is. Something like `comma_ordered`, `comma_ordered_limited`, `comma_unordered`, `distinct_comma`, `no_params`...

Also, just noticed that all of these use `,` as their delimiter, which is also the default value, so again it'd be good to test it with something different to make sure that it's doing what we ask it to instead of being correct by accident. 

It'd be particularly worthwhile to use a multi-character/whitespace-using delimiter for a bit of novelty, e.g. `, ` or something weird like `_|_`

joellabes (changes requested):
The Redshift implementation makes my brain hurt, but it looks right and there's tests that validate its output so I'm on board! Good work 🥳 

Last nitpicks I promise and then let's get this shipped! 🚢

integration_tests/models/cross_db_utils/test_listagg.sql
graciegoheen: Love these ideas! Thanks for all of the feedback :)

graciegoheen:
Working on making this handle special characters

integration_tests/models/cross_db_utils/test_listagg.sql
graciegoheen: Note " , " won't work as a delimiter if you have whitespace in your `measure` column. This is in-line with the note that "if there are instances of `delimiter_text` within your measure, you cannot include a `limit_num`".

joellabes (approved):

joellabes:
@graciegoheen incredible! I'm sorry this turned into such a messy contribution! 

A gold star for you ⭐ ⭐ ⭐ ⭐ ⭐

joellabes (approved):

Issue #531, opened 2022-03-30
get_column_values compiler error never gets raised - failing silently


jpmmcneill:
### Describe the bug

The `get_column_values` compiler error is not working.

https://github.com/dbt-labs/dbt-utils/blob/d2795428e163fee6e4dbc3ab9b8296e9e813c0b4/macros/sql/get_column_values.sql#L25

is blocked by

https://github.com/dbt-labs/dbt-utils/blob/d2795428e163fee6e4dbc3ab9b8296e9e813c0b4/macros/sql/get_column_values.sql#L7

### Steps to reproduce
Run the `get_column_values` against a table where the relation doesn't exist, without a default passed. Instead of raising a compiler error, [] is returned as a value.

### Expected results
A compiler error should be raised in the above case.

### Actual results
The macro fails silently.

**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
1.0.1
```
y other context about the problem here. For example, if you think you know which line of code is causing the issue.
--->

### Are you interested in contributing the fix?
I'll open a PR to fix.

jpmmcneill:
This is quite a serious bug in my opinion. It was causing production code to break in some edge cases, and the macro should  catch this case

joellabes:
Hi @jpmmcneill, thanks for calling this out! The change was made in #386 to resolve a different error. 

A potential PR to resolve this, while retaining the same behaviour, would be to remove this block https://github.com/dbt-labs/dbt-utils/blob/d2795428e163fee6e4dbc3ab9b8296e9e813c0b4/macros/sql/get_column_values.sql#L6-L8

and replace https://github.com/dbt-labs/dbt-utils/blob/d2795428e163fee6e4dbc3ab9b8296e9e813c0b4/macros/sql/get_column_values.sql#L10-L12

with `{{ return(default or []) }}` which I think would have the same result. 

We're looking to cut the next patch of dbt utils next week - if you can get a PR turned around in the next couple of days I'd be happy to include this! 

Please use `next/patch` as your base branch to avoid merge conflicts when it comes time to open your PR.

jpmmcneill:
Thanks @joellabes, PR open over here: https://github.com/dbt-labs/dbt-utils/pull/533

Do you think we should swap out `load_relation` to `get_or_create_relation`?

[...]

PR #533, opened 2022-03-31
Jmcneill/fix get column values

jpmmcneill:
This is a:
- [x] bug fix PR with no breaking changes — please ensure the base branch is `main`

## Description & motivation
Bugfix for: https://github.com/dbt-labs/dbt-utils/issues/531

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/sql/get_column_values.sql
+++ b/macros/sql/get_column_values.sql
@@ -3,11 +3,9 @@
 {% endmacro %}
 
 {% macro default__get_column_values(table, column, order_by='count(*) desc', max_records=none, default=none) -%}
-{% if default is none %}
-    {% set default = [] %}
-{% endif %}
     {#-- Prevent querying of db in parsing mode. This works because this macro does not create any new refs. #}
     {%- if not execute -%}
+        {% set default = [] if not default %}
         {{ return(default) }}
     {% endif %}
 


macros/sql/get_column_values.sql
```diff
@@ -3,11 +3,9 @@
 {% endmacro %}
 
 {% macro default__get_column_values(table, column, order_by='count(*) desc', max_records=none, default=none) -%}
-{% if default is none %}
-    {% set default = [] %}
-{% endif %}
     {#-- Prevent querying of db in parsing mode. This works because this macro does not create any new refs. #}
     {%- if not execute -%}
+        {% set default = [] if not default %}
```
jpmmcneill, line 8: return(default or []) is definitely less code, but I personally find this slightly more readable. Happy to change if lower code is important to you @joellabes 😄

joellabes (approved):
Thank you @jpmmcneill! This is a great fix 🎉 

On the issue, you asked 
>Do you think we should swap out load_relation to get_or_create_relation?

I had a look into it, and I'm almost certain that `get_or_create_relation` would mean that `relation_exists` would always be true, so we wouldn't hit any of the safety checks anymore. 

`get_or_create_relation` will create a `Relation` object if it doesn't exist ([code](https://github.com/dbt-labs/dbt-core/blob/8442fb66a591c4f5353f8baaaaf62f846c5f5171/core/dbt/include/global_project/macros/adapters/relation.sql#L59-L65)), which means that the `is not none` check will never be true. Am I interpreting that correctly? 

I'm going to merge this in as-is to fix the existing bug, and if there's something I've gotten wrong then we can look at swapping out the behaviour in a future PR. Thanks again!

jpmmcneill:
Cheers @joellabes. Happy to help!

Re. The create relation - it seemed that the problem that I ran into was caused by a cache not being updated (basically, a fresh context of DBT where a given model was not being run, but being queried as a relation).

I think that that swap would help in that use case, but it's probably a bit of an overcorrection. I think it could have been gotten around via a --ignore-cache flag in the given run

Issue #542, opened 2022-04-08
Documentation Issues with New "deduplicate" macro


codigo-ergo-sum:
### Describe the bug
The new "deduplicate" macro has a broken link at the top of the page to the documentation section on the front page of the repo. First link is https://github.com/dbt-labs/dbt-utils#deduplicate but then the second link is https://github.com/dbt-labs/dbt-utils#deduplicate-source. Could this be fixed?

Also, the macro itself is a bit confusing. I understand deduplication in general and I'm having a bit of a hard time understanding how this function is intended to be used and what some of the parameters are for. Possible to beef this up?

### Steps to reproduce
Same as above.

### Expected results
Links work on the documentation page for the repo and the macro is clearly understandable along with all parameters to be used with it, whether required or optional.

### Actual results
N/A

### Screenshots and log output
N/A

### System information
N/A

**The output of `dbt --version`:**
N/A

### Are you interested in contributing the fix?
Potentially

dbeatty10:
@judahrand

Do you have insight on how this deduplicate function is intended to be used and what each of the parameters are for?

Here's my guesses, but would be great to hear your thoughts (and suggestions for word-smithing):

**Args:**
- `relation` (required): a [Relation](https://docs.getdbt.com/reference/dbt-classes#relation) (a `ref` or `source`) whose rows you wish to deduplicate
- `group_by` (required, default=`''`): attribute names (or expressions) that compose a key for which to eliminate any duplicates (comma-separated string).
- `order_by` (optional, default=`none`): attribute names (or expressions) that determine the priority order of which row should be chosen first if there are duplicates (comma-separated string).
- `relation_alias` (optional, default=`''`): will prefix all generated fields with an alias (`relation_alias`.`field_name`).

codigo-ergo-sum:
Part of what confused me about this is the name "group_by" for the second parameter. That implies aggregation to me, but this is not doing the logical operation of aggregation (with the exception of the BQ-specific version of the macro which has to do an involved workaround due to BQ issues with window functions not being able to successfully process large volumes of data.) This is really doing a window function. It would properly be more helpful to call the second parameter something like "key_columns" or "deduplication_columns." Also, purely from reading the code, I think instead of saying `attribute names (or expressions) that compose a key for which to eliminate any duplicates (comma-separated string)` it seems a bit more intuitive to me to say `column names (or expressions) to use to identify a set/window of rows out of which to select one as the deduplicated row`.

[...]

PR #548, opened 2022-04-14
Refactor `deduplicate()` arguments

judahrand:
This is a:

## Description & motivation

This Pull Request aims to improve the `deduplicate` macro based on community feedback from its first release. It is stacked on top of #549 and introduces breaking changes.

- #542 discusses argument naming as well as how to improve the documentation.
- #542 also raises the issue that the usage of `relation_alias` to reference CTEs is confusing. This prompted me to investigate whether this is really needed - it was only needed due to the use of `dbt_utils.star` which can be avoided by using a `natural join` which is supported by most SQL syntaxes.

## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,32 @@
 # Unreleased
+
+## 🚨 deduplicate ([#542](https://github.com/dbt-labs/dbt-utils/pull/542), [#548](https://github.com/dbt-labs/dbt-utils/pull/548))
+
+The call signature of `deduplicate` has changed. The previous call signature is marked as deprecated and will be removed in the next minor version.
+
+- The `group_by` argument is now deprecated and replaced by `partition_by`.
+- The `order_by` argument is now required.
+- The `relation_alias` argument has been removed as the macro now supports `relation` as a string directly. If you were using `relation_alias` to point to a CTE previously then you can now pass the alias directly to `relation`.
+
+Before:
+```jinja
+{% macro deduplicate(relation, group_by, order_by=none, relation_alias=none) -%}
+...
+{% endmacro %}
+```
+
+After:
+```jinja
+{% macro deduplicate(relation, partition_by, order_by) -%}
+...
+{% endmacro %}
+```
+
 ## New features
-- Add an optional `where` clause parameter to `get_column_values()` to filter values returned ([#511](https://github.com/dbt-labs/dbt-utils/issues/511), [#583](https://github.com/dbt-labs/dbt-utils/pull/583)) 
+- Add an optional `where` clause parameter to `get_column_values()` to filter values returned ([#511](https://github.com/dbt-labs/dbt-utils/issues/511), [#583](https://github.com/dbt-labs/dbt-utils/pull/583))
 - Add `where` parameter to `union_relations` macro ([#554](https://github.com/dbt-labs/dbt-utils/pull/554))
+- Add Postgres specific implementation of `deduplicate()` ([#548](https://github.com/dbt-labs/dbt-utils/pull/548))
+- Add Snowflake specific implementation of `deduplicate()` ([#543](https://github.com/dbt-labs/dbt-utils/issues/543), [#548](https://github.com/dbt-labs/dbt-utils/pull/548))
 
 ## Fixes
 - Enable a negative part_number for `split_part()` ([#557](https://github.com/dbt-labs/dbt-utils/issues/557), [#559](https://github.com/dbt-labs/dbt-utils/pull/559))
@@ -10,6 +35,7 @@
 - Documentation about listagg macro ([#544](https://github.com/dbt-labs/dbt-utils/issues/544), [#560](https://github.com/dbt-labs/dbt-utils/pull/560))
 - Fix links to macro section in table of contents ([#555](https://github.com/dbt-labs/dbt-utils/pull/555))
 - Contributing guide ([#574](https://github.com/dbt-labs/dbt-utils/pull/574))
+- Add better documentation for `deduplicate()` ([#542](https://github.com/dbt-labs/dbt-utils/pull/542), [#548](https://github.com/dbt-labs/dbt-utils/pull/548))
 
 ## Under the hood
 - Fail integration tests appropriately ([#540](https://github.com/dbt-labs/dbt-utils/issues/540), [#545](https://github.com/dbt-labs/dbt-utils/pull/545))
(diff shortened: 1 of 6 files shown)


judahrand:
I'm not sure why the Redshift test is failing with
```
09:05:22  Database Error in model test_deduplicate (models/sql/test_deduplicate.sql)
09:05:22    SELECT DISTINCT ON is not supported
09:05:22    compiled SQL at target/run/dbt_utils_integration_tests/models/sql/test_deduplicate.sql
09:05:22  Encountered an error:
FailFast Error in model test_deduplicate (models/sql/test_deduplicate.sql)
```
given that `DISTINCT ON` only appears in the Postgres implementation...

judahrand:
Ah, didn't realize Redshift defaults to Postgres rather than default.

README.md
```diff
@@ -744,7 +749,21 @@ This macro returns the sql required to remove duplicate rows from a model or sou
     relation=source('my_source', 'my_table'),
     group_by="user_id, cast(timestamp as day)",
```
dbeatty10, line 750:
```suggestion
    partition_by="user_id, cast(timestamp as day)",
```

README.md
```diff
@@ -744,7 +749,21 @@ This macro returns the sql required to remove duplicate rows from a model or sou
...
+```
+with my_cte as (
+    select *
+    from {{ source('my_source', 'my_table') }}
+    where user_id = 1
+)
+
+{{ dbt_utils.deduplicate(
+    relation='my_cte',
+    group_by="user_id, cast(timestamp as day)",
```
dbeatty10, line 765:
```suggestion
    partition_by="user_id, cast(timestamp as day)",
```

README.md
```diff
@@ -744,7 +749,21 @@ This macro returns the sql required to remove duplicate rows from a model or sou
     relation=source('my_source', 'my_table'),
     group_by="user_id, cast(timestamp as day)",
     order_by="timestamp desc",
-    relation_alias="my_cte"
+   )
+}}
+```
+
```
dbeatty10, line 755: Love that there are usage examples for:
* a `source()` relation
* CTE
* `partition_by` containing multiple expressions

To round out the usage examples, it would be awesome to include:
* a `ref()` relation
* `order_by` containing multiple expressions

Shooting from the hip, the following attempts to do both in one example:
```sql
{{ dbt_utils.deduplicate(
    relation=ref('my_model'),
    partition_by="user_id",
    order_by="effective_date desc, effective_sequence desc",
   )
}}
```

dbeatty10 (commented):
@judahrand love the direction this is heading! Thank you for your detailed coverage of a wide range of databases, especially as it relates to using natural joins.

The updates are definitely more readable.

I left a few comments regarding the documentation -- appeared that the renaming to `partition_by` was overlooked in the usage examples.

I'm wondering if we can do any fancy footwork in the background so that the changes aren't breaking (but still make all the updates we want). I think I've seen some examples in past merge requests, so I'm going to try looking for those, and I'll follow-up in this conversation if/when I find them.

judahrand:
@dbeatty10 I'll deal with these comments now 😄 I also wasn't sure if you thought this would have to wait for a minor release due to breaking changes even though the macro is very very new??

README.md
```diff
@@ -735,16 +735,44 @@ This macro returns the sql required to build a date spine. The spine will includ
 ```
 
 #### deduplicate ([source](macros/sql/deduplicate.sql))
-This macro returns the sql required to remove duplicate rows from a model or source.
+This macro returns the sql required to remove duplicate rows from a model, source, or CTE.
+
+**Args:**
+ - `relation` (required): a [Relation](https://docs.getdbt.com/reference/dbt-classes#relation) (a `ref` or `source`) or string which identifies the model to deduplicate.
+ - `partition_by` (required): column names (or expressions) to use to identify a set/window of rows out of which to select one as the deduplicated row.
+ - `order_by` (required): column names (or expressions) that determine the priority order of which row should be chosen if there are duplicates (comma-separated string).
```
codigo-ergo-sum, line 743: Might I suggest an additional sentence here? "If there are multiple rows within the window specified by `partition_by` which share the same values for the columns specified in `order_by` then the result of deduplication may be nondeterministic over multiple runs of the model using the macro given that there is no way for the database to pick which row should 'win' the tie."

Here's an example of what I am talking about, it's for Snowflake, but it will be mathmatically equivalent for all other database platforms because it's not a specific database implementation thing, it's a math thing: https://community.snowflake.com/s/article/ROW-NUMBER-function-causing-Non-deterministic-behavior

codigo-ergo-sum: And I'm actually seeing this exact behavior in one of the current dbt projects that I am working on where someone used `ROW_NUMBER` to deduplicate but neglected to check if there could be "ties". So it's not just a rare or theoretical problem.

judahrand: Added a small note. Happy to call this resolved?

codigo-ergo-sum: This is great thank you so much :).

judahrand:
@dbeatty10 I'd like to try to move ahead with this? Did you manage to find any examples of the fancy footwork you mentioned?

judahrand:
@dbeatty10 I'd really like to move forward with this if possible?

dbeatty10:
@judahrand thank you for your work on this -- I can dedicate time this week to help overcome any barriers.

One big thing has changed over the last couple weeks -- we're targeting all contributions for the `main` branch now. This is a change from using the  `next/*` branches. Hopefully this makes things easier for folks down the road. But will probably make things a little trickier for you in the meantime. Just let me know if there's anything I can do to help with that piece. 

I think there's two main things we'll need to do to merge this in once all tests are passing:
1. Make sure there's no breaking changes (if possible)
2. Be able to merge to `main` instead of `next/patch`

judahrand:
> I think there's two main things we'll need to do to merge this in once all tests are passing:
> 
> 1. Make sure there's no breaking changes (if possible)
> 2. Be able to merge to `main` instead of `next/patch`

Hey @dbeatty10,

I think that I've managed to overcome both of these barriers thanks to your pointers over Slack.

Are you able to re-review this? 

Once this is merged it would also be great to try to get #550 and #551 in too.

CHANGELOG.md
```diff
@@ -1,7 +1,32 @@
 # Unreleased
+
+## 🚨 deduplicate ([#542](https://github.com/dbt-labs/dbt-utils/pull/542), [#548](https://github.com/dbt-labs/dbt-utils/pull/548))
+
+The call signature of `deduplicate` has changed. The previous call signature is marked as deprecated and will be removed in the next minor version.
+
+- The `group_by` argument is now deprecated and replaced by `partition_by`.
+- The `order_by` argument is now required now required.
```
dbeatty10, line 8:
```suggestion
- The `order_by` argument is now required.
```

CHANGELOG.md
```diff
@@ -10,6 +35,7 @@
 - Documentation about listagg macro ([#544](https://github.com/dbt-labs/dbt-utils/issues/544), [#560](https://github.com/dbt-labs/dbt-utils/pull/560))
 - Fix links to macro section in table of contents ([#555](https://github.com/dbt-labs/dbt-utils/pull/555))
 - Contributing guide ([#574](https://github.com/dbt-labs/dbt-utils/pull/574))
+- Add better documentation for `deduplicate()` ([#542](https://github.com/dbt-labs/dbt-utils/pull/542), [#548](https://github.com/dbt-labs/dbt-utils/pull/549))
```
dbeatty10, line 38:
```suggestion
- Add better documentation for `deduplicate()` ([#542](https://github.com/dbt-labs/dbt-utils/pull/542), [#548](https://github.com/dbt-labs/dbt-utils/pull/548))
```

dbeatty10 (changes requested):
This looks great @judahrand ! Thanks for your documentation of how to handle the deprecated signature and for adding an integration test for that case.

I left two comments on very minor things to update within the documentation.

Looks good to me to merge after that.

judahrand:
@dbeatty10 Sorted!

Issue #543, opened 2022-04-08
Improve Deduplicate Macro to use QUALIFY


codigo-ergo-sum:
### Describe the feature
The deduplicate macro currently uses a combination of dbt_utils.star and a subquery to work around needing to filter based on the result of a window function but not wanting to return the filtering column used. The QUALIFY keyword, recently introduced in Snowflake and BigQuery, allows for filtering the result of a query directly on a window function in a cleaner way.

This current code:

```
    select
        {{ dbt_utils.star(relation, relation_alias='deduped') | indent }}
    from (
        select
            _inner.*,
            row_number() over (
                partition by {{ group_by }}
                {% if order_by is not none -%}
                order by {{ order_by }}
                {%- endif %}
            ) as rn
        from {{ relation if relation_alias is none else relation_alias }} as _inner
    ) as deduped
    where deduped.rn = 1
```

I think could look like this:

```
       select
            *
        from {{ relation if relation_alias is none else relation_alias }} deduped
        qualify
            row_number() over (
                partition by {{ group_by }}
                {% if order_by is not none -%}
                order by {{ order_by }}
                {%- endif %}
            ) = 1
```

### Additional context
Although BQ does support qualify, it also has issues with window functions with too much data choking on single nodes, hence why the BQ override for the macro uses array_agg instead. And Redshift and potentially other databases don't support QUALIFY. But this could at least be overridden more cleanly and probably more performantly for Snowflake.

### Who will this benefit?
What kind of use case will this feature be useful for? Please be specific and provide examples, this will help us prioritize properly.

### Are you interested in contributing this feature?
Possibly, depends on time and what the dbt_utils build process looks like these days? Last time I submitted a fix a few months ago it definitely required some help from the team in troubleshooting the build process.

judahrand:
I suspect that this will still mean that a Relation has to be passed though, right? In order to be compatible with Redshift + Postgres.

judahrand:
I've implemented this (along with some other non-breaking improvements) in #549.

judahrand:
@codigo-ergo-sum This isn't entirely relevant to this issue but having taken a look at this for other DBs as well I've come to the conclusion that SQL dialects are even more annoying and limiting than I'd thought. The actual ANSI way of doing this deduplication is so rarely supported that I didn't even know it existed! As far as I can tell only Trino and PG13+ actually support it (maybe others). The ANSI way of doing this dedupe is:

```sql
select *
[...]

PR #549, opened 2022-04-14
Optimize `deduplicate()` implementations

judahrand:
This is a:

## Description & motivation

- #543 made me aware of a way to optimize the Snowflake implementation of this macro. This also made me investigate if there is a better way of doing this in Postgres which it turns out there is!

## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake

### DIFF
--- a/macros/sql/deduplicate.sql
+++ b/macros/sql/deduplicate.sql
@@ -17,18 +17,16 @@
     )
 
     select
-        data.*
+        distinct data.*
     from {{ relation }} as data
-    join row_numbered using (
-        {{ group_by }}
-        {% for expression in order_by.split(',') %}
-        {% if expression.lower().endswith(' asc') or expression.lower().endswith(' desc') %}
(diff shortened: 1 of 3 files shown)


macros/sql/deduplicate.sql
```diff
@@ -16,8 +14,58 @@
...
+-- https://docs.snowflake.com/en/sql-reference/constructs/qualify.html
+#}
+{%- macro snowflake__deduplicate(relation, group_by, order_by=none, relation_alias=none) -%}
+
+    select *
+    from {{ relation if relation_alias is none else relation_alias }}
+    qualify
+        row_number() over (
+            partition by {{ group_by }}
+            {% if order_by is not none -%}
```
codigo-ergo-sum, line 65: I understand the desire to not make this update a breaking change by requiring a parameter that was previously optional, but if `order_by` is not specified here, the query will fail on Snowflake.

PR #532, opened 2022-03-30
fix issue 506 updating star.sql file

KamranAMalik:
This is a:
- [x] bug fix PR with no breaking changes — please ensure the base branch is `main`
- [ ] new functionality — please ensure the base branch is the latest `dev/` branch
- [ ] a breaking change — please ensure the base branch is the latest `dev/` branch

## Description & motivation
This PR is fixing the issue mentioned here: https://github.com/dbt-labs/dbt-utils/issues/506

## Checklist
- [ ] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
- [ ] I followed guidelines to ensure that my changes will work on "non-core" adapters by:
    - [ ] dispatching any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/star.sql))
    - [ ] using the `limit_zero()` macro in place of the literal string: `limit 0`
    - [ ] using `dbt_utils.type_*` macros instead of explicit datatypes (e.g. `dbt_utils.type_timestamp()` instead of `TIMESTAMP`
- [ ] I have updated the README.md (if applicable)
- [ ] I have added tests & descriptions to my models (and macros if applicable)
- [ ] I have added an entry to CHANGELOG.md

PR #561, opened 2022-04-27
star macro: conditional check to see if cols contains columns, if not replace with *

### LINKED ISSUE #605: Linting `dbt_utils.star` with Sqlfluff has an empty select
### Describe the bug
Previously raised by @swanjson [here](https://getdbt.slack.com/archives/CV2KM39KR/p1651014836414859) and again by **sam I** [here](https://getdbt.slack.com/archives/CV2KM39KR/p1654191972297119):

> I have a dbt project that uses the dbt_utils.star macro that will not pass sqlfluff linting checks because the table that star is pointing at doesn’t exist yet. The current behavior of the macro is to return nothing in this situation, but it creates compiled code that looks like this:
> ```
> select
> 
> from tablename
> ```
> Code like this causes sqlfluff to raise a parsing error.

### Steps to reproduce
1. Create `some_model.sql` like:
    ```
    select
        {{ dbt_utils.star(from=ref('some_table'), except=['some_column']) }}
    from ref('some_table')
    ```
1. Make sure it doesn't exist yet:
    ```
    drop table {{ ref('some_table') }}
    ```
1. Run `dbt compile`
1. Check the compiled SQL in `target/compiled/{PROJECT_NAME}/models/some_model.sql`

### Expected results
```
select
    *
from tablename
```

### Actual results
```
select

from tablename
```

### Workaround
`dbt run` before linting (not ideal)

swanjson:
Fixes #605

This is a:
- [x] bug fix with no breaking changes — please ensure the base branch is `next/patch`
- [ ] new functionality — please ensure the base branch is `next/minor`
- [ ] a breaking change — please ensure the base branch is `next/major`

## Description & motivation
I have been using sqlfluff to lint my dbt sql code. Sqlfluff relies on the dbt compiled code to lint. When dbt_util.star is called with a reference that does not yet exist (parsing mode) without running `dbt run` it does not know how to handle `get_filtered_columns_in_relation` being empty. This PR should add the case where if `get_filtered_columns_in_relation` is empty it will replace the entire dbt_utils query with a star. This allows the compiled version to be properly linted without a parsing error. 
Previously:
```SQL
SELECT
FROM "Table.Name"
```
Now:
```SQL
SELECT *
FROM "Table.Name"
```

I'm not sure if the submitted ["fix"](https://github.com/dbt-labs/dbt-utils/pull/532/files) for this in Release 0.8.3 works as intended, but should be reviewed by someone more familiar with the codebase. 

The method I used for testing was running `dbt compile --select path/to/file` on a file like the following:
```SQL
WITH table1asCTE AS(
    SELECT {{ dbt_utils.star(from=ref('table1'),
        except=["Column1"
        ,"Column2"
    ]) }}
    FROM {{ref ('table1')}}
),

SELECT *
FROM table1asCTE
```
A prerequisite for the above is the referenced table cannot exist in the database.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/sql/star.sql
+++ b/macros/sql/star.sql
@@ -11,10 +11,14 @@
         {{ return('*') }}
     {% endif %}
 
-    {%- for col in dbt_utils.get_filtered_columns_in_relation(from, except) %}
+    {% set cols = dbt_utils.get_filtered_columns_in_relation(from, except) %}
 
-        {%- if relation_alias %}{{ relation_alias }}.{% else %}{%- endif -%}{{ adapter.quote(col)|trim }} {%- if prefix!='' or suffix!='' %} as {{ adapter.quote(prefix ~ col ~ suffix)|trim }} {%- endif -%}
-        {%- if not loop.last %},{{ '\n  ' }}{% endif %}
-
-    {%- endfor -%}
+    {%- if cols|length <= 0 -%}
+      {{- return('*') -}}
+    {%- else -%}
+        {%- for col in cols %}
+            {%- if relation_alias %}{{ relation_alias }}.{% else %}{%- endif -%}{{ adapter.quote(col)|trim }} {%- if prefix!='' or suffix!='' %} as {{ adapter.quote(prefix ~ col ~ suffix)|trim }} {%- endif -%}
+            {%- if not loop.last %},{{ '\n  ' }}{% endif %}
+        {%- endfor -%}
+    {% endif %}
 {%- endmacro %}


dbeatty10 (approved):
This is an elegant and targeted solution -- thank you @swanjson 🥇 

I was hoping to add some integration tests for this case, but in the meantime, I've manually verified the problem and this solution.

PR #566, opened 2022-05-03
Next/patch

dbeatty10:
This pull request contains content that was previously merged to `next/patch`:
- bug fixes with no breaking changes
- new functionality
- (no breaking changes)

## Description & motivation
The intent is to bring all development into the main branch and avoid long-lived change branches in the future.


dbeatty10:
The number of included commits was larger than I expected.

Closing in favor of the following PRs instead:
- #567
- #568 
- #569

PR #575, opened 2022-05-03
Experiment with functional testing

jtcohen6:
This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation
Set up functional testing within the repo, and experiment with how to factor these test cases. A lot of the changes in this PR are boilerplate pytest / CircleCI stuff; I'll leave comments to call out the cool parts.

Each test case is:
- a seed (input) with `actual` and `expected`
- a model (using macro), with `actual` (macro result) and `expected`
- a (dbt) test, to ensure that `actual = expected` All the heavy lifting is being handled within dbt, just like with current integration tests.

Should we have 1:1 mapping between macros and test cases? Or one big test case to run with a single method that defines which macros are supported? The latter is faster to run, and would require less code to define—but it's trickier to reason about, and more difficult for databases that can only support a subset of these macros.

## Checklist
    - [ ] BigQuery — need to figure out `timestamp` vs. `datetime` behavior!
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake
- [ ] ~I followed guidelines to ensure that my changes will work on "non-core" adapters by:~ We're changing it up, baby!


tests/functional/cross_db_utils/fixtures.py
```diff
@@ -0,0 +1,115 @@
+macros__test_assert_equal_sql = """
+{% test assert_equal(model, actual, expected) %}
+select * from {{ model }} where {{ actual }} != {{ expected }}
+
+{% endtest %}
+"""
+
+
+# dateadd
```
jtcohen6, line 9: This file contains the "project" file contents, all defined as Python strings. These can also live right alongside the test cases, but it can be nice to keep the concerns separated.

tests/functional/cross_db_utils/test_cross_db.py
```diff
@@ -0,0 +1,72 @@
+import os
+import pytest
+from dbt.tests.util import run_dbt
+from tests.functional.cross_db_utils.fixtures import (
```
jtcohen6, line 4: Here's where we import those file contents

tests/functional/cross_db_utils/test_cross_db.py
```diff
@@ -0,0 +1,72 @@
...
+    seeds__data_dateadd_csv,
+    models__test_dateadd_sql,
+    models__test_dateadd_yml,
+    seeds__data_datediff_csv,
+    models__test_datediff_sql,
+    models__test_datediff_yml,
+)
+
+
+class BaseCrossDbMacro:
```
jtcohen6, line 15: This class sets up the basic scaffolding for all test cases below. Child methods will inherit methods that don't change (e.g. `packages`, `macros`, `test_build_assert_equal`), and can reimplement others that do (`seeds` + `models`)

tests/functional/cross_db_utils/test_cross_db.py
```diff
@@ -0,0 +1,72 @@
...
+    def seeds(self):
+        return {}
+        
+    def models(self):
+        return {}
+
+    # actual test sequence
+    def test_build_assert_equal(self, project):
+        run_dbt(['deps'])
+        run_dbt(['build'])    # seed, model, test -- all handled by dbt
```
jtcohen6, line 36: This is actually running dbt, and making sure the thing passes. We could also:
- add `assert` statements
- inspect results / logs
- ensure that dbt **fails** if given bad input, via `expect_pass=False`

tests/functional/cross_db_utils/test_cross_db.py
```diff
@@ -0,0 +1,72 @@
...
+    @pytest.fixture(scope="class")
+    def seeds(self):
+        return {"data_dateadd.csv": seeds__data_dateadd_csv}
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "test_dateadd.yml": models__test_dateadd_yml,
+            "test_dateadd.sql": models__test_dateadd_sql,
+        }
```
jtcohen6, line 49: This is it, our three ingredients:
- seed with inputs
- model that calls the macro being tested
- test (`assert_equal`) to ensure that `actual = expected` in the model

tests/functional/cross_db_utils/test_cross_db.py
```diff
@@ -0,0 +1,72 @@
...
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "test_datediff.yml": models__test_datediff_yml,
+            "test_datediff.sql": models__test_datediff_sql,
+        }
+
+
+class TestDateDiff(BaseDateAdd):
+    pass
```
jtcohen6, line 72: By convention, pytest "discovers" tests as:
- methods whose name starts with `test_`
- defined on classes starting with `Test`
- in files starting with `test_`

Everything else is just setup, helpers, etc

run_functional_test.sh
```diff
@@ -0,0 +1,13 @@
...
+if [[ ! -f $VENV ]]; then
+    python3 -m venv venv
+    . $VENV
+
+    pip install --upgrade pip setuptools
+    pip install --pre "dbt-$1" -r dev-requirements.txt
+fi
+
+. $VENV
+python3 -m pytest tests/functional --profile $1
```
jtcohen6, line 13: Once all the setup is sorted, this is really it:
```
python3 -m pytest tests/functional --profile postgres
```

dbeatty10:
Work continued within https://github.com/dbt-labs/dbt-utils/pull/577

PR #586, opened 2022-05-12
Use built-in adapter functionality for datatypes

### LINKED ISSUE #598: Rewrite `type_*` macros to use built-in adapter capabilities
Split out from https://github.com/dbt-labs/dbt-utils/pull/577 + https://github.com/dbt-labs/dbt-utils/pull/597

### Describe the feature

In `dbt-utils`: Rewrite the type_* macros to use each adapter's `Column` object. Think about whether it makes sense to push these macro into dbt-core as well, as a sugarier interface over the underlying Python methods.

Potential follow-up work in core + plugins:
- Extend / correct `Column` types as needed (e.g. https://github.com/dbt-labs/dbt-core/issues/5238)
- [Refactoring] Reconcile / consolidate [`agate` type conversion methods](https://github.com/dbt-labs/dbt-core/blob/72c17c446440afe7d4b6501ed364effe8e14c9f2/core/dbt/adapters/sql/impl.py#L68-L92) with [`Column` class type translation](https://github.com/dbt-labs/dbt-core/blob/72c17c446440afe7d4b6501ed364effe8e14c9f2/core/dbt/adapters/base/column.py#L9-L24)

Leave current `dbt_utils.type_*` macros as they are

### Additional context
Yes, database-specific!

### Who will this benefit?

- People relying on a mixture of `dbt_utils.type_*` macros (their own model SQL / installed package code) and `Column`-object behavior built into dbt's handling for, e.g., column type expansion in incremental models. It'd be better if these both pulled from the same set of information.
- All of us, in the long run, by consolidating these types and keeping them more consistent

Yes!

jtcohen6:
resolves #598

This is a:
- [ ] documentation update
- [x] new functionality
- [x] a breaking change ??? (maybe???)

## Description & motivation

Follow-up to TODOs left from #577. Experiment with using `api.Column.translate_type` for our existing `type_*` macros.

Feels good:
- This is code defined within the adapter plugin, and already used for internal purposes.
- For the most part, it "just works" as a drop-in replacement. There are a few gaps in the core / adapter methods where we could encode slightly more of this logic.
- We can preserve dispatch on all legacy `dbt_utils` macros, so that a project / package maintainer can still intervene / override if needed.

What feels less good:
- We had some explicit behaviors here previously, that are now implicit. Do they still work as before?
- What's the right interface for package authors to access this functionality? Do they want to call `{{ api.Column.translate_type('string') }}`, `{{ get_data_type('string') }}`, or `{{ type_string() }}`?
- Within `dbt-core`: We should aim to reconcile / consolidate the [`agate` type conversion methods](https://github.com/dbt-labs/dbt-core/blob/72c17c446440afe7d4b6501ed364effe8e14c9f2/core/dbt/adapters/sql/impl.py#L68-L92) with [`Column` class type translation](https://github.com/dbt-labs/dbt-core/blob/72c17c446440afe7d4b6501ed364effe8e14c9f2/core/dbt/adapters/base/column.py#L9-L24).

## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake
    - [x] using `dbt_utils.type_*` macros instead of explicit datatypes (e.g. `dbt_utils.type_timestamp()` instead of `TIMESTAMP` — hah!


dbeatty10:
This turned out awesome! 💪 

> We had some explicit behaviors here previously, that are now implicit. Do they still work as before?

Is there anything we can do to make this feel more good? e.g., like adding in some kind of unit testing?

> What's the right interface for package authors to access this functionality? Do they want to call {{ api.Column.translate_type('string') }}, {{ get_data_type('string') }}, or {{ type_string() }}?

It seems like package authors will be able to use any of those 3 no matter what, right? But we are just try to choose which of those we want to _encourage_?

Will the package authors have a variable in hand (like `my_value = 'string'`)? If so, then either of these options make sense:
* `{{ get_data_type('string') }}`
* `{{ api.Column.translate_type('string') }}`

Otherwise, this is nice and pithy:
* `{{ type_string() }}`

jtcohen6:
> Is there anything we can do to make this feel more good? e.g., like adding in some kind of unit testing?

Definitely! I'll look into how we might do this

> But we are just try to choose which of those we want to encourage?

Correct! Which of these do we want to _encourage_, in a world where they actually all do the same thing? Just a question of the right syntactic sugar to sprinkle in.

jtcohen6:
I've added tests for these data types, using both:
- new (redefined) `type_*` macros
- previous `type_*` macro definitions, manually readded in each `Legacy` test case

Two hesitations with the current implementation:
- I'm doing a "more lenient" comparison for Postgres `type__string()`. Previously this returned `varchar`, now it returns `text`. Those are true aliases within Postgres, but (for some complex reasons) dbt doesn't perfectly map between them today. (Related: https://github.com/dbt-labs/dbt-core/issues/5238) I opted to make the test `xfail`, rather than truly pass or truly fail.
- What should `type_timestamp` do on BigQuery? The most straightforward answer is, return a `timestamp`, and that's what it does today. But on other databases, `timestamp` means (by default) `timestamp without time zone`. On BigQuery that's `datetime`, whereas `timestamp` _always_ means `timestamp with time zone`. Other macros (e.g. `dateadd`) return `datetime` values on BigQuery. I was able to get this working and these tests passing as is, and I think it's simplest for everyone if `timestamp` just means `timestamp`. But I also think we might do well to encourage folks to use more-precise types (`datetime`, `timestampntz`, `timestamptz`) that have clear and consistent analogues across databases.

All of the actual changes for this PR are happening in this repo _only_, for now. Things we could do in core/plugins to make the code here slightly cleaner:
- Move the "syntactic sugar" of `get_data_type(x)` into `dbt-core` — it's just a wrapper for `api.Column.translate_type(x)`, which is already there
- Update the `BigQueryColumn.numeric_type` function to _never_ return parametrized `numeric` types. They're technically supported on BQ, but a real nightmare to use.

macros/cross_db_utils/datatypes.sql
```diff
@@ -61,12 +86,35 @@
...
+A column of this kind will not coerce input values to any particular scale,
+whereas numeric columns with a declared scale will coerce input values to that scale.
+(The SQL standard requires a default scale of 0, i.e., coercion to integer precision.
+We find this a bit useless. If you're concerned about portability, always specify
+the precision and scale explicitly.)
+*/
+
 {% macro default__type_numeric() %}
-    numeric(28, 6)
+    {{ return(api.Column.numeric_type("numeric", 28, 6)) }}
```
jtcohen6, line 108: TODO: SparkSQL wants this to be called `decimal` instead of `numeric`. Investigate whether that works on other standard DBs, or if we should use `translate_type` for it

tests/functional/data_type/base_data_type_macro.py
```diff
@@ -0,0 +1,48 @@
...
+        # we need to be a little more lenient when mapping between 'legacy' and 'new' types that are equivalent
+        # e.g. 'character varying' and 'text'
+        elif expected_cols == actual_cols:
+            # cool, no need for jank
+            pass
+        else:
+            # this is pretty janky
+            for i in range(0, len(expected_cols)):
+                expected = project.adapter.Column(*expected_cols[i])
+                actual = project.adapter.Column(*actual_cols[i])
```
dbeatty10, line 38: An untested(!) alternative syntax based on zip:

```suggestion
            for (expected_col, actual_col) in zip(expected_cols, actual_cols):
                expected = project.adapter.Column(expected_col)
                actual = project.adapter.Column(actual_col)
```

Not sure if it's a little easier to grok or not. Regardless, the juice probably isn't worth the squeeze.

tests/functional/data_type/base_data_type_macro.py
```diff
@@ -0,0 +1,48 @@
...
+                print(f"Subtle type difference detected: {expected.data_type} vs. {actual.data_type}")
+                if any((
+                    expected.is_string() and actual.is_string(),
+                    expected.is_float() and actual.is_float(),
+                    expected.is_integer() and actual.is_integer(),
+                    expected.is_numeric() and actual.is_numeric(),
+                )):
```
dbeatty10, line 45: Spit-balling an untested(!) idea:
Define the [rich comparison method](https://portingguide.readthedocs.io/en/latest/comparisons.html#rich-comparisons) `__eq__` within [columns.py](https://github.com/dbt-labs/dbt-core/blob/a642b20abc3b729cb04fd0616db065c5d0c19868/core/dbt/adapters/base/column.py) as:

```python
    def __eq__(self, other):
        return any((
                    self.is_string() and other.is_string(),
                    self.is_float() and other.is_float(),
                    self.is_integer() and other.is_integer(),
                    self.is_numeric() and other.is_numeric(),
                ))
```

It would take some thought and care to go down this route since this would literally override the behavior of `==` for Columns. You'd probably expand the implementation above to handle all the desired cases (like comparing additional properties too). Or define a ColumnDataType class and override the `__eq__` method for that thing instead.

dbeatty10 (approved):
I feel good about this when you do.

macros/cross_db_utils/datatypes.sql
```diff
@@ -1,25 +1,37 @@
...
-    varchar
-{%- endmacro -%}
-
-{% macro postgres__type_string() %}
-    varchar
-{% endmacro %}
-
-{% macro snowflake__type_string() %}
-    varchar
+    {{ return(dbt_utils.get_data_type("string")) }}
```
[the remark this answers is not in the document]
jtcohen6: Okay, I think you make a compelling case in favor of `type_{X}` macros!

`api.Column.translate_type(dtype)` will accept ANY input string, even `api.Column.translate_type('fake_type_xyz')`. It will translate that input string if it's recognized, or just return the input string if it isn't.

So, there is benefit to the "stronger typing" achieved by macros that have the standard type right in the name, even if (behind-the-scenes) it still just shells out to `api.Column.translate_type`.

tests/functional/data_type/base_data_type_macro.py
jtcohen6: I like it! Let's see if I can get it working

tests/functional/data_type/base_data_type_macro.py
jtcohen6: It's a neat idea. I'm disinclined to change `dbt-core` functionality to accomplish this, though. The goal of this "more lenient" type comparison is to acknowledge that there are some subtle difference between the previous value yielded by the `dbt_utils.type_{X}` macro, and the new ones I'm proposing in this PR. This logic says, as long as they're close enough, this _probably_ isn't a breaking change to `dbt-utils` functionality, and we should feel comfortable making the migration.

It is a good callout for me to try refactoring this logic, however.

jtcohen6:
I removed `cross_db_utils/base_cross_db_macro.py` and `cross_db_utils/fixture_cross_db_macro.py` because I realized they were unused. Not exactly related to this PR, though, so I can kick out that change if preferred.

dbeatty10:
Looks great to me!

@jtcohen6 Do these sound like the right next steps?
1. Merge https://github.com/dbt-labs/dbt-core/pull/5428 (you)
1. Update `requirements.txt` in this PR (you or me)
1. Merge this PR (me)
1. Open PR to resolve [docs.getdbt.com #1644](https://github.com/dbt-labs/docs.getdbt.com/issues/1644) (me), review (you), merge (me)

dev-requirements.txt
```diff
@@ -1,8 +1,6 @@
...
 pytest-dotenv
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-core&subdirectory=core
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-tests-adapter&subdirectory=tests/adapter
-git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-postgres&subdirectory=plugins/postgres
-git+https://github.com/dbt-labs/dbt-redshift.git
-git+https://github.com/dbt-labs/dbt-snowflake.git
-git+https://github.com/dbt-labs/dbt-bigquery.git
+git+https://github.com/dbt-labs/dbt-core.git@jerco/data-type-macros#egg=dbt-core&subdirectory=core
+git+https://github.com/dbt-labs/dbt-core.git@jerco/data-type-macros#egg=dbt-tests-adapter&subdirectory=tests/adapter
+git+https://github.com/dbt-labs/dbt-bigquery.git@jerco/data-type-macros
```
jtcohen6, line 5: Now that the `dbt-core` PR is merged, only the `dbt-bigquery` PR (https://github.com/dbt-labs/dbt-bigquery/pull/214) is actually blocking this one, since the Redshift + Snowflake PRs don't have substantive changes (just inheriting tests).

TODO: add back imports from `main`
```
git+https://github.com/dbt-labs/dbt-redshift.git
git+https://github.com/dbt-labs/dbt-snowflake.git
```

dbeatty10: @jtcohen6 Just pushed these in a commit and CI is running now.

Also added back this one along with `dbt-redshift` and `dbt-snowflake`:
- `git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-postgres&subdirectory=plugins/postgres`

Please let me know if I shouldn't have added that one in -- obviously easy to pull it back out.

PR #597, opened 2022-05-25
Lift + shift for cross-db macros

dbeatty10:
Continues #594

This is a:
- [ ] documentation update
- [ ] new functionality
- [ ] a breaking change
- [x] behind-the-scenes refactor

## Description & motivation

See:
- https://github.com/dbt-labs/dbt-core/pull/5298
- https://github.com/dbt-labs/dbt-redshift/pull/120
- https://github.com/dbt-labs/dbt-snowflake/pull/162
- https://github.com/dbt-labs/dbt-bigquery/pull/192

## To do before merge
- [x] Reset `dev-requirements.txt` back to original
- [x] Explicitly document the definition of `dbt_utils.current_timestamp()` before and after (for each database)
    - [x] Determine if the definitions are logically equivalent or not
- [x] Determine how to handle tests decorated with `@pytest.mark`

## `current_timestamp()`
Database adapters already had a previous definition of `current_timestamp()`, and it is often expressed differently than within dbt-utils. I worry that this could lead to a breaking change, so we should determine if they are _logically_ equivalent or not.

core/dbt/include/global_project/macros/adapters/freshness.sql

We have some options for how to handle this:
1. Keep the existing definition within each database adapter and discard the dbt-utils definition
2. Overwrite the existing definition within each database adapter in favor of the dbt-utils definition
3. Keep the existing definition within each database adapter and re-institute the dbt-utils definition back into the `macros/cross_db_utils` folder.

### Comparisons

See below for a comparison between the existing adapter definition vs. the dbt-utils definition.

#### Default

Adapter:

[core/dbt/include/global_project/macros/adapters/freshness.sql](https://github.com/dbt-labs/dbt-core/blob/aeee1c23a6826aa1d254a939655cb514f38db649/core/dbt/include/global_project/macros/adapters/freshness.sql#L5-L8)
```sql
{% macro default__current_timestamp() -%}
  {{ exceptions.raise_not_implemented(
    'current_timestamp macro not implemented for adapter '+adapter.type()) }}
{%- endmacro %}
```

dbt-utils:

[cross_db_utils/current_timestamp.sql](https://github.com/dbt-labs/dbt-utils/blob/44eeb1978c1c720baca2e2cfe07671e94db02c7c/macros/cross_db_utils/current_timestamp.sql#L5-L7)
```sql
{% macro default__current_timestamp() %}
    current_timestamp::{{dbt_utils.type_timestamp()}}
{% endmacro %}
```

#### Postgres

Adapter:

[plugins/postgres/dbt/include/postgres/macros/adapters.sql
](https://github.com/dbt-labs/dbt-core/blob/aeee1c23a6826aa1d254a939655cb514f38db649/plugins/postgres/dbt/include/postgres/macros/adapters.sql#L121-L123
)
```sql
{% macro postgres__current_timestamp() -%}
  now()
{%- endmacro %}
```

dbt-utils:

[cross_db_utils/current_timestamp.sql](https://github.com/dbt-labs/dbt-utils/blob/44eeb1978c1c720baca2e2cfe07671e94db02c7c/macros/cross_db_utils/current_timestamp.sql#L5-L7)
```sql
{% macro default__current_timestamp() %}
    current_timestamp::{{dbt_utils.type_timestamp()}}
{% endmacro %}
```

#### Redshift

Adapter:

[dbt/include/redshift/macros/adapters.sql
](https://github.com/dbt-labs/dbt-redshift/blob/50399f56d912c30a55d813ec61047fb3dc9da223/dbt/include/redshift/macros/adapters.sql#L219-L221)
```sql
{% macro redshift__current_timestamp() -%}
  getdate()
{%- endmacro %}
```

dbt-utils:

[cross_db_utils/current_timestamp.sql](https://github.com/dbt-labs/dbt-utils/blob/44eeb1978c1c720baca2e2cfe07671e94db02c7c/macros/cross_db_utils/current_timestamp.sql#L9-L11)
```sql
{% macro redshift__current_timestamp() %}
    getdate()
{% endmacro %}
```

#### Snowflake

Adapter:

[dbt/include/snowflake/macros/adapters.sql
](https://github.com/dbt-labs/dbt-snowflake/blob/c086d61650a3f3e6f1733b0d9243603bf566ee90/dbt/include/snowflake/macros/adapters.sql#L147-L149)
```sql
{% macro snowflake__current_timestamp() -%}
  convert_timezone('UTC', current_timestamp())
{%- endmacro %}
```

dbt-utils:

[cross_db_utils/current_timestamp.sql](https://github.com/dbt-labs/dbt-utils/blob/44eeb1978c1c720baca2e2cfe07671e94db02c7c/macros/cross_db_utils/current_timestamp.sql#L5-L7)
```sql
{% macro default__current_timestamp() %}
    current_timestamp::{{dbt_utils.type_timestamp()}}
{% endmacro %}
```

#### BigQuery

Adapter:

[dbt/include/bigquery/macros/adapters.sql](https://github.com/dbt-labs/dbt-bigquery/blob/69d1bf211199202642e0b3965b00f70f1d77e7b2/dbt/include/bigquery/macros/adapters.sql#L103-L105)
```sql
{% macro bigquery__current_timestamp() -%}
  CURRENT_TIMESTAMP()
{%- endmacro %}
```

dbt-utils:

[cross_db_utils/current_timestamp.sql](https://github.com/dbt-labs/dbt-utils/blob/44eeb1978c1c720baca2e2cfe07671e94db02c7c/macros/cross_db_utils/current_timestamp.sql#L13-L15)
```sql
{% macro bigquery__current_timestamp() %}
    current_timestamp
{% endmacro %}
```


jtcohen6:
@dbeatty10 Amazing work thus far!

Let's split out the `current_timestamp` work into its own separate effort: https://github.com/dbt-labs/dbt-utils/issues/599. That way, the unresolved questions there don't have to block the bulk of this work, even if it means we need another follow-on PR in each plugin.

One realization I had with regard to backwards compatibility, which I don't think I was clear on in https://github.com/dbt-labs/dbt-utils/pull/594: This approach will break, I believe, anyone who is currently using a "shim" package (e.g. `spark-utils`, `tsql-utils`), and `dispatch` config that redirects macros from the `dbt_utils` namespace to prefer the "shim" package namespace. That is, if I have this in my project, [as recommended](https://github.com/dbt-labs/spark-utils#compatibility):
```yml
# dbt_project.yml
dispatch:
  - macro_namespace: dbt_utils
    search_order: ['spark_utils', 'dbt_utils']
```
Because `dbt_utils.dateadd` is now being dispatched into the `dbt` namespace, rather than the `dbt_utils` one, it will stop preferring `spark_utils.spark__dateadd`. That's ameliorable in one of three ways:
- We also add `spark__dateadd` to `dbt-spark`, which we should do (https://github.com/dbt-labs/dbt-spark/pull/359), but can't necessarily rely on all adapter maintainers to be complete with this
- We tell users to switch the `macro_namespace` in that `dispatch` config from `dbt_utils` to `dbt` —  which works, but we should really aim to avoid _any_ required changes to user project code when upgrading minor versions
- We add slightly more boilerplate code in this PR (blech), to preserve `dbt_utils` as one step along the dispatch train. So replacing this:
```sql
{% macro any_value(expression) -%}
    {{ return(adapter.dispatch('any_value', 'dbt') (expression)) }}
{% endmacro %}
```
With this:
```sql
{% macro any_value(expression) -%}
    {{ return(adapter.dispatch('any_value', 'dbt_utils') (expression)) }}
{% endmacro %}

{% macro default__any_value(expression) -%}
    {{ return(adapter.dispatch('any_value', 'dbt') (expression)) }}
{% endmacro %}
```

And maybe a note at the top of every file, "This is here for backwards compatibility only"?

dbeatty10:
Sounds great @jtcohen6 -- I will split out the `current_timestamp` work as you described.

In terms of backwards compatibility, I'm most attracted any solution that preserves full backwards compatibility without users or shim maintainers needing to do anything. So I'll take a swing at implementing the approach you described in the 3rd option for amelioration.

If we want, we can also add a deprecation warning so we have the option to fully remove this boilerplate in a future major release.

jtcohen6:
@dbeatty10 Oy, this gets deep quickly. I stuck a few debuggers in a few places, and came up with strong evidence that this is indeed a circular dependency. Some printing from [the end of this method](https://github.com/dbt-labs/dbt-core/blob/72b6a80b07b3acc9aa64ae65c5d82191dbb1aa11/core/dbt/parser/manifest.py#L514-L516):
```
dbt_redshift.redshift__dateadd: MacroDependsOn(macros=['macro.dbt_utils.default__dateadd'])
dbt_postgres.postgres__dateadd: MacroDependsOn(macros=[])
dbt.dateadd: MacroDependsOn(macros=['macro.dbt_redshift.redshift__dateadd'])
dbt.default__dateadd: MacroDependsOn(macros=[])
dbt_utils.dateadd: MacroDependsOn(macros=['macro.dbt_utils.default__dateadd'])
dbt_utils.default__dateadd: MacroDependsOn(macros=['macro.dbt_utils.dateadd', 'macro.dbt_redshift.redshift__dateadd'])
```

`dbt_redshift.redshift__dateadd` &rarr; `macro.dbt_utils.default__dateadd` &rarr; `macro.dbt_redshift.redshift__dateadd`

That first bit doesn't look quite right, but everything else does. I'd expect `dbt_redshift.redshift__dateadd` to depend on `macro.dbt.default__dateadd`, not `macro.dbt_utils.default__dateadd`! Well... I think this is because we have special macro resolution rules for "internal" macros (`dbt.*`), since they're more eagerly overridden by end users who want to reimplement internal functionality: [`get_macro`](https://github.com/dbt-labs/dbt-core/blob/72b6a80b07b3acc9aa64ae65c5d82191dbb1aa11/core/dbt/context/macro_resolver.py#L110-L120) &rarr; [this search order](https://github.com/dbt-labs/dbt-core/blob/72b6a80b07b3acc9aa64ae65c5d82191dbb1aa11/core/dbt/context/macro_resolver.py#L52-L75) is defined here (prefer `dbt_utils` > `dbt`). Ultimately, instead of returning `dbt.default__dateadd`, [this line](https://github.com/dbt-labs/dbt-core/blob/72b6a80b07b3acc9aa64ae65c5d82191dbb1aa11/core/dbt/parser/manifest.py#L535) returns `dbt_utils.default__dateadd`.

We could go _very_ deep on this logic... or we could just duplicate the `dateadd` logic in `dbt-redshift`:
```
{#-- redshift should use default instead of postgres --#}
{% macro redshift__dateadd(datepart, interval, from_date_or_timestamp) %}
    dateadd(
        {{ datepart }},
        {{ interval }},
        {{ from_date_or_timestamp }}
        )
{% endmacro %}
```

That "just works."

dbeatty10:
Thank you for the explanation and examples, @jtcohen6!

### As an aside
Do you interpret this as an issue? i.e., is it worth creating a ticket in dbt-core that describes this (undersireable?) behavior when mixing dispatch to an internal and non-internal package?

### My next actions
As you suggested, I will copy-paste all the relevant logic into the Redshift adapter. Then all the tests should pass and then we'll be safe to merge this batch of pull requests.

jtcohen6:
> Do you interpret this as an issue? i.e., is it worth creating a ticket in dbt-core that describes this (undersireable?) behavior when mixing dispatch to an internal and non-internal package?

It's a _very_ niche edge case, but it all adds up to some unexpected behavior. I think an issue for it is totally justified. That could be me or could be you, let's take it as a TODO.

jtcohen6 (approved):
🚢

dbeatty10:
### Comparison, revisited
In [this](https://github.com/dbt-labs/dbt-utils/pull/597#issuecomment-1142239494) comment, I showed the results of a functional test to comparing the competing definitions for `current_timestamp()` from the adapter vs. dbt-utils. At the time, that testing didn't find any differences when comparing values.

The method used to obtain this result always made me feel uneasy -- just because it didn't find any differences doesn't guarantee that there weren't any differences! 😅 

Since @colin-rogers-dbt is digging into an implementation for https://github.com/dbt-labs/dbt-core/issues/5521, I revisited this comparison and took a different approach.

### Approach
1. Get the rendered SQL for both adapter vs. dbt-utils for each adapter
1. Manually look up the vendor-specific documentation for relevant functions to determine the final data type
1. Translate the final data type to its semantic data type

### Semantic data types

Snowflake has 3 different timestamp data types that are (generally\*) representative of the observed behaviors across different databases: `TIMESTAMP_TZ`, `TIMESTAMP_NTZ`, and `TIMESTAMP_LTZ`.

At risk of further confusion (and introducing a [15th competing standard](https://xkcd.com/927/)), I will use the following nomenclature instead of the Snowflake-specific counterpart:
- [`TIMESTAMP_TZ`](https://docs.snowflake.com/en/sql-reference/data-types-datetime.html#timestamp-ltz-timestamp-ntz-timestamp-tz) → "offset timestamp"
- [`TIMESTAMP_NTZ`](https://docs.snowflake.com/en/sql-reference/data-types-datetime.html#timestamp-ltz-timestamp-ntz-timestamp-tz) → "civil timestamp"
- [`TIMESTAMP_LTZ`](https://docs.snowflake.com/en/sql-reference/data-types-datetime.html#timestamp-ltz-timestamp-ntz-timestamp-tz) → "session timestamp"

### Results

image

### Takeaways
- ❌  There appear to be data type mismatches between dbt-utils and dbt-core+adapters for Snowflake and Postgres
    - ✅ Meanwhile, Redshift, BigQuery, and Spark appear to have matching data types between implementations
- ⚠️ The implementations across dbt-utils use a mix of all 3 timestamp data types
    - Snowflake, Postgres, Redshift all use civil timestamps whereas BigQuery\* uses offset timestamps and Spark uses session timestamps
- ⚠️ The implementations across dbt-core+adapters also use a mix of all 3 timestamp data types
    - Redshift uses civil timestamps, Snowflake and BigQuery\* use offset timestamps, and Postgres and Spark use session timestamps

The best way to verify these results intra-database would be to materialize tables with the results and then query the datatypes.

Being able to automatically discover the semantics of different data types across databases (based on behaviors) will require a bit more ingenuity.

---

\* From the best I can tell, there are 3 different alternatives for how to think about the [`TIMESTAMP`](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp_type) data type in BigQuery:
1. "session timestamp" where UTC is the session default, but it can be changed (👈 maybe this is the correct way to think about it? [This](https://cloud.google.com/bigquery/docs/migration/redshift-sql) seems to lend credence in this direction.)
1. "offset timestamp" where the offset is _always_ UTC (what I've chosen here, somewhat arbitrarily)
1. A unique fourth timestamp data type

Would love to be shown the light by anyone that knows _for certain_.

Regardless of which of the three is most accurate, it doesn't affect the takeaways above.

Issue #331, opened 2021-02-01
Change union macro to not always include the _dbt_source_relation column


joellabes:
### Describe the feature
When developing, the _dbt_source_relation column from the [union macro](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/union.sql) is useful to check that things are behaving properly, but once it's all working properly I don't like exposing implementation details to output models. I'd like to be able to opt out of that column being added at the front of my model. 

### Describe alternatives you've considered
* Adding another intermediary model which uses the star macro, passing the column into the `except` arg
* Maintaining an internal copy of this macro

### Who will this benefit?
Me! In this case, it's a legacy hard-delete-finding snapshot using my [snapshot-sandwich approach](https://discourse.getdbt.com/t/handling-hard-deletes-from-source-tables-in-snapshots/1005/14), and we've added another column but they're out of order and don't union nicely anymore.

### Are you interested in contributing this feature?
Yes - I was planning to just wrap the `cast({{ dbt_utils.string_literal(relation) }} as {{ dbt_utils.type_string() }}) as {{ source_column_name }}` line in an `{% if source_column_name %}` block, so you can pass `None` in if you don't want the column. Open to alternative approaches though (a new bool is more explicit, but getting towards a pretty unwieldy signature)

chloe-lubin:
Hi @joellabes , that would be a great feature. I look forward to seeing its release in the future. I'd like to suggest another feature to the `union_relations` macro: is it possible to preserve the order of the columns if either of the models does not have a column instead of sending it to the end of the table? Thanks!

Ex:

### table 1
| id        | name           | email  |
| ------------- |:-------------:| -----:|

### table 2
| id        | email           |
| ------------- |:-------------:|

### unioned table
| id        | **name**           | email  |
| ------------- |:-------------:| -----:|

instead of 
| id        | email           | **name**  |
| ------------- |:-------------:| -----:|

joellabes:
@chloe-lubin thanks for suggesting this! Could I ask you to open a new issue so that it can be prioritised/discovered separately to this one?

chloe-lubin:
No problem! I opened a new issue [here](https://github.com/dbt-labs/dbt-utils/issues/538).

christineberger:
@joellabes is there any plans to fix this one?

joellabes:
@christineberger I have to say that I absolutely agree with my past self here! The joys of triaging issues you made while still on the outside 😅 

[...]

PR #624, opened 2022-07-21
Feature: Add option to remove the source_column_name on the union_relations macro

christineberger:
This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation

This covers issue #331, which allows passing in `none` in order to omit a `source_column_name` column when unioning.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake - This PR only adds a Jinja snippet and does not modify any SQL.
- [x] ~~I followed guidelines to ensure that my changes will work on "non-core" adapters by:~~ N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -886,7 +886,8 @@ final query. Note the `include` and `exclude` arguments are mutually exclusive.
 * `column_override` (optional): A dictionary of explicit column type overrides,
 e.g. `{"some_field": "varchar(100)"}`.``
 * `source_column_name` (optional, `default="_dbt_source_relation"`): The name of
-the column that records the source of this row.
+the column that records the source of this row. `source_column_name: none` can 
+be used to omit this column.
 * `where` (optional): Filter conditions to include in the `where` clause.
 
 #### generate_series ([source](macros/sql/generate_series.sql))
--- a/README.md
+++ b/README.md
@@ -886,8 +886,7 @@ final query. Note the `include` and `exclude` arguments are mutually exclusive.
 * `column_override` (optional): A dictionary of explicit column type overrides,
 e.g. `{"some_field": "varchar(100)"}`.``
 * `source_column_name` (optional, `default="_dbt_source_relation"`): The name of
-the column that records the source of this row. `source_column_name: none` can 
-be used to omit this column.
+the column that records the source of this row. Pass `None` to omit this column from the results.
 * `where` (optional): Filter conditions to include in the `where` clause.
 
 #### generate_series ([source](macros/sql/generate_series.sql))
--- a/integration_tests/tests/generic/expect_table_columns_to_match_set.sql
+++ b/integration_tests/tests/generic/expect_table_columns_to_match_set.sql
@@ -17,7 +17,7 @@
                                     | list 
     -%}
 
-    {# Replaces dbt_exoectations._list_intersect() #}
+    {# Replaces dbt_expectations._list_intersect() #}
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -886,7 +886,8 @@ final query. Note the `include` and `exclude` arguments are mutually exclusive.
 * `column_override` (optional): A dictionary of explicit column type overrides,
 e.g. `{"some_field": "varchar(100)"}`.``
 * `source_column_name` (optional, `default="_dbt_source_relation"`): The name of
-the column that records the source of this row.
+the column that records the source of this row. `source_column_name: none` can 
+be used to omit this column.
```
joellabes, line 890: A bit concerned that using a colon might make people think that's how you pass the argument. (As I type this, I'm suddenly paranoid that you can use colons as well as equals signs 😱)

```suggestion
the column that records the source of this row. Pass `None` to omit this column from the results.
```

joellabes (changes requested):
I love it 😻 thank you for doing this! For integration tests, ~you can follow the same pattern as the existing union relations tests:~ update: this is good advice for testing in general, but not in this case! see below

- make a model that calls `union_relations` but passes `None` as the column name, like [this one](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/test_union_base.sql)
- Make a seed file for the expected output, like [this one](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/data/sql/data_union_expected.csv)
- Add an equality test [like this](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/schema.yml#L151-L154)

So here's a fun thing I found while building this list! We actually are already [dropping the column in our testing](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/test_union.sql). The commit message implies that it's because the tables change depending on the environment which makes sense! 

Perhaps what you do instead is run a test that validates the columns that you expect are there, but don't worry about actually testing the union itself. You could do that either by making an empty seed file and comparing that to a [`{{ limit_zero() }}`'d](https://github.com/dbt-labs/dbt-utils/blob/73a65562bf3d9b31a1eca0b27118056ad048d8fc/integration_tests/macros/limit_zero.sql) version of the table, or by using dbt-expectations' [expect_table_columns_to_match_ordered_list](https://github.com/calogica/dbt-expectations#expect_table_columns_to_match_ordered_list)

christineberger:
@joellabes I committed your suggestion and also added integration tests! Requested re-review. 😊

integration_tests/tests/generic/expect_table_columns_to_match_set.sql
```diff
@@ -0,0 +1,54 @@
...
+    {%- set column_list = column_list | map(transform) | list -%}
+    
+    {# Replaces dbt_expectations._get_column_list() #} 
+    {%- set relation_column_names = adapter.get_columns_in_relation(model) 
+                                    | map(attribute="name") 
+                                    | map(transform) 
+                                    | list 
+    -%}
+
+    {# Replaces dbt_exoectations._list_intersect() #}
```
joellabes, line 20:
```suggestion
    {# Replaces dbt_expectations._list_intersect() #}
```

joellabes (approved):
outstanding! thank you ⭐ i'll commit the typo fix on your behalf and then get this merged!

PR #625, opened 2022-07-22
dbt Core 1.2 and utils 0.9.0 prep

joellabes:
This is a:
- [ ] documentation update
- [ ] new functionality
- [ ] a breaking change

## Description & motivation
The cross-db macros have been moved to dbt Core in v1.2. They already were dispatched to the `dbt` project, but now there are also deprecation warnings which will show during compilation. I've also bumped the min dbt version to 1.2-rc1. 

I think that with these changes, we could release utils 0.9.0-rc1 immediately, and release 0.9.0 final when 1.2 final ships.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -1,7 +1,7 @@
 name: 'dbt_utils'
 version: '0.1.0'
 
-require-dbt-version: [">=1.2.0-rc1", "<2.0.0"]
+require-dbt-version: [">=1.2.0", "<2.0.0"]
 
 config-version: 2
 
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -1,7 +1,7 @@
 name: 'dbt_utils'
 version: '0.1.0'
 
-require-dbt-version: [">=1.0.0", "<2.0.0"]
+require-dbt-version: [">=1.2.0-rc1", "<2.0.0"]
 
 config-version: 2
 
--- a/macros/cross_db_utils/xdb_deprecation_warning.sql
+++ b/macros/cross_db_utils/xdb_deprecation_warning.sql
@@ -1,4 +1,4 @@
 {% macro xdb_deprecation_warning(macro, package, model) %}
-    {%- set error_message = "Warning: the `" ~ macro ~"` macro is no longer available in dbt_utils and backwards compatibility will be removed in a future version. Use `dbt." ~ macro ~ "` instead. The " ~ package ~ "." ~ model ~ " model triggered this warning." -%}
+    {%- set error_message = "Warning: the `" ~ macro ~"` macro is now provided in dbt Core. It is no longer available in dbt_utils and backwards compatibility will be removed in a future version of the package. Use `" ~ macro ~ "` (no prefix) instead. The " ~ package ~ "." ~ model ~ " model triggered this warning." -%}
     {%- do exceptions.warn(error_message) -%}
 {% endmacro %}
--- a/macros/cross_db_utils/xdb_deprecation_warning.sql
+++ b/macros/cross_db_utils/xdb_deprecation_warning.sql
@@ -1,0 +1,4 @@
+{% macro xdb_deprecation_warning(macro, package, model) %}
+    {%- set error_message = "Warning: the `" ~ macro ~"` macro is no longer available in dbt_utils and backwards compatibility will be removed in a future version. Use `dbt." ~ macro ~ "` instead. The " ~ package ~ "." ~ model ~ " model triggered this warning." -%}
+    {%- do exceptions.warn(error_message) -%}
+{% endmacro %}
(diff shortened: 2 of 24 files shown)


macros/cross_db_utils/xdb_deprecation_warning.sql
```diff
@@ -0,0 +1,4 @@
+{% macro xdb_deprecation_warning(macro, package, model) %}
+    {%- set error_message = "Warning: the `" ~ macro ~"` macro is no longer available in dbt_utils and backwards compatibility will be removed in a future version. Use `dbt." ~ macro ~ "` instead. The " ~ package ~ "." ~ model ~ " model triggered this warning." -%}
+    {%- do exceptions.warn(error_message) -%}
+{% endmacro %}
```
dbeatty10, line 4: @jtcohen6 and I decided [here](https://github.com/dbt-labs/docs.getdbt.com/pull/1619#pullrequestreview-1021406942) to omit the `dbt` namespace for the [public-facing documentation](https://docs.getdbt.com/reference/dbt-jinja-functions/cross-database-macros) for cross-database macros.

I feel the public-facing documentation and the deprecation warning should be in tight alignment; it's confusing for end users when they differ.

There are two different _permissible_ options here (with `dbt.` prefix or without). I wonder if [this](https://github.com/dbt-labs/corp/blob/main/dbt_style_guide.md) style guide or somewhere else would be good to document our explicit recommendation for which option to use?

```suggestion
{% macro xdb_deprecation_warning(macro, package, model) %}
    {%- set error_message = "Warning: the `" ~ macro ~"` macro is no longer available in dbt_utils and backwards compatibility will be removed in a future version. Drop the `dbt_utils.` prefix and use just `" ~ macro ~ "` instead. The " ~ package ~ "." ~ model ~ " model triggered this warning." -%}
    {%- do exceptions.warn(error_message) -%}
{% endmacro %}
```

dbt_project.yml
```diff
@@ -1,7 +1,7 @@
 name: 'dbt_utils'
 version: '0.1.0'
 
-require-dbt-version: [">=1.0.0", "<2.0.0"]
+require-dbt-version: [">=1.2.0-rc1", "<2.0.0"]
```
dbeatty10, line 4: I'm guessing this will drop the `-rc1` suffix and flip to just `>=1.2.0` prior to the final release of dbt_utils 0.9.0?

dbeatty10 (changes requested):
Super exciting to see all this prepped for the final planned release prior to dbt_utils 1.0! 🚀 

The implementation turned out really clean with the reusable `xdb_deprecation_warning` macro 🤩 

One blocking item:
- @jtcohen6 and I made a [decision regarding the `dbt.` namespace prefix](https://github.com/dbt-labs/docs.getdbt.com/pull/1619#pullrequestreview-1021406942), but I'm not sure if it was a local decision only affecting [these](https://docs.getdbt.com/reference/dbt-jinja-functions/cross-database-macros) docs or if it was a more global decision in relation to style. I'd like to get that clarified so the deprecation warning reflects the recommended style one way or the other. Is there anyone responsible for dbt code style?

dbt_project.yml
joellabes: Indeed! 
- dbt-utils 0.9.0-rc1 supports dbt-core 1.2.0-rc1. 
- dbt-utils 0.9.0 will support dbt-core 1.2.0

macros/cross_db_utils/xdb_deprecation_warning.sql
joellabes: I 💯 agree with your desire for consistency of advice, especially for the exact same functions. 

_Ideally_ that should be true for all macros across the board, which would mean that we should have no `dbt.` to be in alignment with other out of the box macros (e.g it would be weird for someone to start going around saying `{{ dbt.ref('my_model') }}`. 

BUT. 

My gut feeling on why these ones are different is that they get a bit uncanny-valley-y. If I am new to dbt, then coming across `{{ dateadd() }}`, `{{ except() }}` etc could feel quite confusing - _do I need to put all SQL keywords into Jinja braces?_ (In fairness, people shouldn't come across this in their early days, because it should all be deep in the bowels of packages' code.)

If we do `{{ dbt.dateadd() }}`, then it's clear that it's behaviour provided by dbt, even though it looks a bunch like standard SQL.

This sort of matches up with something else I've been thinking of, which is the shorthand node selection syntax being the same for a package or a model or a directory (`dbt run -s something`). I think we're doing a disservice letting the default advice be "type whatever you want! We'll work it out!" versus the more explicit `package:something`, `path:something`. But that's a different issue for a different day. 

If you or jerco feel very strongly in the opposite direction, I'm willing to be talked around. I just like to know where my magic is coming from 🐰🎩

dbeatty10: Jerco and I both considered the two options and decided on the version that is shorter to read and write and aligns with `{{ ref() }}`, `{{ source() }}`, etc. i.e., without `dbt.` prefix.

Since we both had the same preference, we didn't discuss it in detail.

PR #626, opened 2022-07-26
Fix - Quote columns in unpivot

### LINKED ISSUE #216: Enable quoting column names for `unpivot` macro
I have a table whose column names are quoted, like "BX 759 - long column name".

When I try to unpivot this table, the column names fetched don't contain quotes and the macro doesn't put the quotes in them, thus the macro fails.

GtheSheep:
This is a:
- [ ] documentation update
- [ ] new functionality
- [ ] a breaking change

## Description & motivation
 Closes #216 

When unpivoting column names that require quotations (dates, strings with spaces, etc) `unpivot.sql` doesn't respect `quote_columns=true` or anything like that.

I wanted to open this to get input on best methodology before adding tests:
* Should this be a new argument?
* Should this respect the `quote_columns` config instead?
* Other?

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake

### DIFF
--- a/macros/sql/unpivot.sql
+++ b/macros/sql/unpivot.sql
@@ -10,13 +10,14 @@ Arguments:
     remove: A list of columns to remove from the resulting table. Default is none.
     field_name: Destination table column name for the source table column names.
     value_name: Destination table column name for the pivoted values
+    quote_columns: Flag for whether or not the columns being pivoted require quoting
 #}
 
-{% macro unpivot(relation=none, cast_to='varchar', exclude=none, remove=none, field_name='field_name', value_name='value', table=none) -%}
-    {{ return(adapter.dispatch('unpivot', 'dbt_utils')(relation, cast_to, exclude, remove, field_name, value_name, table)) }}
+{% macro unpivot(relation=none, cast_to='varchar', exclude=none, remove=none, field_name='field_name', value_name='value', quote_columns=false, table=none) -%}
+    {{ return(adapter.dispatch('unpivot', 'dbt_utils')(relation, cast_to, exclude, remove, field_name, value_name, quote_columns, table)) }}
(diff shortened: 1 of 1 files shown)


joellabes:
Thanks for taking this on @GtheSheep! 

> I wanted to open this to get input on best methodology before adding tests:
> 
> * Should this be a new argument?

Yes I think it should! 

> * Should this respect the `quote_columns` config instead?

I don't think so - normally for dbt utils, we aren't trying to be clever and inspect the state of yaml files. I personally consider it a bit of an antipattern to hide configuration behaviour for macros inside of yaml, because it means you have to check two places to understand how it's going to work. (If there was a global quote_columns config ever added to dbt_project.yml, I'd probably change my tune). 

Does that get you unstuck? Sorry for the delay in getting back to you 🐌

GtheSheep:
Hey @joellabes thanks for this, super helpful!

yeah I was thinking there was a global config for `quote_columns` but it seems to just be for seeds, my bad! But yep, this unblocks me, will keep it as close as possible to how this works for `pivot.sql` and get on with adding a test for this.

Thanks 😸

joellabes:
@GtheSheep Heads up that we've just done some surgery on the unpivot macro (most notably #671). Will be putting out a first beta of dbt utils 1.0 in the immediate future, which you'll probably want to pull into this branch. Would love to get these changes into 1.0 as well if possible, so that we can keep all the behavioural changes to a single release. Doesn't matter _as much_ for this one because it's backwards compatible.

seajhawk:
@GtheSheep & @joellabes - 
I'm very excited to see this PR in progress!

Is there a known workaround I can use while the PR process continues?

Issue #629, opened 2022-08-01
Relation name 'xyz__dbt_incremental_period0_tmp' is longer than 63 characters


moltar:
### Describe the bug

We have a model with a long name.

When we use `materialized: insert_by_period` and the suffix is appended by `dbt-utils`, the resulting model name is too long.

### Steps to reproduce

Use a model name that is 62 characters.

### Expected results

Not result in an error and have the ability to override the name.

### Actual results

An error is thrown:

```
20:41:47    Relation name 'xyz__dbt_incremental_period0_tmp' is longer than 63 characters
```

### Screenshots and log output

N/A

### System information
**The contents of your `packages.yml` file:**

**Which database are you using dbt with?**
- [x] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
Core:
  - installed: 1.2.0
  - latest:    1.2.0 - Up to date!

Plugins:
  - postgres: 1.2.0 - Up to date!
```

joellabes:
Hi @moltar! We're in the process of extracting the insert_by_period materialization from this project and moving it to the experiments repo, to better reflect its level of maturity. 

Once that move is complete, we'd welcome a PR to incorporate this fix!

Issue #192, opened 2020-02-14
Modernize insert_by_period materialization


clrcrl:
The [insert_by_period](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/materializations/insert_by_period_materialization.sql) materialization was written ~2 years ago, and since then, we've improved the way we write materializations. e.g.:
* Better API for relations (or maybe I just didn't use them correctly)
* No `non_destructive` flag
* Helpful functions like `make_temp_relation`

Additionally, this materialization doesn't work on warehouses other than Redshift (see #189)

It's time to give this some love: check out the refreshed code for the [incremental]https://github.com/fishtown-analytics/dbt/blob/dev/barbara-gittings/core/dbt/include/global_project/macros/materializations/incremental/incremental.sql) materialization as a good starting point.

DVAlexHiggs:
This would be really useful for us for use with our [dbtvault](https://github.com/Datavault-UK/dbtvault)  package which currently uses Snowflake. Starting to experiment! Has anything changed since this issue was created that will be useful to know?

drewbanin:
Hey @DVAlexHiggs - check out the latest docs on custom materializations over here: https://docs.getdbt.com/docs/writing-code-in-dbt/extending-dbts-programming-environment/creating-new-materializations

We've also added some helper functions that should simplify materialization code a whole bunch. Check out the default incremental materialization for some inspiration: https://github.com/fishtown-analytics/dbt/blob/dev/octavius-catto/core/dbt/include/global_project/macros/materializations/incremental/incremental.sql

DVAlexHiggs:
Thanks for the response. I'm currently experiencing some issues. We have incremental models which generate their own SQL. This SQL is generated differently when they are in incremental mode, by using the `is_incremental()` macro. I created another macro, similar to this which checks if the model is using `insert_by_period`. Currently we're running a modified `insert_by_period` materialization so that it works with snowflake, based on the current (old-style) materialization.  When running the model with the modified materialization, our incremental SQL isn't generated. This is because on the first cycle (day in this case) it is a base load.

DVAlexHiggs:
Hi. We've recently released v0.7.0 of dbtvault, which includes a modified snowflake implementation of this macro.

I believe it is a more modern version, using some of the new adapter functions and such. 

Have a look here: 
https://github.com/Datavault-UK/dbtvault/blob/v0.7.0/macros/materialisations/vault_insert_by_period_materialization.sql

[...]

PR #630, opened 2022-08-02
fix: PostgreSQL compatibility issues

moltar:
Fixes a few PostgreSQL compatibility issues.

1. Fix documented here: <https://github.com/dbt-labs/dbt-utils/issues/192#issuecomment-761029060>
2. Another issue regarding long identifiers. Documented here: <https://github.com/dbt-labs/dbt-utils/issues/629>. The solution is just a temporary hack which makes the identifier shorter.


moltar:
Did not meant to open a PR against this project. Was only intending to open one in my own fork.

Issue #447, opened 2021-11-23
Add partition argument to sequential_values test


macklin-fluehr:
### Describe the feature

The sequential values test is useful for ensuring that an entire column over the whole table is incrementing properly. In some cases, multiple types of rows can be updating in parallel within the same table. We would like the sequential_values test to include a partition feature, similar to other dbt tests. 

### Describe alternatives you've considered

We are planning to copy the source code of dbt utils and implement this ourselves.

### Additional context
Is this feature database-specific? Which database(s) is/are relevant? Please include any other relevant context here.

### Who will this benefit?

This will help cases where multiple row types are incrementing and need to be tested for incrementally. 

Example 1:
Items published on a website. Each row is a day an item is active on the website. The sequential test should test that each item has sequential data for each day it was active on the website.

Example 2:
User is using an incremental model and this test to ensure that the model is pulling data correctly daily. the incremental model pulls information on 4 different marketing channels. We'd like to ensure that the data from each marketing channel is updated incrementally.

### Are you interested in contributing this feature?
I think the solution could be as simple as adding this:
```
with windowed as (

    select
        {{ column_name }},
        lag({{ column_name }}) over (
            {% if partition_by is not none %}
            partition by {{ partition_by }}
            {% endif %}
            order by {{ column_name }}
        ) as previous_{{ column_name }}
    from {{ model }}
),
```

where partitioned by is a new parameter to the test, defaulting to None

joellabes:
Hey @macklin-fluehr! Sounds like partitioning is the hot thing at the moment. @emilyriederer's writeup covers this in a bit more detail, so even though yours came first I'm going to close it as a dupe in favour of #450 if that's OK? 

As I wrote there, I'm 100% on board with this sort of solution - my only ask is that the changed behaviour comes with tests so that we're confident that both the new and existing behaviour keep working 🛠️

macklin-fluehr:
Sounds great! Thanks

PR #633, opened 2022-08-08
Implement `group_by_columns` argument for relevant tests

### LINKED ISSUE #450: Add grouping/partitioning to relevant schema tests

emilyriederer:
This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

### Description

This PR closes #450 and #447 by implementing an optional `group_by_columns` argument across many of the core tests in `dbt-utils`. Specifically, I extended this check to allow of the relevant tests. Collectively:

- equal_rowcount()
- fewer_rows_than()
- recency()
- at_least_one()
- not_constant()
- sequential_values()
- non_null_proportion()

For example, to test for at least one valid value by group, the `group_by_columns` argument could be used as follows:

```
  - name: data_test_at_least_one
    columns:
      - name: field
        tests:
          - dbt_utils.at_least_one:
              group_by_columns: ['grouping_column']
```

### Motivation

The motivation for this PR is outlined as greater length in [this blog post](https://emilyriederer.netlify.app/post/grouping-data-quality/). In short:

- Some data checks can only be expressed within a group (e.g. ID values should be unique within a group but can be repeated between groups)
- Some data checks are more precise when done by group (e.g. not only should table rowcounts be equal but the counts within each group should be equal) 

### Implementation

In implementing this PR, I considered a few core principles:
- Make this feature as unobtrusive and isolated as possible with respect to the macros broader implementation
- Follow standard DRY principles (e.g. specifically, render needed text as few times as possible)
- Implement consistently across macros

With these principles in mind, the majority of implementations are like that of the [`recency`](https://github.com/dbt-labs/dbt-utils/blob/241090d75cb786b1f3cd636308c23560a3ef7814/macros/generic_tests/recency.sql#L7-L11) macro (L7-11)  where all relevant SQL strings are precomputed:

```
{% set threshold = dbt_utils.dateadd(datepart, interval * -1, dbt_utils.current_timestamp()) %}
{% if group_by_columns|length() > 0 %}
  {% set select_gb_cols = group_by_columns|join(' ,') + ', ' %}
  {% set groupby_gb_cols = 'group by ' + group_by_columns|join(',') %}
{% endif %}
```

The main deviations to this were the `sequential()` macro (requiring a window function) and the `equal_rowcount()`/`fewer_rows_than()` (requiring joins)

### Notes

- A "test" PR for initial design feedback is discussed in #451. Due to significant changes to `dbt-utils`, I'm sending this as a fresh PR, but the prior issue may add additional context.
- Per the checklist below, I do not believe that it is relevant to change the README.md for this PR since it only demonstrates simpler examples of macro usage. Please let me know if you would like me to do so.

## Checklist
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake


joellabes:
YAY. Thanks @emilyriederer - I will dig into this probably next week?

macros/generic_tests/equal_rowcount.sql
```diff
@@ -1,35 +1,68 @@
...
+
         count_a,
         count_b,
         abs(count_a - count_b) as diff_count
+
     from a
-    cross join b
+    full join b
+    on
+    a.id = b.id
```
joellabes, line 63: A nitpick: you could also do `full outer join b on 1 = 1` instead of creating an otherwise-unused `id` column. 

Actually as I typed this, I realised it's actually a problem not a nitpick - if either model already contains a column called `id` (not at all unlikely) then you will wind up with two columns called `id` won't you?

macros/generic_tests/fewer_rows_than.sql
```diff
@@ -1,28 +1,57 @@
...
+          a.{{c}} as {{c}}_a,
+          b.{{c}} as {{c}}_b,
+        {% endfor %}
+
         count_our_model,
         count_comparison_model
     from a
-    cross join b
+    full join b on
+    a.id = b.id
```
joellabes, line 53: Same problem here if there was a column called `id` already in the model. `full outer join b on 1 = 1` or similar would be safer

macros/generic_tests/sequential_values.sql
```diff
@@ -1,18 +1,25 @@
...
 {% endtest %}
 
-{% macro default__test_sequential_values(model, column_name, interval=1, datepart=None) %}
+{% macro default__test_sequential_values(model, column_name, interval=1, datepart=None, group_by_columns = []) %}
 
 {% set previous_column_name = "previous_" ~ dbt_utils.slugify(column_name) %}
 
+{% if group_by_columns|length() > 0 %}
+  {% set select_gb_cols = group_by_columns|join(' ,') + ', ' %}
+  {% set partition_gb_cols = 'partition by ' + group_by_columns|join(',') %}
```
joellabes, line 13: _this_ one's a real nitpick - why `' ,'` instead of `', '`? This will render as select `col_1 ,col_2` right? Obviously doesn't break anything.

joellabes (changes requested):
This is really elegant. I kept starting to write comments along the lines of "if you did XXX instead it would be a bit tidier", and then as I wrote it I realised that it was actually covering like 3 different edge cases that my proposal wouldn't have. 

I've got one nitpick around comma structure and one pattern where I think we could end up in a column name conflict, but otherwise this is magnifique 🤩

macros/generic_tests/sequential_values.sql
emilyriederer: Excellent question. Indefensible on my part! 🙂 Fixed.

macros/generic_tests/equal_rowcount.sql
emilyriederer: Thanks, @joellabes !

Overall this makes sense. I think it would only cause a problem if `id` is one of the `group_by_columns` - not any column in the table - since `a` and `b` are aggregates and not the raw tables. That said, it's still risky/likely. 

I switched it over to your proposed logic to give it a try. However, I'm not sure if joining on not-column-names is supported across all SQL dialects. It seems to have caused a failure in Redshift. (I don't have access to test locally, but I'm guessing that is the core issue?) 

A few options. Let me know which you prefer:

- Go back to initial logic but decrease chance of conflict with a "preposterous" name like `id_dbtutils_test_equal_rowcount`
- Alternatively, dynamically use a `cross join` if `group_by_columns` is length 0 or a `full join`  if we have keys
- Open to other suggestions, of course!

The first feels like a bit of a "dumb hack" but the second adds bulkier logic which potentially makes the whole macro a bit less readable

emilyriederer:
Thanks for reviewing, @joellabes ! 

I fixed the simpler comma issue and outlined a few options for the other. 

Since it feels like we are close, I'll ask one bigger picture question: Is there any better way I should document these changes? 

This felt like too small of a feature to note on the `README` that discusses macro usage. However, as it stands, this will be quite a "hidden feature" that users only learn if they read the underlying source code for the macros. (And, even there, there's no comment to define what that argument does so they'd literally have to parse it out for themselves.)

macros/generic_tests/equal_rowcount.sql
joellabes: > I think it would only cause a problem if `id` is one of the `group_by_columns` - not any column in the table - since `a` and `b` are aggregates and not the raw tables.

Yes you're right! 

The integration test gives the ambiguous error `could not devise a query plan for the given query`. Stack Overflow reports that this is when you try to use a [formula in the left join clause](https://stackoverflow.com/questions/44141879/erro-could-not-devise-a-query-plan-for-the-given-query), so I'm willing to believe that it also doesn't let you use constants for some reason.

In that case, 

> * Go back to initial logic but decrease chance of conflict with a "preposterous" name like `id_dbtutils_test_equal_rowcount`

feels like a winner to me! I agree with you that adding a lot of complex logic to change the behaviour of the macro is probably asking for trouble. 

Could you also add a comment explaining why it's necessary, as a gift to future people trying to make an optimisation?

joellabes: > I think it would only cause a problem if `id` is one of the `group_by_columns` - not any column in the table - since `a` and `b` are aggregates and not the raw tables.

Yes you're right! 

The integration test gives the ambiguous error `could not devise a query plan for the given query`. Stack Overflow reports that this is when you try to use a [formula in the left join clause](https://stackoverflow.com/questions/44141879/erro-could-not-devise-a-query-plan-for-the-given-query), so I'm willing to believe that it also doesn't let you use constants for some reason.

In that case, 

> * Go back to initial logic but decrease chance of conflict with a "preposterous" name like `id_dbtutils_test_equal_rowcount`

feels like a winner to me! I agree with you that adding a lot of complex logic to change the behaviour of the macro is probably asking for trouble. 

Could you also add a comment explaining why it's necessary, as a gift to future people trying to make an optimisation?

joellabes:
@emilyriederer thanks for doing the research! have replied on that comment above. 

> This felt like too small of a feature to note on the `README` that discusses macro usage

Nah I think this is a big deal! To save a _ton_ of repeated documentation, I would recommend documenting it all once at the list of the generic tests, and then saying something like "This test also supports the `group_by` parameter; see [group by](#group-by) for details". (that anchor obviously doesn't work in this issue)

macros/generic_tests/equal_rowcount.sql
emilyriederer: Done! Nice to see those CI checks passing again ✅

emilyriederer:
Thanks @joellabes ! I've added that section to the `README` along with other changes. Let me know if you want either more/less detail.

README.md
```diff
@@ -546,6 +559,34 @@ models:
...
+
+To use this feature, the names of grouping variables can be passed as a list. For example, to test for at least one valid value by group, the `group_by_columns` argument could be used as follows:
+
+```
+  - name: data_test_at_least_one
+    columns:
+      - name: field
+        tests:
+          - dbt_utils.at_least_one:
+              group_by_columns: ['group_var']
```
joellabes, line 587:
```suggestion
              group_by_columns: ['customer_segment']
```

I love me a real-world example; happy to have anything else along the same lines though

macros/generic_tests/fewer_rows_than.sql
```diff
@@ -13,15 +13,21 @@
       and a.{{c}} = b.{{c}}
     {% endfor %}
   {% endset %}
+  {% set groupby_gb_cols = 'group by ' + group_by_columns|join(',') %}
 {% endif %}
-{% set group_by_columns = ['id'] + group_by_columns %}
+
+{#-- We must add a fake join key in case additional grouping variables are not provided --#}
+{#-- Redshift does not allow for dynamically created join conditions (e.g. full join on 1 = 1 --#}
+{#-- The same logic is used in fewer_rows_than. In case of changes, maintain consistent logic --#}
```
joellabes, line 21: Guessing this is a copy-paste mishap?

```suggestion
{#-- The same logic is used in equal_rowcount. In case of changes, maintain consistent logic --#}
```

joellabes (changes requested):
sooooo close 😍 thanks for sticking with it! These two changes are the only things I can see holding it back; I would just commit them myself to save you a job but want to check that I've understood them properly!

README.md
emilyriederer: Love it - great point! So much more helpful to see something closer to "in the wild"

macros/generic_tests/fewer_rows_than.sql
emilyriederer: 😳 Yes indeed! Thanks for the catch

joellabes (approved):
DONE. INCREDIBLE. 

image

What a way to earn the badge! Thank you so much 🌟🌟🌟🌟🌟

emilyriederer:
Thanks for the review, @joellabes ! Excited to have this merged and to start using 🤓

joellabes:
I just reverted it as I'll put it onto the 1.0 branch, not `main` - I think it'll get very confusing if it's in the readme now but doesn't come out for a while longer! Sorry for the confusion 😩 will sort it out properly when I'm back online next week 🙇‍♂️

joellabes:
(as with all git installs, you'll be able to install that branch directly if you want to get ahead of the curve though!)

emilyriederer:
Totally makes sense - thanks for the headsup!

mgcdanny:
@emilyriederer

- Can this feature be used to test that the count of groups are monotonically increasing over time? `SELECT COUNT(*) FROM table GROUP BY (month-year, segment)` 
- Can this feature be used to test that the count of "groups" and count of rows in the groups are the same across two tables? `SELECT COUNT(*) FROM table_a GROUP BY (month-year, segment)` , `SELECT COUNT(*) FROM table_b GROUP BY (month-year, segment)` 

Appreciate it!

Thanks!

emilyriederer:
Hi @mgcdanny - this feature extends the following checks:

- equal_rowcount()
- fewer_rows_than()
- recency()
- at_least_one()
- not_constant()
- sequential_values()
- non_null_proportion()

Those all work the same as in dbt-utils, but they are now assessed separately for each group provided. 

None of these compare *between* groups, so I don't believe any could accomplish your first question. However, I believe the second can be implement with `equal_rowcount()`

PR #634, opened 2022-08-09
Add not_empty_string test

### LINKED ISSUE #632: not_blank test
### Describe the feature
dbt has the `not_null` test to monitor for null values, but in certain datasets blank values are also problematic. However, there is no `not_blank` test that I found when searching utils or dbt-expectations.
Creating a custom test that errors if there are blanks.
### Additional context
Is this feature database-specific? Which database(s) is/are relevant? Please include any other relevant context here.
I think blank entries are a possibility on all databases
### Who will this benefit?
What kind of use case will this feature be useful for? Please be specific and provide examples, this will help us prioritize properly.
Anyone needing to check data integrity to ensure there are no blank values in their dataset

Yes plz 🙏🏻

epapineau:
This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation

Resolves #632 
## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake


epapineau:
We discussed on the original issue also adding a `not_empty_array` test, but I need to put more thought into what that would look like from a testing perspective (see [core issue 2365](https://github.com/dbt-labs/dbt-core/issues/2365)). I wanted to get this out for review now though.

integration_tests/models/sql/schema.yml
```diff
@@ -182,3 +182,29 @@ models:
...
+
+  - name: test_not_empty_string_failing
+    columns:
+      - name: string
+        tests:
+          - dbt_utils.not_empty_string:
+              trim_whitespace: True
+              config:
+                severity: error
+                error_if: "<1"
```
joellabes, line 194: Should these be greater than signs?

joellabes: No, they shouldn't! This is a _very_ clever way to force the tests to check the negative and positive case. 

I really like this 🧠

integration_tests/models/sql/schema.yml
```diff
@@ -182,3 +182,29 @@ models:
...
+                warn_if: "<0"
+
+  - name: test_not_empty_string_passing
+    columns:
+      - name: string
+        tests:
+          - dbt_utils.not_empty_string:
+              trim_whitespace: True
+          - dbt_utils.not_empty_string:
+              trim_whitespace: False
```
joellabes, line 210: If you can run the same test on the same column with two different configs, is one of the configs redundant? This feels odd to me

joellabes (commented):
I like where this is going! Thank you for your patience 🙏 a couple of comments/questions where I'm not sure that I'm understanding it correctly, but otherwise let's get this merged

integration_tests/models/sql/schema.yml
epapineau: All credit goes to @dbeatty10 on this approach, but it feels good knowing that we're testing both the pass and fails.

integration_tests/models/sql/schema.yml
epapineau: I totally see your point here. I was trying to also test the `trim_whitespace` option. I've removed the redundant ones for now, but let me know if you have another suggestion for testing the option.

joellabes: I _think_ you just need to test two different columns? One with whitespace that needs trimmed to pass, and one that doesn't.

epapineau: @joellabes brilliant as always~✨ Added a `string_trim_whitespace_false` column to `test_not_empty_string_passing` that would cause the test to fail if we weren't properly trimming whitespace :)

joellabes (commented):
@epapineau this looks great! just checking if you meant to change the postgres env vars?

integration_tests/.env/postgres.env
```diff
@@ -1,5 +1,5 @@
 POSTGRES_TEST_HOST=localhost
-POSTGRES_TEST_USER=root
+POSTGRES_TEST_USER=elizepapineau
```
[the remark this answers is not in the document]
epapineau: No 🤦🏻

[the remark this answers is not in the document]
epapineau: Fixed, thanks for pointing out~

joellabes (approved):

PR #660, opened 2022-08-30
Add upper_bound_column to order by

### LINKED ISSUE #659: mutually_exclusive_ranges gives non-deterministic results in certain situations where zero-length ranges are allowed
### Describe the bug

The window functions used in the test only order by the `lower_bound_column`. If `zero_length_ranges` are allowed, it is possible for this to result in failure when it shouldn't.

### Steps to reproduce

```
RECORD    STATUS    LOWER_BOUND            UPPER_BOUND  
1         good      2021-12-01 00:00:00    2022-01-01 00:00:00
2         ok        2022-01-01 00:00:00    2022-01-01 00:00:00
3         bad       2022-01-01 00:00:00    2022-02-01 00:00:00
```

### Expected results

This data *should* pass a `mutually_exclusive_ranges` test with configuration
```
lower_bound_column: lower_bound
upper_bound_column: upper_bound
zero_length_range_allowed: true
```

and fail `mutually_exclusive_ranges` test with configuration
```
lower_bound_column: lower_bound
upper_bound_column: upper_bound
zero_length_range_allowed: false
```

### Actual results

Both of the following are recordsets ordered by `LOWER_BOUND`. The database could produce either of the following orderings of the records given above.

Ordering A:
```
RECORD    STATUS    LOWER_BOUND            UPPER_BOUND  
1         good      2021-12-01 00:00:00    2022-01-01 00:00:00
2         ok        2022-01-01 00:00:00    2022-01-01 00:00:00
3         bad       2022-01-01 00:00:00    2022-02-01 00:00:00
```

Ordering B:
```
RECORD    STATUS    LOWER_BOUND            UPPER_BOUND  
1         good      2021-12-01 00:00:00    2022-01-01 00:00:00
3         bad       2022-01-01 00:00:00    2022-02-01 00:00:00
2         ok        2022-01-01 00:00:00    2022-01-01 00:00:00
```

If the first ordering is returned, the dataset will pass the test. If the 2nd ordering is returned, the dataset will fail the test. Which ordering is returned appears to be non-deterministic, at least in Snowflake.

Also, at least in Snowflake, there is no guarantee that the same ordering will be used in the different window functions. I.e., it's possible for `next_lower_bound` to use Ordering B, but `is_last_record` to use the reverse of Ordering A.

### System information
```
packages:
  - package: dbt-labs/dbt_utils
    version: 0.8.4
```

- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x] snowflake

```
installed version: 1.0.0
   latest version: 1.0.0

Up to date!

Plugins:
  - snowflake: 1.0.0
```

Yes, I am preparing a PR to include the `upper_bound_column` as a second field in the `order by`, which would still not quite provide a deterministic ordering, but *should* provide orderings that are identical in every way that matters for the test.

sfc-gh-ancoleman:
resolves #659

This is a:
- [ ] documentation update
- [ ] new functionality
- [ ] a breaking change

## Description & motivation

The order by clause(s) used in the window functions in the `mutually_exclusive_ranges` test give non-deterministic results when more than one range within a partition has the same lower_bound. If zero-length ranges are not allowed, the test will FAIL (as it should) if this occurs. But if zero length ranges are allowed and gaps are not required, it should be expected that one could have more than one range within a partition with the same lower_bound.

This PR changes the order by clause(s) to use `{{ lower_bound_column }}, {{ upper_bound_column }}` as the ordering criteria. This will ensure that when multiple ranges have the same lower_bound, they are sorted based on their upper_bound, which will place all zero-length ranges together, with any non-zero-length range with the same lower_bound appearing as the last record in this group, causing it's upper_bound to be compared to the next distinct lower_bound.

Since the test only considers the lower and upper bound columns, the fact that records with the same lower and upper bounds won't have a guaranteed order shouldn't be problematic, since these are effectively interchangeable as far as the test is concerned. So this *should* be enough to guarantee deterministic results (i.e., the test will always PASS or will always FAIL, unless changes are made to the dataset or to the test configuration).

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake


joellabes:
@sfc-gh-ancoleman I like this! Would you be able to add a test case just like your one from the initial issue which fails under the current setup (subject to the vagaries of non-deterministic sort algorithms at least) and passes under the new system?

There are sample tests in the integration_tests project in this repo - let me know if you need a hand digging into them

sfc-gh-ancoleman:
@joellabes Since I believe that I just identified an edge-case that the existing test data doesn't include, I modified the existing `data_test_mutually_exclusive_ranges_with_gaps_zero_length.csv` seed so that it includes a zero-length record for `2020-05-08` as well as a positive-length record that also begins on `2020-05-08`. I placed the positive-length record above the zero-length record in a (possibly misguided) attempt to influence ordering.

I had initially also extended the final record for supscription_id `3` so that it overlaps with some records for subscription_id `4`, so that if at some point in the future the macro is changed in a way that breaks the functioning of the partition_by argument, the test on the dataset I modified should fail. I have undone this change, since it's not really related to this issue, and there is probably a better place to incorporate such a test.

joellabes:
Works for me! Thanks @sfc-gh-ancoleman 🌟

joellabes (approved):

PR #674, opened 2022-09-15
Fully remove varargs usage in surrogate_key and safe_add

### LINKED ISSUE #667: Fully remove varargs usage in surrogate_key and safe_add
### Describe the feature
surrogate_key has long had a deprecation warning when using the legacy syntax `{{ dbt_utils.surrogate_key('col_1', 'col_2') }}` instead of `{{ dbt_utils.surrogate_key(['col_1', 'col_2']) }}` (note the square brackets for the list): 
https://github.com/dbt-labs/dbt-utils/blob/fc8b083f07b72790234f3227df40c4f54dd60c38/macros/sql/surrogate_key.sql#L12-L18

In dbt utils 1.0, we should finally tear this out. 

Also, safe_add has the same situation. It _hasn't_ had the deprecation warning so this will be a bumpier change, but it's an easy change for users to make and having a consistent experience across the board is exactly the sort of thing we want to enable in 1.0 (I recently assumed that it already behaved this way, and it bit me). 

This looks like: 
- removing the `frustrating_jinja_feature` handling (yes, that’s what it's called!)
- removing everything that creates a list based on the varargs 
- iterating directly over the list of fields. 
- Extra for experts: consider validating the type of the input (is it a list?) and the number of arguments (is it only one?) I don't mind if we don't do this - If someone does `{{ dbt_utils.safe_add('col_1', 'col_2') }}` then it will generate `c + o + l + _ + 1`, which will fail in a reasonably obvious way. 

There are already tests for [surrogate_key](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/test_surrogate_key.sql) and [safe_add](https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/models/sql/test_safe_add.sql), they will need to be updated too. 

### Who will this benefit?
People who want a nice, tidy, consistent API in dbt utils 1.0 and beyond 

This is a great opportunity for a community member who would like to chip in!

miles170:
Closes #667.

This is a:
- [ ] documentation update
- [ ] new functionality
- [x] a breaking change

## Description & motivation

Remove the `varargs` usage in `surrogate_key`, which now only takes a single list argument.

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/sql/surrogate_key.sql
+++ b/macros/sql/surrogate_key.sql
@@ -1,47 +1,30 @@
 {%- macro surrogate_key(field_list) -%}
-    {# needed for safe_add to allow for non-keyword arguments see SO post #}
-    {# https://stackoverflow.com/questions/13944751/args-kwargs-in-jinja2-macros #}
-    {% set frustrating_jinja_feature = varargs %}
-    {{ return(adapter.dispatch('surrogate_key', 'dbt_utils')(field_list, *varargs)) }}
+    {{ return(adapter.dispatch('surrogate_key', 'dbt_utils')(field_list)) }}
 {% endmacro %}
 
 {%- macro default__surrogate_key(field_list) -%}
 
-{%- if varargs|length >= 1 or field_list is string %}
+{%- if field_list is not iterable or field_list is string or field_list is mapping -%}
 
 {%- set error_message = '
 Warning: the `surrogate_key` macro now takes a single list argument instead of \
-multiple string arguments. Support for multiple string arguments will be \
-deprecated in a future release of dbt-utils. The {}.{} model triggered this warning. \
+string arguments. The {}.{} model triggered this warning. \
 '.format(model.package_name, model.name) -%}
 
 {%- do exceptions.warn(error_message) -%}
 
-{# first argument is not included in varargs, so add first element to field_list_xf #}
-{%- set field_list_xf = [field_list] -%}
-
-{%- for field in varargs %}
-{%- set _ = field_list_xf.append(field) -%}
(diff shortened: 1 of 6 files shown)


macros/sql/surrogate_key.sql
```diff
@@ -1,40 +1,23 @@
...
-
 {%- endif -%}
 
-
 {%- set fields = [] -%}
 
-{%- for field in field_list_xf -%}
+{%- for field in field_list -%}
 
     {%- set _ = fields.append(
```
joellabes, line 22:
```suggestion
    {%- do fields.append(
```

This `set _ = ` pattern is actually a hangover - we can do do `{% do fields.append('abc') %} and it will work without creating a redundant temporary variable. 

This also applies to L44, but that's not part of the diff so I can't comment on it.

joellabes (changes requested):
This looks great @miles170! Two things: 
- a bit of cleanup of some old syntax 
- Issue #667 also talks about `safe_add` - could you make the same changes for it please? (Unless there's a separate PR for it, but I haven't seen one)

Thanks!

miles170:
> This looks great @miles170! Two things:
> 
> * a bit of cleanup of some old syntax
> * Issue #667 also talks about `safe_add` - could you make the same changes for it please? (Unless there's a separate PR for it, but I haven't seen one)
> 
> Thanks!

I'd love to

joellabes (approved):
🌟 🌟 🌟 🌟 🌟

PR #685, opened 2022-09-23
add generate_surrogate_key macro 

### LINKED ISSUE #488: Treat blank strings differently to nulls in surrogate_key()
### Describe the bug
If I had a table like so:
| field1 (STRING) | field2 |
| ----- | ----- |
| NULL | 1 |
| '' | 1 |

running `dbt_utils.surrogate_key(['field1', 'field2'])` would produce the same value for both rows

This is because of the coalesce() that happens on each value during concatenation [see here
](https://github.com/dbt-labs/dbt-utils/blob/51ed999a44fcc7f9f502be11e5f190f5bc84ba4b/macros/sql/surrogate_key.sql#L40)

DISTINCT and GROUP BY operations treat null different from '', so should we do that as well when making a unique key? Moreover, what would be the proper way to implement this? You likely cannot use another string as a universal substitute and adding a randomly-generated int or string will break idempotency for surrogate keys.

### System information
```
  - package: dbt-labs/dbt_utils
    version: 0.7.6
```

- [ ] postgres
- [ ] redshift
- [x] bigquery
- [ ] snowflake

```
0.20.2
```

dave-connors-3:
resolves #488 

This is a:
- [ ] documentation update
- [x] new functionality
- [x] a breaking change

## Description & motivation

This PR is a breaking change for users of the `surrogate_key` macro -- `dbt_utils.surrogate_key` now throws a compilation error, and the `generate_surrogate_key` macro uses a default value for the coalesce function of `_dbt_utils_surrogate_key_null_` to disambiguate from empty string values. 

Parity can be achieved by adding a boolean variable to your dbt project:

```yml
# dbt_project.yml
...
vars:
  dbt_utils:
    surrogate_key_treat_nulls_as_empty_strings: True
```

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake


dave-connors-3:
@joellabes -- assuming tests pass here I think we're in decent shape -- decided not to take the handling for non-list args into the new world of `generate_surrogate_key` -- if we're breaking everyone's project, we might as well break it for the better

macros/sql/generate_surrogate_key.sql
```diff
@@ -0,0 +1,29 @@
...
+
+{%- for field in field_list -%}
+
+    {%- set _ = fields.append(
+        "coalesce(cast(" ~ field ~ " as " ~ type_string() ~ "), '" ~ default_null_value  ~"')"
+    ) -%}
+
+    {%- if not loop.last %}
+        {%- set _ = fields.append("'-'") -%}
+    {%- endif -%}
```
joellabes, line 23: The modern way to do this is to use `do`: 
```suggestion
    {%- do fields.append(
        "coalesce(cast(" ~ field ~ " as " ~ type_string() ~ "), '" ~ default_null_value  ~"')"
    ) -%}

    {%- if not loop.last %}
        {%- do fields.append("'-'") -%}
    {%- endif -%}
```

macros/sql/surrogate_key.sql
```diff
@@ -1,51 +1,19 @@
...
 {%- set error_message = '
-Warning: the `surrogate_key` macro now takes a single list argument instead of \
-multiple string arguments. Support for multiple string arguments will be \
-deprecated in a future release of dbt-utils. The {}.{} model triggered this warning. \
+Warning: the `dbt_utils.surrogate_key` macro has been fully deprecated in favor of \
+the new `generate_surrogate_key` macro. This new macro adjusts the default string value \
+passed to the coalesce function. To achieve parity with the original macro, \
+add a variable called `surrogate_key_treat_nulls_as_empty_strings` to your project with a \
+value of True. \
+The {}.{} model triggered this warning. \
```
joellabes, line 14:
```suggestion
Warning: `dbt_utils.surrogate_key` has been replaced by \
`dbt_utils.generate_surrogate_key`. The new macro treats null values \
differently to empty strings. To restore the behaviour of the original \
macro, add a variable scoped to the dbt_utils package called \
`surrogate_key_treat_nulls_as_empty_strings` to your \
dbt_project.yml file with a value of True. \
The {}.{} model triggered this warning. \
```

Tweaked the wording. Eyeballed the alignment in the GitHub UI so it's probably not 80 chars exactly or whatever number you were going for.

joellabes (changes requested):
This is wonderful! Thank you @dave-connors-3. In the interest of getting the next beta out, I'm going to accept my suggestions and merge this, but if you want to make any changes then there will be another beta release before GA I'm sure!

joellabes (approved):
🤩 😍

PR #686, opened 2022-09-25
Limit columns selected in `expression_is_true` if failures not stored.

### LINKED ISSUE #683: expression_is_true is costly when applied to a large table
### Describe the bug

When running an `expression_is_true` test I noticed that the test required ~500gb of data (on BigQuery), which in my opinion is extremely costly for a simple test. 

Because the way the test is setup (`SELECT *` in the last statement), the total cost of the test is the same as doing a `SELECT * FROM TABLE_THAT_WE_TEST`. Which we know can be quite expensive for long and wide tables.

### Steps to reproduce

1. CREATE A LARGE TABLE
2. Run on BQ `SELECT * FROM LARGE_TABLE` and check the cost
3. Run a `expression_is_true` test against `LARGE_TABLE` and see that it is equally costly as the `SELECT *`
4. 

### Expected results
I expect a simple test to be really cheap. I do not want to take into account the cost of a simple column test when developing.

### Actual results
Its expensive.

If applicable, add screenshots or log output to help explain your problem.

### System information

- [ ] postgres
- [ ] redshift
- [X ] bigquery
- [ ] snowflake

```
<output goes here>
```

### Additional context
Add any other context about the problem here. For example, if you think you know which line of code is causing the issue.

Sure! 

`expression_is_true.sql`:
```sql

{% test expression_is_true(model, expression, column_name=None, condition='1=1') %}
{# T-SQL has no boolean data type so we use 1=1 which returns TRUE #}
{# ref https://stackoverflow.com/a/7170753/3842610 #}
  {{ return(adapter.dispatch('test_expression_is_true', 'dbt_utils')(model, expression, column_name, condition)) }}
{% endtest %}

{% macro default__test_expression_is_true(model, expression, column_name, condition) %}

with meet_condition as (
    select * from {{ model }} where {{ condition }}
)

select
    1 -- Change the * to a fixed single column, there might be some relevant info you could pass here for debugging, but just not ALL columns.
from meet_condition
{% if column_name is none %}
where not({{ expression }})
{%- else %}
where not({{ column_name }} {{ expression }})
{%- endif %}

{% endmacro %}
```

elyobo:
Fixes #683

resolves #

This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation
As discussed in https://github.com/dbt-labs/dbt-utils/issues/683, the default behaviour for the `expression_is_true` test is to `SELECT *` which is expensive in some engines. The change instead does `SELECT 1` unless `should_store_failures()` is true, resulting in faster and cheaper tests.

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake
- [ ] ~I followed guidelines to ensure that my changes will work on "non-core" adapters by:~ <b>Not applicable</b>
    - [ ] ~dispatching any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/star.sql))~
    - [ ] ~using the `limit_zero()` macro in place of the literal string: `limit 0`~
    - [ ] ~using `dbt.type_*` macros instead of explicit datatypes (e.g. `dbt.type_timestamp()` instead of `TIMESTAMP`~
- [ ] ~I have updated the README.md (if applicable)~ <b>Not applicable</b>
- [ ] I have added tests & descriptions to my models (and macros if applicable) <b>not sure if required, existing tests pass</b>


elyobo:
Tests not added, but existing tests pass; happy to add tests but not sure how to do so.

joellabes (approved):
Thanks @elyobo! Since this is changing the behaviour of the test, I've moved it to come out with the v1 release of dbt utils next month.

elyobo:
Cool, thanks @joellabes, appreciate the merge.

PR #691, opened 2022-09-28
Add optional `condition` to `not_null_proportion` test

elyobo:
This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation
The dbt core `not_null` test allows a `where` parameter and the dbt utils `expression_is_true` allows a `condition` parameter, both of which allow the test to operate on a subset of rows.

This adds a similar functoinality to the `not_null_proportion` test, sticking with the `condition` convention from `expression_is_true.

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ## New features
 - New feature to omit the `source_column_name` column on the `union_relations` macro ([#331](https://github.com/dbt-labs/dbt-utils/issues/331), [#624](https://github.com/dbt-labs/dbt-utils/pull/624))
+- New feature to allow a `condition` option to `not_null_proportion` tests ([#691](https://github.com/dbt-labs/dbt-utils/pull/691))
 
 ## Fixes
 - Better handling of whitespaces in the star macro ([#651](https://github.com/dbt-labs/dbt-utils/pull/651))
--- a/README.md
+++ b/README.md
@@ -317,6 +317,23 @@ models:
               at_least: 0.95
 ```
 
+The macro accepts an optional argument `condition` that allows for asserting the proportion is not null on a subset of all records.
+
+**Usage:**
+
+```yaml
+version: 2
+
+models:
+  - name: my_model
+    columns:
+      - name: id
+        tests:
+          - dbt_utils.not_null_proportion:
+              at_least: 0.95
+              condition: "created_at > '2018-12-31'"
+```
+
 #### not_accepted_values ([source](macros/generic_tests/not_accepted_values.sql))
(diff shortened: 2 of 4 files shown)


joellabes:
@elyobo all tests get a `where` clause for free: https://docs.getdbt.com/reference/resource-configs/where, so I don't think this PR is necessary - let me know if I've misunderstood though!

elyobo:
:exploding_head: woah, I did not know that, that's great - I'll test that out for my use case to confirm but it should be fine.

Question now would be - why does `expression_is_true` have `condition` then? Doesn't `where` make it redundant?

Incidentally it looks like that `where` substitution is another "this is might cost a lot in BigQuery" concern as per https://github.com/dbt-labs/dbt-utils/issues/683 :thinking: Another possible feature (I haven't looked for an issue yet) would be reporting on data usage in tests as it does when running models in build/run, costs like this are hidden away.

joellabes:
> Question now would be - why does `expression_is_true` have `condition` then? Doesn't `where` make it redundant?

That test predates the addition of native where support - we're removing the `not_null_where` and `unique_where` tests in dbt utils 1.0 next month for the same reason. It'd be nice to remove this `condition` argument, but migration would be messy... 🤔 i'll have a ponder.



> Incidentally it looks like that `where` substitution is another "this is might cost a lot in BigQuery" concern

How come? because it makes a CTE containing a `select *` that gets a filter applied to it, and then that CTE gets tested? (I dont remember how the tests are implemented offhand, but I think that’s it). 

I would expect any query optimiser worth its salt to notice that those columns aren't used in later parts of the query and ignore them. [Here's an example from a very good Snowflake-flavoured article](https://teej.ghost.io/understanding-the-snowflake-query-optimizer/#:~:text=geolocation%0AFROM%20cancelled_orders-,Column%20pruning,-Column%20pruning%20eliminates) talking about column pruning.

elyobo:
Yes, sorry, you're quite right about the pruning - so long as the columns are eventually not selected then BQ will prune them, so it's not making the problem any worse and can be addressed in specific tests that end up doing `select *` :+1:

`condition` removal is a BC break, but the migration for users isn't too bad (there's an exact replacement that provides the same functionality) so a major release would normally be appropriate; if it's likely to cause users pain then maybe flag it as deprecated (can you emit warnings?) and remove it in a future major release instead?

joellabes:
Yeah OK I'm sold - I've made #695 to track it.

PR #696, opened 2022-10-01
Feat: add macro get_query_results_as_single_value

### LINKED ISSUE #617: New macro: get_query_results_as_single_value
### Describe the feature
It's common to need to pull a single value from a table or query and then use it somewhere else, but it's super clunky (example: @invinceyble's [Coalesce talk last year](https://youtu.be/OcCB7BxbxDY?t=618))
image

This is exactly the sort of thing that this package should abstract away! 

Continuing to use lower-level macros (run_query, get_query_results_as_dict)

### Additional context
It should all just work! It's handled by dbt code which is already adapter aware. 

### Who will this benefit?
People who want to pull a single date/value/string and include it in a query for some reason. 

Sure am!

CR-Lough:
resolves #617

This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Checklist
-  [ x ] This code is associated with an Issue which has been triaged and [accepted for development](https://docs.getdbt.com/docs/contributing/oss-expectations#pull-requests). 
- [ x ] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [x] BigQuery
    - [ x ] Postgres
    - [x] Redshift
    - [ x ] Snowflake
    - [ x ] dispatching any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/star.sql))
    - [ x ] using the `limit_zero()` macro in place of the literal string: `limit 0`
- [ x ] I have added tests & descriptions to my models (and macros if applicable)


CR-Lough:
First pull request! I had some trouble with the dbt.type_string() macro, it kept throwing the error `Model 'model.dbt_utils_integration_tests.test_get_query_results_as_single_value' (models/sql/test_get_query_results_as_single_value.sql) depends on a node named '' which was not found`

joellabes:
This is cool! Thanks @CR-Lough, I'll try to review this week 🤞

macros/sql/get_query_results_as_single_value.sql
```diff
@@ -0,0 +1,29 @@
...
+
+        {% set sql_result = load_result('get_query_result').table.columns[nth_column].values()[nth_row]|string %}
+    
+    {%- else -%}
+    
+        {% set sql_result = "" %}
+    
+    {%- endif -%}
+
+    {{ return("CAST('"+sql_result+"' AS "+dbt.type_string()+")") }}
```
joellabes, line 27: I think we can just return the `sql_result` as a Jinja variable, and let the end user decide whether it needs cast or not. They might want to use an int as an input to another macro for example

Using `{% do return() %}` vs `{{ return() }}` is just a preference - `{%` is a statement block which feels better for commands. 
```suggestion
    {% do return(sql_result) %}
```

macros/sql/get_query_results_as_single_value.sql
```diff
@@ -0,0 +1,29 @@
...
+
+    {%- call statement('get_query_result', fetch_result=True,auto_begin=false) -%}
+
+        {{ query }}
+
+    {%- endcall -%}
+
+    {%- if execute -%}
+
+        {% set sql_result = load_result('get_query_result').table.columns[nth_column].values()[nth_row]|string %}
```
joellabes, line 19:
```suggestion
        {% set sql_result = load_result('get_query_result').table.columns[nth_column].values()[nth_row] %}
```
For the same reason as the note below, I don't think we want to cast this?

macros/sql/get_query_results_as_single_value.sql
```diff
@@ -0,0 +1,29 @@
+{% macro get_query_results_as_single_value(query, row_position=0, column_position=0) %}
```
joellabes, line 1: allowing custom row and column positions is interesting! I guess I'm not against it for column position, but it also feels a lil bit more complicated than is maybe necessary. Is there a use case you had in mind (other than making the integration tests a bit tidier)?

If we do _go_ down this route, it makes sense for column position but probably not row position. I can't imagine a situation where I want the third row's data for this sort of query - it should be returning a single row only.

macros/sql/get_query_results_as_single_value.sql
```diff
@@ -0,0 +1,29 @@
...
+    {{ return(adapter.dispatch('get_query_results_as_single_value', 'dbt_utils')(query, row_position, column_position)) }}
+{% endmacro %}
+
+{% macro default__get_query_results_as_single_value(query, row_position=0, column_position=0) %}
+
+{# This macro returns the (row_position, column_position) record in a query as a string #}
+    {%- set nth_row = row_position -%}
+    {%- set nth_column = column_position -%}
```
joellabes, line 9: These feel redundant - any specific reason you're renaming the variables two lines under where they were first created? I don't mind whether you name it nth_column or column_position, but I think it should be consistent the whole way through. 

This is of course dependent on the result of our discussion above about whether these arguments even need to exist or should always be `0`.

joellabes (changes requested):
Thanks for taking the initiative to move this forward @CR-Lough! 

No complaints with how the code you've written works, but a few concerns about the specific implementation choices. Let me know what you think - if you've got a specific vision in mind then I'm willing to join you on that journey, but if it's there "just in case" then I think we keep it simple and only make it more complex if needed in the future.

joellabes:
@CR-Lough I'm hoping to cut the release candidate of utils v1 soon and would love to include this - how are you feeling about having time to have a look at the review feedback?

CR-Lough:
Hey @joellabes , got caught up in Coalesce and the week back at work. I want to finish this up too! Will Sunday evening make the cut? Or will it be too late?

joellabes:
I am travelling from tomorrow and back on deck next Tuesday (NZ time), so that timing works perfectly 👌 thank you!

macros/sql/get_query_results_as_single_value.sql
CR-Lough: I've had a change of heart, I don't think these are necessary. They felt useful at the time, but I can't actually think of a time that I'd give this macro a query, but not specify exactly which column/row I was looking for in the query itself. Removing this!

macros/sql/get_query_results_as_single_value.sql
CR-Lough: Redundant, and no longer needed based on the decision above. Removing!

CR-Lough:
Hey @joellabes, I remember now why I forced everything to string.
When I run the query within the macro and take value at position (0, 0), the Agate table doesn't keep my date value as a date. I have forced the seed's schema, so I'm sure it's querying a date column, but retrieving the value from Agate only returns an integer.

Can you think of a way to get around this? It would seem to fail as a macro if users had to CAST their single value every time they used the macro.

joellabes:
@CR-Lough as you might have seen, I fleshed out some tests today! Pretty happy with how it's looking now, I'll get @dbeatty10 to have a second look though since I'm now too close to it

dbeatty10 (approved):
Awesome to see all these edge cases covered 💪 

👍 Approving

---

But while we're here...

If I had ruby red slippers on, I'd close my eyes, click my heels, and ask for new names for both:
- `get_query_results_as_dict`
- `get_query_results_as_single_value`

Design aspiration:
- Macro names nearly as terse as `run_query`, but still sufficiently descriptive.

I don't personally have any great ideas here, although I appreciate the brevity that the Elementary Data folks have with [these](https://github.com/elementary-data/dbt-data-reliability/tree/1225ddb/macros/utils/run_queries):
- `result_value`
- `result_row_to_dict`
- `result_column_to_list`

Maybe new names like these?
- `single_result` or `result_singleton`
- `dict_result` or `result_dict`

'cause probs won't get a ton of love for suggesting anything with [`monuple` or `monad`](https://en.wikipedia.org/wiki/Tuple#Names_for_tuples_of_specific_lengths) in the name. 😂

joellabes:
> ask for new names for both

I was writing a bigger reply on why I wasn't going to change this (mostly around adding extra thrash to an already-thrashy upgrade season), but then realised I could get on board with `get_single_value()` as a sibling to `get_column_values()`, so I'm gonna do that, and leave get_query_results_as_dict as a slight outlier.

joellabes (approved):

joellabes:
Thanks for coming on the journey @CR-Lough! great contribution 🎉

PR #698, opened 2022-10-03
Add ClickHouse support to get_tables_by_pattern_sql

mlazowik:
This is a:
- [ ] documentation update
- [ ] new functionality
- [ ] a breaking change

## Description & motivation
The generic code does not work with ClickHouse. The main motivation is to make the source generation macro from `dbt--codegen` work.

Note that ClickHouse has no concept of schemas, only databases. The schema name from dbt is used as a database name. See https://docs.getdbt.com/reference/warehouse-profiles/clickhouse-profile#description-of-clickhouse-profile-fields

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [x] ClickHouse 
    - [x] appending db-specific code to an existing macro

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,12 +17,14 @@
 - Better handling of whitespaces in the star macro ([#651](https://github.com/dbt-labs/dbt-utils/pull/651))
 - Fix to correct behavior in `mutually_exclusive_ranges` test in certain situations when `zero_length_range_allowed: true` and multiple ranges in a partition have the same value for `lower_bound_column`. ([[#659](https://github.com/dbt-labs/dbt-utils/issues/659)], [#660](https://github.com/dbt-labs/dbt-utils/pull/660))
 - Fix to utilize dbt Core version of `escape_single_quotes` instead of version from dbt Utils ([[#689](https://github.com/dbt-labs/dbt-utils/issues/689)], [#692](https://github.com/dbt-labs/dbt-utils/pull/692))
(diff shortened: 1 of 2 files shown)


joellabes:
@mlazowik thanks for opening this, and sorry for the delay in getting back to you! dbt utils only commits to working with the original dbt Labs-supported adapters - in short, we can't vend a package under our name for which we don't have the skills to resolve problems. 

When another adapter misbehaves, current best practice is to create a "shim" package which end users add to their dispatch logic (see [tsql-utils](https://github.com/dbt-msft/tsql-utils) and [spark-utils](https://github.com/dbt-labs/spark-utils) for example). 

With that said, we're trying to reduce the need for these shim packages and to instead support these sorts of things being doable using primitives in dbt Core. Sometime post-Coalesce we might need to think about whether this sort of thing should be achievable using an adapter method. 

Two options spring to mind, if we go down that road: 
- maybe we should use [list_relations_without_caching](https://docs.getdbt.com/docs/contributing/adapter-development/3-building-a-new-adapter#required-macros) and filter the results locally (while retaining the native macros for the already-supported ones). Presumably slower, but how much slower? Computers are pretty fast. And this has the benefit of happening entirely in package-space without any changes needed in Core
- maybe we should get get_tables_by_pattern added to Core, and make this another macro that adapters need to implement? Feels EXTREMELY specific. 

All of this to say that I see your PR and won't be merging it as-is, but also don't have an answer for what we do instead (yet). Please bear with me!

mlazowik:
Got it, converted to draft so that it does not look like something that needs immediate attention

PR #706, opened 2022-10-17
Update star.sql to allow for non-quote wrapped column names

### LINKED ISSUE #644: Shouldn't force quoted identifiers
### Describe the bug

`Adapter.quote` is used in many places in this package which forces quoted identifiers. My specific use case is with "star" in which case it will quote all of the columns. Depending on the target system this can cause quoted identifiers to be required to query columns generated by dbt. 

### Steps to reproduce

```
with source as (

    select * from {{ source('_lenel__sources', 'accesspane') }}

),

renamed as (

    select
        name as panel_name,
        panelid as panel_id,
        daylightsave as daylight_savings,
        paneltype as panel_type,
        panelnum as panel_number,
        paneltz as panel_timezone,
        audit_etl_job_ins_key as etl_insert_key,
        audit_etl_job_upd_key as etl_update_key,
        audit_etl_job_ins_ts as etl_inserted_at,
        audit_etl_job_upd_ts as etl_updated_at,
        table_processing_datetime as table_processed_at,
        {{ dbt_utils.star(
            from=source('_lenel__sources', 'accesspane'),
            except=["name", "panelid", "daylightsave", "paneltype", "panelnum", "paneltz", "audit_etl_job_ins_key", "audit_etl_job_upd_key", "audit_etl_job_ins_ts","audit_etl_job_upd_ts","table_processing_datetime"],
            prefix='panel_'
        )|lower }}

    from source

)

select * from renamed
```

All columns expanded by star will be quoted

### Expected results

We would expect this to respect the settings of the project.
```
baudrate as panel_baudrate,
comport as panel_comport,
terminalna as panel_terminalna,
cardholdsz as panel_cardholdsz,
cardtype as panel_cardtype,
pintype as panel_pintype,
```

### Actual results

``` 
"baudrate" as "panel_baudrate",
"comport" as "panel_comport",
"terminalna" as "panel_terminalna",
"cardholdsz" as "panel_cardholdsz",
"cardtype" as "panel_cardtype",
"pintype" as "panel_pintype",
```

### System information
```
packages:
  - package: dbt-labs/codegen
    version: 0.7.0
  - package: dbt-labs/dbt_utils
    version: 0.8.6
```

- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [X] snowflake

```
➜ dbt --version
Core:
  - installed: 1.1.0
  - latest:    1.2.0 - Update available!

Plugins:
  - bigquery:  1.1.0 - Update available!
  - snowflake: 1.1.0 - Update available!
```

Slightly outside my wheelhouse, though I'm interested in the resolution.

CR-Lough:
resolves #644

This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation
The star.sql macro inside of dbt-utils/sql forces users to encase column names in quotes. This change adds the optional 'quote_identifier' parameter (default True) to the macro.

## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake


macros/sql/star.sql
```diff
@@ -17,8 +17,14 @@
       {{- return('*') -}}
     {%- else -%}
         {%- for col in cols %}
-            {%- if relation_alias %}{{ relation_alias }}.{% else %}{%- endif -%}{{ adapter.quote(col)|trim }} {%- if prefix!='' or suffix!='' %} as {{ adapter.quote(prefix ~ col ~ suffix)|trim }} {%- endif -%}
+            {%- if relation_alias %}{{ relation_alias }}.{% else %}{%- endif -%}
+                {%- if quote_identifiers -%}
+                    {{ adapter.quote(col)|trim }} {%- if prefix!='' or suffix!='' %} as {{ adapter.quote(prefix ~ col ~ suffix)|trim }} {%- endif -%}
+                {% elif not quote_identifiers %}
```
dbeatty10, line 23: Can this just be `else`?

I didn't notice this earlier when we were looking at this together in person 😅 

```suggestion
                {% else %}
```

CR-Lough: It could! I had it as that, but a coworker of yours suggested I be even more specific.
Shall I remove it?

CR-Lough: I just finished adding all the changes that we spoke of!

dbeatty10: Yep, I'd remove it. It will be more readable without it.

CR-Lough: Got it!

integration_tests/models/sql/test_star_no_quotes.sql
```diff
@@ -0,0 +1,10 @@
+with data as (
+
+    select
+        {{ dbt_utils.star(from=ref('data_star_no_quotes'), quote_identifiers=False) | upper }}
+
+    from {{ ref('data_star_no_quotes') }}
+
+)
+
+select * from data
```
dbeatty10, line 10: Will the test fail if this is changed to `quote_identifiers=True`? I _suspect_ not! If not, then it's not really covering the test case.

I'd be happy to be wrong, so give it a shot and let me know either way.

Presuming that it's not covering the case, I'll add an idea how to handle this.

integration_tests/models/sql/test_star_quotes.sql
```diff
@@ -0,0 +1,10 @@
+with data as (
+
+    select
+        {{ dbt_utils.star(from=ref('data_star_quotes')) }}
+
+    from {{ ref('data_star_quotes') }}
+
+)
+
+select * from data
```
dbeatty10, line 10: Will the test fail if `quote_identifiers=False` is specified ? I _suspect_ not here either!

dbeatty10 (changes requested):
Nice persistence on this @CR-Lough 💪 

This one is kinda tricky to test! Let me try to explain why, and a possible path forward.

This PR is enabling new behavior to choose between quoted identifiers (default) and non-quoted identifiers. To me, it seems the best way to test this is by actually examine the string produced by `star()` and seeing if the value matches the expectations (e.g., starting and ending with the `adapter.quote` value or not depending on the arguments).

Creating such a test could be a bit different than existing tests. It might look something like this (which I didn't actually try out!):

integration_tests/data/sql/single_column_table.csv
```csv
my_column
x
```

integration_tests/models/sql/test_star_quote_identifiers.sql
```sql
select
    {{ dbt.string_literal(adapter.quote("my_column")) | lower }} as expected,
    {{ dbt_utils.star(from=ref('single_column_table', quote_identifiers=True)) | trim | lower }} as actual

union all

select
    {{ dbt.string_literal("my_column") | lower }} as expected,
    {{ dbt_utils.star(from=ref('single_column_table', quote_identifiers=False)) | trim | lower }} as actual
```

integration_tests/models/sql/schema.yml
```yaml
  - name: test_star_quote_identifiers
    tests:
      - assert_equal:
          actual: actual
          expected: expected
 ```
 
I put in all the extra `lower` filters because some databases [will return lowercase and others uppercase](https://discourse.getdbt.com/t/table-field-syntax-best-practices-when-using-snowflake/33), and we want the same logic to work across adapters. It could just as easily be `upper` filters instead.

integration_tests/models/sql/test_star_no_quotes.sql
dbeatty10: Indeed, changing to `quote_identifiers=True` did **not** induce a test failure, but we would have expected a sufficient test to fail in this scenario. Read on for details.

When I made these changes:
https://github.com/dbt-labs/dbt-utils/compare/37a2b69..bf60046

Everything still passed:
image

So you'll want to reassess the approach to testing this.

Suggestion for next steps: try the approach outlined [here](https://github.com/dbt-labs/dbt-utils/pull/706#pullrequestreview-1145006024).

CR-Lough: Agreed, I had to enclose the column name in quotes for it to return a failure with `quote_identifiers=True`. my thought for this one was that it would only return true because of the case sensitive column requirements.

Checking out your suggested solution though, I like the approach and appreciate your time!

integration_tests/models/sql/test_star_quotes.sql
CR-Lough: It still succeeds, as you suspected! Removing this test altogether (will test inside of your suggested path).

integration_tests/models/sql/test_star_no_quotes.sql
CR-Lough: With a slight edit to your suggested approach, the new test is passing.
And, if I flip the `quote_identifiers` boolean in each of quotes/no quotes, the tests both fail this time!

dbeatty10 (commented):
This looks awesome @CR-Lough -- nice work figuring out those tests 💪 

I took another pass through this PR and saw that there's no updates reflecting the new argument within the README.

Could you add your new argument here?
https://github.com/dbt-labs/dbt-utils/blob/6a5dd1eb97700301eaeba14c23c3682ceac9fb2b/README.md?plain=1#L936-L942

And an example usage here?
https://github.com/dbt-labs/dbt-utils/blob/6a5dd1eb97700301eaeba14c23c3682ceac9fb2b/README.md?plain=1#L944-L965

CR-Lough:
@dbeatty10 that last addition to the README.md should be the last checkbox in the list of items at the top of the page 👍

dbeatty10 (approved):
@CR-Lough Thanks for this contribution to add the optional configuration to `star()` for quoting identifiers! 🚀

PR #707, opened 2022-10-17
Slugify for snowflake

fivetran-catfritz:
resolves #

This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation

Snowflake does not allow columns to begin with a number, so to preserve the number, the changes adds an underscore to the beginning of the string if the target is Snowflake. 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake

Cc: @crowemi thanks so much for collaborating with us on this. I would love to have your feedback!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 # Unreleased
 
 ## New features
+- Added Snowflake rules to the `slugify` macro to preserve column names beginning with a number.
 - Implemented an optional `group_by_columns` argument across many of the generic testing macros to test for properties that only pertain to group-level or are can be more rigorously conducted at the group level. Property available in `recency`, `at_least_one`, `equal_row_count`, `fewer_rows_than`, `not_constant`, `not_null_proportion`, and `sequential` tests [#633](https://github.com/dbt-labs/dbt-utils/pull/633)
 - New feature to omit the `source_column_name` column on the `union_relations` macro ([#331](https://github.com/dbt-labs/dbt-utils/issues/331), [#624](https://github.com/dbt-labs/dbt-utils/pull/624))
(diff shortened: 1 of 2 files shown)


joellabes (changes requested):
Hey @fivetran-catfritz! Happy Hackathon and thanks for opening this. 

It's not just Snowflake that has this problem, right? From some quick Googling, Postgres and BigQuery's columns must start with a letter or underscore, and Redshift can't start with a number or `$` unless quoted (so basically the same rule). 

I think this change can be applied across the board! Which also makes things cleaner for cross-db purposes - I don't _love_ the idea of different column names based on the warehouse this runs on.

fivetran-catfritz:
Hi @joellabes--thanks for your feedback on this. I agree you're totally right about the same naming requirements for the other dbs! I have made the updates and changed the target branch. Just let me know if you need any further changes!

joellabes:
@fivetran-catfritz thanks for this! Sorry for the delay getting back to you. Looks good - only thing left would be to add a test that makes sure it works correctly. 

The easiest thing to do would be to copy https://github.com/dbt-labs/dbt-utils/blob/main/integration_tests/tests/jinja_helpers/test_slugify.sql but make a new version that starts with a number

The slightly nicer option would be along the lines of 

```
with comparisons as (
  select '{{ dbt_utils.slugify("!Hell0 world-hi") }}' as output, 'hell0_world_hi' as expected
  union 
  select '{{ dbt_utils.slugify("0Hell0 world-hi") }}' as output, '_0hell0_world_hi' as expected
)
select * 
from comparisons
where output != expected
```

but I haven't tested that code so don't know if it'll work perfectly!

fivetran-catfritz:
@joellabes Thanks again! The test you suggested worked great--just had to change to a "union all" for bigquery. It should be good to go now! 😃

joellabes (approved):
🎉

PR #708, opened 2022-10-17
Fix recency test to work properly with `day` datepart

### LINKED ISSUE #414: Recency does not work as expected for datepart day
### Describe the bug

Given a current date of `'2021-09-10'` running this test with `datepart: day` and `interval: 3` it should succeed if I have data in my table with a `datestamp >= '2021-09-07'`.
However, this test fail because the `threshold` is defined as type `datetime`.

To fix this, I could configure the interval to be `4` and the threshold will be a day before. The test will work anyway but it is be semantically wrong and easy to miss interpret.

### Steps to reproduce
Configure the test as follow, `datestamp` is a column of type `date`.
```yml
      - dbt_utils.recency:
          datepart: day
          field: datestamp
          interval: 3
```

### Expected results
Assuming that the current date `'2021-09-10'` this tests should succeed if I have data with `datestamp >= '2021-09-07'`. 

### Actual results
The test fail because because it use as threshold a type `datetime` and compared with the type `date` of the table it returns a failure.

The generated test SQL:
```sql
select
      count(*) as failures,
      count(*) != 0 as should_warn,
      count(*) != 0 as should_error
    from (
      



with recency as (

    select max(datestamp) as most_recent
    from `project_id`.`dataset_id`.`table_id`

)

select

    most_recent,
    

        datetime_add(
            cast( 
    current_timestamp
 as datetime),
        interval -3 day
        )

 as threshold

from recency
where most_recent < 

        datetime_add(
            cast( 
    current_timestamp
 as datetime),
        interval -3 day
        )




      
    ) dbt_internal_test
```

If I run just the code for `dbt_internal_test` the result I get is the following:
```
most_recent | threshold
2021-09-07 | 2021-09-07T06:40:19.603626
```

This is because `most_recent` is type `date` and `threshold` is type `datetime`. If I cast the `threshold` to `date`, the test succeed as expected:
```sql
date(
        datetime_add(
            cast( 
    current_timestamp
 as datetime),
        interval -3 day
        )
)
```

The generated SQL should cast/truncate the `threshold` to the same "type" or "level" of the column we are comparing too.

### System information
```yml
packages:
  - package: dbt-labs/dbt_utils
    version: 0.7.1
```

- [ ] postgres
- [ ] redshift
- [X] bigquery
- [ ] snowflake

```bash
installed version: 0.20.2
   latest version: 0.20.2

Up to date!

Plugins:
  - bigquery: 0.20.2
```

yuanminglee:
resolves #414

This is a:
- [ ] documentation update
- [] new functionality
- [ ] a breaking change

## Description & motivation
- When datepart is `day` or greater, the `recency` test should compare against the current date instead of current timestamp
- The test wasn't working with `timestamp` columns in BigQuery before (since you can't compare a `timestamp` with `datetime`). This PR fixes that by converting timestamp to datetime first.
- Note: this PR was done in collaboration with @rodrigoaps

## Checklist
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake


macros/generic_tests/recency.sql
```diff
@@ -4,7 +4,21 @@
...
+
+{#- Use a date-based threshold if column type is DATE or datepart is day or greater #}
+{%- if column_type == 'DATE' and datepart in ['year', 'quarter', 'month', 'week', 'day'] %}
+  {%- set threshold = dbt.dateadd(datepart, interval * -1, dbt.safe_cast(dbt.current_timestamp_backcompat(), api.Column.translate_type("date"))) %}
+{%- else %}
+  {%- set threshold = dbt.dateadd(datepart, interval * -1, dbt.current_timestamp_backcompat()) %}
+{%- endif %}
+
+
+{%- set field_datetime = dbt.safe_cast(field, api.Column.translate_type("datetime")) %}
```
joellabes, line 21: This doesn't seem to be used anywhere?

macros/generic_tests/recency.sql
```diff
@@ -4,7 +4,21 @@
...
-{% set threshold = dbt.dateadd(datepart, interval * -1, current_timestamp_backcompat()) %}
+{#- Get columns in model #}
+{%- set columns = adapter.get_columns_in_relation(model) %}
+
+{#- Get column type of field in model #}
+{%- set column_type = columns | selectattr("name", "==", field) | map(attribute="data_type") | first %}
+
+{#- Use a date-based threshold if column type is DATE or datepart is day or greater #}
+{%- if column_type == 'DATE' and datepart in ['year', 'quarter', 'month', 'week', 'day'] %}
+  {%- set threshold = dbt.dateadd(datepart, interval * -1, dbt.safe_cast(dbt.current_timestamp_backcompat(), api.Column.translate_type("date"))) %}
```
joellabes, line 15:
```suggestion
  {%- set threshold = dbt.dateadd(datepart, interval * -1, dbt.safe_cast(dbt.current_timestamp_backcompat(), 'date')) %}
```

we don't _necessarily_ want `translate_type` to be leaking into user-facing code, but we don't actually have a `dbt.type_date()` macro in Core yet... this one is tricky. 

As far as I can tell, almost all databases call their date-only datatype `date`. The only exception I could easily find was SQLite which either does strings or integers for unix time, which isn't going to work out of the box anyway. 

All this to say I think we're OK to hardcode the string in on this occasion. Lmk if there's something important I've missed!

macros/generic_tests/recency.sql
```diff
@@ -4,7 +4,21 @@
...
+{%- else %}
+  {%- set threshold = dbt.dateadd(datepart, interval * -1, dbt.current_timestamp_backcompat()) %}
```
joellabes, line 17: There was a merge conflict which I had to resolve, and while I was there I changed these to `dbt.current_timestamp_backcompat()` which is the equivalent of what was `dbt_utils.current_timestamp()`.

joellabes (changes requested):
This is looking great @yuanminglee! A couple of comments and potential changes, let me know what you think! 

I'm aiming to cut the release candidate for 1.0 at the start of next week, so please also let me know if you won't be able to get to this before then.

macros/generic_tests/recency.sql
yuanminglee: LGTM! For my learning's sake, why is it that we don't want `translate_type` to leak into user-facing code?

macros/generic_tests/recency.sql
yuanminglee: Got it!

macros/generic_tests/recency.sql
yuanminglee: You're right! Missed this from a previous iteration of a solution, thanks for catching it.

yuanminglee, line 21:
```suggestion
```

yuanminglee, line 21:
```suggestion
```

yuanminglee:
> This is looking great @yuanminglee! A couple of comments and potential changes, let me know what you think!
> 
> I'm aiming to cut the release candidate for 1.0 at the start of next week, so please also let me know if you won't be able to get to this before then.

Thanks @joellabes! I've made the suggested change, and removed the unused line 😃

macros/generic_tests/recency.sql
joellabes: > why is it that we don't want `translate_type` to leak into user-facing code?

It's better for there to only be one consistent way to do things - if some parts of the package use dbt.type_int() and some use api.Column.translate_type('int') then it makes it a trickier maintenance burden.

macros/generic_tests/recency.sql
```diff
@@ -4,7 +4,19 @@
...
 
-{% set threshold = dbt.dateadd(datepart, interval * -1, current_timestamp_backcompat()) %}
+{#- Get columns in model #}
+{%- set columns = adapter.get_columns_in_relation(model) %}
+
+{#- Get column type of field in model #}
+{%- set column_type = columns | selectattr("name", "==", field) | map(attribute="data_type") | first %}
+
+{#- Use a date-based threshold if column type is DATE or datepart is day or greater #}
+{%- if column_type == 'DATE' and datepart in ['year', 'quarter', 'month', 'week', 'day'] %}
```
joellabes, line 14: The comment and the code disagree here (one says and, the other says or) - which one did you intend to implement? 

Regardless, can you say more about why the checking of the datepart is even necessary? The original issue (#414) only talks about issues caused by the datatype of the column being date-only. Dropping the time components for _all_ thresholds will return inaccurate results. Consider this situation: 

```yaml
      - dbt_utils.recency:
          datepart: day
          field: my_datetime_column
          interval: 1
```

As I write this, it's `2022-11-23T04:57:19Z`. By dropping the time component even when it's available, the threshold will be 2022-11-22T00:00:00Z instead of T04:57:19Z, so the validity window is 5 hours larger, which would lead to false negatives. For example a record created at 1am would be considered valid even though it's actually 4 hours too old. 

```suggestion
{#- Use a date-based threshold if column type is DATE #}
{%- if column_type == 'DATE' %}
```
Let me know if I'm missing something!

joellabes (changes requested):

joellabes:
@yuanminglee I spent some more time noodling on this today, and have changed my mind from what I put on the original issue, where I said: 

>We could ask the user to specify the granularity to which we truncate [threshold](https://github.com/dbt-labs/dbt-utils/blob/master/macros/schema_tests/recency.sql#L7), but it would be nicer for them if we [got the column](https://docs.getdbt.com/reference/dbt-jinja-functions/adapter#get_columns_in_relation) and used its [data_type](https://docs.getdbt.com/reference/dbt-classes#column) inside the macro instead.

This falls down in a couple of places, most significantly the fact that it doesn't work for ephemeral models where we can't get the column type. We could just use the old behaviour for CTEs, but a core tenet of dbt is that you can change materializations without any impacts to other parts of the experience. If someone changed to/from an ephemeral model, their tests would behave differently which I don't like. 

Alternatively we could refuse to test ephemeral models, but there's no good justification for that. 

Soooo instead I think we should just add an optional setting on the test (called something like `ignore_time_component`) which determines whether it should truncate to the day level or not. When that's set to true, I think we should truncate both columns - it would be weird for an ignore_time_component command to only impact one of the columns. It should basically never come up because I imagine people will only use this when their column is a date only, but I couldn't think of a concise way to express that so it'll range from a no-op to "consistent with what it says on the tin" if you do use it on a datetime column.

joellabes:
Closing in favour of #731 - sorry for all the hassle @yuanminglee!

PR #731, opened 2022-11-25
Recency truncate date option

### LINKED ISSUE #414: Recency does not work as expected for datepart day

joellabes:
resolves #414, replaces #708

This is a:
- [ ] documentation update
- [ ] new functionality
- [ ] a breaking change

## Description & motivation
#414 points out that when testing recency against a `date` column, the fact that the comparison timestamp is a datetime leads to false negatives. 

This PR adds an option to `ignore_time_component` which truncates the threshold timestamp (and the tested column, but that's assumed to already not have a time component if you're doing this so is somewhat redundant)

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake


dbeatty10 (commented):
Instead of introducing a `ignore_time_component` parameter, could the `field` expression be cast as the applicable timestamp data type instead?

I didn't try out such an implementation myself, but it seems like it would have less moving pieces if it could work.

joellabes:
> could the `field` expression be cast as the applicable timestamp data type instead?

The original community-member-created PR did this, but in my testing I discovered that it gave incorrect results: https://github.com/dbt-labs/dbt-utils/pull/708#discussion_r1030025015

joellabes:
Same situation here as 732:

Even if there are tweaks to make to the final result, I think this directionally reflects what will ultimately ship. In the interests of getting the RC out I'm going to merge this and we can make any changes in a separate PR if needed (our pals at Fivetran are blocked by this release being delayed)

PR #732, opened 2022-11-28
Output a warning when star finds no columns, not '*'

### LINKED ISSUE #681: dbt-utils "star" returns all columns even if all columns are listed in except list
### Describe the bug
When using the `star` dbutils function, if you (inadvertently) include *every* column of a table in the `except` list, star.sql returns `*`. This leads to unexpected behavior where the query attempts to return all columns. (I would expect star.sql to return an empty string in this situation.)

In the web app, the SQL compiles fine, adding the * in where expected. Bizarrely, it even previews just fine, as if the * were not there. The issue only arises when doing `dbt run`. 

### Steps to reproduce
A simplified example of where I discovered it. 

```
with salesorderdetail as (
  select * from {{ ref( 'fact_sales_order_detail') }}
),

dim_product as (
  select 
    column_a,
    column_b,
    column_c
  from {{ ref('dim_product') }}
),

select
    sod.*,
    {{ dbt_utils.star(from=ref('dim_product'), relation_alias='pr', except=["column_a", "column_b", "column_c"]) }},
from salesorderdetail sod
    left join dim_product pr
        on pr.column_a = sod.product_id
```

### Expected results
I suppose I would expect a compile error to be thrown, spurred by an error return from `star`. 

### Actual results
The query compiles just fine, adding a `*` in where the `star` function was called. No error is even seen when previewing the query, although you can see the `*` in the compiled query. Specifically, the above query would compile to:

```
with salesorderdetail as (

  select * from [long string pointing to fact_sales_order_detail]
),

dim_product as (
  
  select 
    column_a,
    column_b,
    column_c
  from [long string pointing to dim_product]
),

select
    sod.*,
    *,
from salesorderdetail sod
    left join dim_product pr
        on pr.column_a = sod.product_id
```

This attempts to load all the columns of `sod` twice, leading to a compile error due to having multiple columns with identical names. 

### System information
```
packages:
  - package: dbt-labs/dbt_utils
    version: 0.9.1
```

- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x] snowflake

1.2.0

Sure but not 100% sure how. There needs to be some handling to determine if all columns appear in the `except` list, either in `star` or in `get_filtered_columns_in_relation` I believe. Perhaps a check as simple as if the `except` list length is equal to (or greater) the table's column list length.

joellabes:
resolves #681 

This is a:
- [ ] documentation update
- [ ] new functionality
- [x] a breaking change

## Description & motivation
#561 solved a SQLFluff linting error where the `star()` macro returned an empty string if there were no columns found (either because the model doesn't exist or because all columns were excluded). 

The side effect of this is that if someone excludes all columns, they get `'*'` (every column) back instead, which is the exact opposite of what you'd expect. 

This change means that if no columns are found, it returns `'no columns returned from star() macro' as outcome_msg` instead. 

This is still valid SQL, so SQLFluff won't complain about it, but it better represents the outcome that got us here than a `*`. 

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake

### DIFF
--- a/integration_tests/models/sql/schema.yml
+++ b/integration_tests/models/sql/schema.yml
@@ -180,7 +180,7 @@ models:
 
   - name: test_star_no_columns
     columns: 
-      - name: outcome_msg #don't want to check for the specific error message, but do want to check that the error column is there
+      - name: canary_column #If the no-columns state isn't hit, this table won't be queryable because there will be a missing comma
         tests: 
           - not_null
 
--- a/integration_tests/models/sql/schema.yml
+++ b/integration_tests/models/sql/schema.yml
@@ -178,6 +178,12 @@ models:
       - dbt_utils.equality:
           compare_model: ref('data_star_expected')
 
+  - name: test_star_no_columns
+    columns: 
+      - name: outcome_msg #don't want to check for the specific error message, but do want to check that the error column is there
+        tests: 
+          - not_null
+
   - name: test_generate_surrogate_key
     tests:
       - assert_equal:
--- a/integration_tests/models/sql/test_star_no_columns.sql
+++ b/integration_tests/models/sql/test_star_no_columns.sql
@@ -2,7 +2,8 @@ with data as (
 
     select
         {{ dbt_utils.star(from=ref('data_star'), except=['field_1', 'field_2', 'field_3']) }}
-
+        -- if star() returns `*` or a list of columns, this query will fail because there's no comma between the columns
+        1 as canary_column
     from {{ ref('data_star') }}
 
 )
--- a/integration_tests/models/sql/test_star_no_columns.sql
+++ b/integration_tests/models/sql/test_star_no_columns.sql
@@ -1,0 +1,10 @@
+with data as (
+
+    select
+        {{ dbt_utils.star(from=ref('data_star'), except=['field_1', 'field_2', 'field_3']) }}
+
+    from {{ ref('data_star') }}
+
+)
+
+select * from data
--- a/macros/sql/star.sql
+++ b/macros/sql/star.sql
@@ -8,13 +8,23 @@
 
     {#-- Prevent querying of db in parsing mode. This works because this macro does not create any new refs. #}
     {%- if not execute -%}
-        {{ return('*') }}
+        {% do return('*') %}
     {%- endif -%}
 
     {% set cols = dbt_utils.get_filtered_columns_in_relation(from, except) %}
 
     {%- if cols|length <= 0 -%}
-      {{- return(dbt.string_literal("no columns returned from star() macro") ~ " as outcome_msg") -}}
+        {% if flags.WHICH == 'compile' %}
+            {% set response %}
+*
+/* No columns were returned. Maybe the relation doesn't exist yet 
+or all columns were excluded. This star is only output during  
+dbt compile, and exists to keep SQLFluff happy. */
+            {% endset %}
+            {% do return(response) %}
+        {% else %}
+            {% do return("/* no columns returned from star() macro */") %}
+        {% endif %}
     {%- else -%}
         {%- for col in cols %}
             {%- if relation_alias %}{{ relation_alias }}.{% else %}{%- endif -%}
--- a/macros/sql/star.sql
+++ b/macros/sql/star.sql
@@ -14,7 +14,7 @@
     {% set cols = dbt_utils.get_filtered_columns_in_relation(from, except) %}
 
     {%- if cols|length <= 0 -%}
-      {{- return('*') -}}
+      {{- return(dbt.string_literal("no columns returned from star() macro") ~ " as outcome_msg") -}}
     {%- else -%}
         {%- for col in cols %}
             {%- if relation_alias %}{{ relation_alias }}.{% else %}{%- endif -%}


dbeatty10:
I buy into `*` being [surprising / astonishing](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) when there are zero selected columns. But in that case, I'm wondering (and hoping!) if we can do better than adding an unrequested `outcome_msg` column.

### Example
- If we query for a list of English words that rhyme with "dangerous", we'd expect an empty list.
- We'd be _astonished_ if we got a list containing every word in the dictionary. Likewise, we'd be surprised by a list with one item: "no rhyming words found".

### Counter-proposal

Considering the four different scenarios between `execute` and  `compile` vs `run`, as-is this PR would yield the following:

|                    | `dbt compile` | `dbt run`/`build` |
|--------------------|---------------|-------------------|
| `execute == False` | `*`           | `'no columns returned from star() macro' as outcome_msg`               |
| `execute == True`  | `*`           | `'no columns returned from star() macro' as outcome_msg`               |

What would stop us from implementing the following instead?

|                    | `dbt compile` | `dbt run`/`build` |
|--------------------|---------------|-------------------|
| `execute == False` | `*`           | `*`               |
| `execute == True`  | `*`           | empty string¹     |

We could do something like [here](https://github.com/dbt-labs/dbt-utils/pull/607/files#diff-e1ea6fc39c905553c4c68b14b4e5965568cf79383a0112201303541667202550R63-R66) to determine the dbt sub-command and handle the `execute == True` cases like:

```sql
    {%- if cols|length <= 0 and flags.WHICH == 'compile' -%}
      {{- return('*') -}}
    {%- elif cols|length <= 0 -%}
      {{- return('') -}}
    {%- else -%}
```

---

### Notes

¹An alternative to an empty string would be to return an error as described in https://github.com/dbt-labs/dbt-utils/issues/681.

https://github.com/dbt-labs/dbt-utils/issues/605 was concerned with the combo of `execute == True` and `dbt compile`.

https://github.com/dbt-labs/dbt-utils/issues/681 is concerned with the combo of `execute == True` and `dbt run`/`build`.

dbeatty10 (changes requested):
Counter-proposal included in the comments.

I'm not a pro with `flags.WHICH`, `execute`, or SQLFluff, so I might be misconstruing or misunderstanding how things work in my counter-proposal. But regardless of the details, I would advocate for an implementation different than what we have in this PR.

joellabes:
I'm answering this assuming that the current summary table's rows and columns are the wrong way around (we currently return `'*'` in all cases when `execute` is false)

With that said, I really like this idea! Thanks for digging up the history on different invocation commands. 

I think checking against `flags.WHICH` is a goer.

joellabes:
OK in `compile` mode I output: 
- a star `*`
- a comment explaining that the behaviour is different to what happens during `dbt run` invocations etc

In all other modes, I output:
- a comment that says no columns were found. 

This was actually what I originally wanted to do yesterday (instead of creating a real column in the output), but I realised that SQLFluff was probably smart enough to realise that a comment wasn't columns either 😬 I hadn't thought of using the WHICH flag. 

Why a comment instead of a blank string as you suggested? In the past I have been tripped up by star returning nothing and not being able to work out why - this feels like a more useful pointer while still accurately reflecting the query results that will happen.

joellabes:
Even if there are tweaks to make to the final result, I think this directionally reflects what will ultimately ship. In the interests of getting the RC out I'm going to merge this and we can make any changes in a separate PR if needed (our pals at Fivetran are blocked by this release being delayed)

dbeatty10:
> I'm answering this assuming that the current summary table's rows and columns are the wrong way around (we currently return `'*'` in all cases when `execute` is false)

Yeah, the table had some values flip-flopped! 😅  For completeness, I'll update the table in [this](https://github.com/dbt-labs/dbt-utils/pull/732#issuecomment-1329942693) comment to be:

|                    | `dbt compile` | `dbt run`/`build` |
|--------------------|---------------|-------------------|
| `execute == False` | `*`           | `*`               |
| `execute == True`  | `'no columns returned from star() macro' as outcome_msg`           | `'no columns returned from star() macro' as outcome_msg`               |

instead of the incorrect flip-flopped version:

|                    | `dbt compile` | `dbt run`/`build` |
|--------------------|---------------|-------------------|
| `execute == False` | `*`           | `'no columns returned from star() macro' as outcome_msg`               |
| `execute == True`  | `*`           | `'no columns returned from star() macro' as outcome_msg`               |

dbeatty10:
### Feedback on implementation

This strategy you proposed looks great to me:

|                    | `dbt compile` | `dbt run`/`build` |
|--------------------|---------------|-------------------|
| `execute == False` | `*`           | `*`               |
| `execute == True`  | `/* a comment */`<br>`*`           | `/* another comment */`               |

But in the implementation, I didn't see the `*` during `dbt compile` and `execute == True` that is meant for SQLFluff. Did I overlook it? Is it covered the way you intended?

dbeatty10:
Of lesser importance, I am now also curious about cross-database compatibility of single line `--` and multi-line `/* */` comments.

joellabes:
> I didn't see the `*` during `dbt compile` and `execute == True` that is meant for SQLFluff. Did I overlook it?

I think so! It's here: 
https://github.com/dbt-labs/dbt-utils/pull/732/files#diff-b009b0ae1dc2f9c557c2313910e67fa8c90a78490910b5780fcf3caed2087819R19

> Of lesser importance, I am now also curious about cross-database compatibility of single line `--` and multi-line `/* */` comments.

I always use /*bounded comments*/ inside of Jinja macros because I can't guarantee what `{{- whitespace management -}}` end users are going to do. I also had the same thought - if it becomes a problem, we can add `dbt.multiline_comment()` for the one adapter that woke up and chose violence 😬

dbeatty10:
Cool -- I see the `*` now 👁️ ! Admittedly, I didn't expect it _before_ the comment or fully left-justified, but I can guess some reasons why you put it there (namely prioritizing the indentation of the `*` over the comment during Jinja rendering).

Speaking of comments... agreed with your `dbt.multiline_comment()` strategy (if needed).

But might not be necessary; the [IBM Informix docs](https://www.ibm.com/docs/en/informix-servers/14.10?topic=syntax-how-enter-sql-comments) claim that:
- Double hyphen ( `--` ) complies with the ANSI/ISO standard for SQL.
- C-style slash-and-asterisk ( `/*` . . . `*/` ) comply with the SQL-99 standard.

vperron:
Hi ! I stumbled across this issue but it seems that in my case, `sqlfluff` does **not** enter in the WHICH = "compile" mode. I took a very deep look at the invocation of the DBT compiler and did not see anything too fancy, but it is possible that this is a regression and SQLFluff does not pass the `compile` mode anymore. Anyone else has noticed this ?

My config has nothing very fancy:
```
[tool.sqlfluff.core]
dialect = "postgres"
exclude_rules = "ambiguous.distinct,layout.long_lines"
templater = "dbt"
max_line_length = 119

[tool.sqlfluff.templater.jinja]
apply_dbt_builtins = true
load_macros_from_path = "dbt/packages"

[tool.pytest.ini_options]
addopts = "--ignore=dbt/packages"
```
and it started breaking immediately after the first inclusion of our `star()` macro against an empty database.

vperron:
To be clearer, it seems that the SQLFluff mitigation that is wisely decided here is not used by `sqlfluff` anymore, or there is some configuration switch that I still have to do.

dbeatty10:
What error message are you seeing @vperron ?

vperron:
Hello ! 
```
sqlfluff lint --exclude-rules ambiguous.distinct,layout.long_lines,references.consistent,references.from,references.qualification --disable-progress-bar --nocolor dags dbt tests
=== [dbt templater] Sorting Nodes...                                                                                                                                                                               
16:22:59  Unable to do partial parsing because env vars used in profiles.yml have changed                                                                                                                          
16:22:59  [WARNING]: Configuration paths exist in your dbt_project.yml file which do not apply to any resources.
There are 1 unused configuration paths:                                                                                                                                                                            
- models.dbt_pilotage.meta                                                                                                                                                                                         
=== [dbt templater] Compiling dbt project...                                                                                                                                                                       
=== [dbt templater] Project Compiled.                                                                                                                                                                              
== [dbt/models/marts/suivi_candidatures_prescripteurs_habilites.sql] FAIL                                                                                                                                          
L:   2 | P:  11 |  PRS | Line 2, Position 11: Found unparsable section: '\n                              
                       | /* no columns returned from sta...'                                                                                                                                                       
WARNING: Parsing errors found and dialect is set to 'postgres'. Have you configured your dialect correctly?
All Finished!               
```
The original source code is:
```
{{ dbt_utils.star(ref('candidatures_echelle_locale')) }},
```

The issue is that it is replaced by: (note the comma)

```
/* no columns returned from star() macro */,
```
which does not parse because of the comma. It clearly was the intent though, seemingly with this PR, to have:
```
*
/* No columns were returned. Maybe the relation doesn't exist yet 
or all columns were excluded. This star is only output during  
dbt compile, and exists to keep SQLFluff happy. */
```
instead, which would indeed have made SQLFluff and myself happy, but it chose the other path (the "non compile" path).

It seems from my debug sessions that the `WHICH` variable is set to `None` with my current sqlfluff setup, which seems to indicate either a configuration issue on my end (but it's not very exotic) or a regression somewhere.

dbeatty10:
Thanks for the write-up @vperron 🤩

Here's my very simple dbt project:

`models/my_model.sql`
```sql
select 1 as id
```

`models/my_model_2.sql`
```sql
select {{ dbt_utils.star(ref('my_model')) }}, 'me' as my_name
```

Starting with a completely empty database, here's the output of `dbt compile`:
```sql
select 
*
/* No columns were returned. Maybe the relation doesn't exist yet 
or all columns were excluded. This star is only output during  
dbt compile, and exists to keep SQLFluff happy. */
            , 'me' as my_name
```

Output of `sqlfluff render models/my_model_2.sql`:
```sql
select /* no columns returned from star() macro */, 'me' as my_name
```

### Proposed next steps

@vperron Could you open a new issue in dbt-utils that references the relevant discussion here? We'll want to coordinate with the maintainers of SQLFluff on that new issue to determine a path forward.

cc: @alanmcruickshank

tlfbrito:
I'm experiencing the same issue

katieclaiborne:
@dbeatty10 @vperron, thanks for the investigation!

Has a new issue been created to determine a path forward? I'm happy to create one, but wanted to double-check here first.

vperron:
Hello ! No I am just coming back from 3 weeks of vacation, sorry about this. So yeah, a new issue should be created as @dbeatty10 pointed out. Specifically, it seems that `dbt-utils` is working just as expected but the link between `sqlfluff` and `dbt` seems to not use the `compile` mode: the `WHICH` variable is set to `None`, which is surprising.

I am a little overwhelmed today since I'm just getting back to work, but I'll happily add my input to the issue in a few days if you create one.

PR #737, opened 2022-12-08
Add option to ignore columns in equality test

### LINKED ISSUE #734: Add option to ignore columns in `equality` test
### Describe the feature

I would like to be able to specify a list of columns to exclude on an `equality` test.

We can of course use `compare_columns` to list all columns, except the ones we want to exclude, but this leads to large yml files, difficult to maintain as our models evolve.

### Additional context

Not database specific, it can be implemented across all different databases.

### Who will this benefit?

This feature would be great for anyone who wants to compare data across environments where, for example, some of the fields are masked in one of the environments.
Another scenario where this can be useful is when updating the logic on a small subset of fields in a model and wanting to compare the development model with the production one, but obviously excluding the updated fields (which are expected to be different)

I have implemented this locally already. Happy to raise a PR if people think this feature is helpful.

brunocostalopes:
resolves #734

This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation
This is to include an option to ignore columns in the equality test. Currently the only way to do this is to use the `compare_columns` option and include all columns we want to compare, leaving out any that we want to exclude. In certain cases this might result in large yml files, difficult to maintain. The `ignore_columns` option fixes that.

## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [x] Snowflake


macros/generic_tests/equality.sql
```diff
@@ -1,10 +1,14 @@
...
+{% test equality(model, compare_model, compare_columns=None, ignore_columns=None) %}
+  {{ return(adapter.dispatch('test_equality', 'dbt_utils')(model, compare_model, compare_columns, ignore_columns)) }}
 {% endtest %}
 
-{% macro default__test_equality(model, compare_model, compare_columns=None) %}
+{% macro default__test_equality(model, compare_model, compare_columns=None, ignore_columns=None) %}
 
-{% set set_diff %}
+{%- if compare_columns and ignore_columns -%}
+    {{ exceptions.raise_compiler_error("Both an compare and ignore list were provided to the `equality` macro. Only one is allowed") }}
```
deanna-minnick, line 8:
```suggestion
    {{ exceptions.raise_compiler_error("Both a compare and an ignore list were provided to the `equality` macro. Only one is allowed") }}
```

deanna-minnick: one small tweak here!

brunocostalopes: thanks for the suggestion!

joellabes:
Thanks @brunocostalopes - seems like something odd is happening with the CI, I'll have to look into that! Code itself looks good though 🎉

moreaupascal56:
Hello! Nice, would love this feature! Just a suggestion:

don't you think you could remove the elements from `compare_columns` instead of creating a second list ? 
the loop would be:

for column in ignore_columns:
  compare_columns.remove(column)

So it would use only existing objects and loop only on ignore_columns elements (which schould be fewer than compare_columns ones) . 
I do not really use Jinja a lot but the remove syntax is a Python one so it is maybe possible.

brunocostalopes:
@moreaupascal56 thanks for the tip. I think you are right. If we remove instead of append it makes the code neater and I've pushed those changes.

I've also tried your other suggestion of looping through the elements in the ignore list first but that would make the code a bit more complex when removing from the original compare list so I've decided against it as I don't think the potential performance gains would be that relevant here.

@joellabes seems like the checks are passing again, thank you for looking into it. I think the PR is now ready for your review.

pauldalewilliams:
Excited to see this come through - have been running into a need for this during legacy code conversion.

pauldalewilliams:
@joellabes @deanna-minnick Can this PR get reviewed and approved?  I'd love to have this feature available.

brunocostalopes:
I've just realised that the latest change I've pushed has a bug which I can't explain.

This is the code with the issue:
```
    {%- for column in compare_columns -%}
        {%- if column | lower in ignore_columns -%}
            {% do compare_columns.remove(column) %}
        {%- endif %}
    {%- endfor %}
```

whenever a match is found a column is removed from `compare_columns`, the next column in the list is ignored and it is not tested against the `ignore_columns` list.

I can't understand why this is happening. If anyone has a clue please let me know, I would be happy to fix. In the meantime, I'm going to revert back to the previous version which, even thought it is not the tidiest implementation, it seems to work well.

pauldalewilliams:
> I've just realised that the latest change I've pushed has a bug which I can't explain.
> 
> This is the code with the issue:
> 
> ```
>     {%- for column in compare_columns -%}
>         {%- if column | lower in ignore_columns -%}
>             {% do compare_columns.remove(column) %}
>         {%- endif %}
>     {%- endfor %}
> ```
> 
> whenever a match is found and a column is removed from `compare_columns`, the next column in the list is ignored and it is not tested against the `ignore_columns` list.
> 
> I can't understand why this is happening. If anyone has a clue please let me know, I would be happy to fix. In the meantime, I'm going to revert back to the previous version which, even thought it is not the tidiest implementation, it seems to work well.

I believe @moreaupascal56's suggestion was that you wouldn't loop over compare_columns anyway.  Why not try it like this?  Should avoid the bug you're seeing and be a faster loop anyway.

```
    {%- for column in ignore_columns -%}
        {%- if column | lower in compare_columns -%}
            {% do compare_columns.remove(column) %}
        {%- endif %}
    {%- endfor %}
```

PR #748, opened 2022-12-17
Safe subtract

### LINKED ISSUE #745: It would helpful to have a subtraction equivalent of the `safe_add()` macro
Essentially this macro would function identically to [`safe_add()`](https://github.com/dbt-labs/dbt-utils#safe_add-source) but would subtract values instead. This would help eliminate the need to hack the `safe_add()` macro by multiplying values by -1 or by needing to do this without the **dbt_utils** macro using `coalesce` statements.

This function would be beneficial whenever there is a need to handle subtraction involving null values at a row level. I would be open to submitting this as PR if there is interest. I think it would require just one change to this line (but correct me if I'm wrong):
https://github.com/dbt-labs/dbt-utils/blob/42003d7f6dfeb83c8ff3169d6911cc60f5b54545/macros/sql/safe_add.sql#L26

dchess:
resolves #745

This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation
As discussed in #745, this new macro will provide a method for taking the difference across nullable fields similar in functionality to `safe_add()`. Currently it is necessary to multiply fields by -1 with the `safe_add()` macro or use coalesce statements in sql directly to achieve the same result.

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Check [dbt Hub](https://hub.getdbt.com/dbt-labs/dbt_utils/latest/) for the lates
   - [generate_surrogate_key](#generate_surrogate_key-source)
   - [safe_add](#safe_add-source)
   - [safe_divide](#safe_divide-source)
+  - [safe_subtract](#safe_subtract-source)
   - [pivot](#pivot-source)
   - [unpivot](#unpivot-source)
   - [width_bucket](#width_bucket-source)
@@ -1069,6 +1070,16 @@ This macro performs division but returns null if the denominator is 0.
 
 ```
 {{ dbt_utils.safe_divide('numerator', 'denominator') }}
+```
+
+#### safe_subtract ([source](macros/sql/safe_subtract.sql))
+
+This macro implements a cross-database way to take the difference of nullable fields using the fields specified.
+
+**Usage:**
+
+```
+{{ dbt_utils.safe_subtract('field_a', 'field_b'[,...]) }}
 ```
 
 #### pivot ([source](macros/sql/pivot.sql))
--- a/README.md
+++ b/README.md
@@ -1054,7 +1054,7 @@ This macro implements a cross-database way to sum nullable fields using the fiel
 **Usage:**
 
 ```
-{{ dbt_utils.safe_add('field_a', 'field_b'[,...]) }}
+{{ dbt_utils.safe_add(['field_a', 'field_b', ...]) }}
 ```
 
 #### safe_divide ([source](macros/cross_db_utils/safe_divide.sql))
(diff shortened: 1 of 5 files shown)


macros/sql/safe_subtract.sql
```diff
@@ -0,0 +1,28 @@
...
+{%- macro default__safe_subtract(field_list) -%}
+
+{%- if field_list is not iterable or field_list is string or field_list is mapping -%}
+
+{%- set error_message = '
+Warning: the `safe_subtract` macro now takes a single list argument instead of \
+string arguments. The {}.{} model triggered this warning. \
+'.format(model.package_name, model.name) -%}
+
+{%- do exceptions.warn(error_message) -%}
```
joellabes, line 14: In general, I don't think this note is necessary - we had it for `safe_add` as part of its behaviour migration, but this is a net-new macro that is being built the right way. 

If someone does `{{ dbt_utils.safe_subtract('first', 'second') }}` it will render as `coalesce(f, 0) - coalesce(i, 0), ...` which will send them back to double check the docs anyway. 

So two options: 
- Remove altogether
- Change it to an exception and don't say "now takes", because it never didn't take that form

joellabes (changes requested):
this is looking great! sorry for the delay in getting back to you with a review 🌟 

A couple of changes but overall very excited to get this in

macros/sql/safe_subtract.sql
dchess: That makes total sense to me. I think since the shift is still fairly new it makes sense to throw an exception and have updated it accordingly, but let me know if you would prefer to remove altogether instead.

README.md
```diff
@@ -1071,6 +1072,16 @@ This macro performs division but returns null if the denominator is 0.
...
 ```
 
+#### safe_subtract ([source](macros/sql/safe_subtract.sql))
+
+This macro implements a cross-database way to take the difference of nullable fields using the fields specified.
+
+**Usage:**
+
+```
+{{ dbt_utils.safe_subtract('field_a', 'field_b'[,...]) }}
```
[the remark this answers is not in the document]
dchess: Updated!

dchess:
@joellabes looks like the redshift integration tests failed after my updates. Can you share the errors or any insights you may have about the reason? I'd also love to get this in.

dchess:
@joellabes Sorry not used to CI being publicly visible! I was able to access the details myself. 

It looks like the seed files aren't loading due to a permission error:
> permission denied for schema dbt_utils_integration_tests_redshift

joellabes:
@dchess I'm looking into this!

joellabes (approved):
Redshift is passing again, so we're in business! Thanks for getting this over the line @dchess 

(And thanks to @dave-connors-3 and @dbeatty10 for the assist with CI here!)

PR #765, opened 2023-02-06
add precision + exclude_columns option to equality test

### LINKED ISSUE #734: Add option to ignore columns in `equality` test

### LINKED ISSUE #757: Add Precision argument to `equality` for numeric and float type columns
### Describe the feature
It would be helpful to have an argument to the `equality` test that would specify the precision required for equality for numeric and float type columns by first rounding them before comparison. Currently equality can fail on seemingly identical tables due to precision errors as part of processing.

Removing these columns from the comparison, or implementing an overwritten macro to do this myself.

### Who will this benefit?
Anyone using the equality test with numeric/float type columns.

I have a working solution that I use locally based off the existing `equality` code, it works fine but is maybe not the cleanest way to do it.

```jinja2
{% test equality(model, compare_model, compare_columns=None, precision = None) %}
  {{ return(adapter.dispatch('test_equality', 'dbt_utils')(model, compare_model, compare_columns, precision)) }}
{% endtest %}

{% macro default__test_equality(model, compare_model, compare_columns=None, precision = None) %}

{% set set_diff %}
    count(*) + coalesce(abs(
        sum(case when which_diff = 'a_minus_b' then 1 else 0 end) -
        sum(case when which_diff = 'b_minus_a' then 1 else 0 end)
    ), 0)
{% endset %}

{#-- Needs to be set at parse time, before we return '' below --#}
{{ config(fail_calc = set_diff) }}

{%- if not execute -%}
    {{ return('') }}
{% endif %}

-- setup
{%- do dbt_utils._is_relation(model, 'test_equality') -%}

{%- if not precision -%}
    {#-
        If the compare_cols arg is provided, we can run this test without querying the
        information schema — this allows the model to be an ephemeral model
    -#}
    {%- if not compare_columns -%}
        {%- do dbt_utils._is_ephemeral(model, 'test_equality') -%}
        {%- set compare_columns = adapter.get_columns_in_relation(model) | map(attribute='quoted') -%}
    {%- endif -%}

    {% set compare_cols_csv = compare_columns | join(', ') %}
{% else %}
    {#-
        If rounding is required, we need to get the types, so it can't be ephermeral
    -#}
    {%- do dbt_utils._is_ephemeral(model, 'test_equality') -%}
    {%- set columns = adapter.get_columns_in_relation(model) -%}

    {% set columns_list = [] %}
    {%- for col in columns -%}
        {%- if (compare_columns and col.name|lower in compare_columns|map('lower')) or not compare_columns -%}
            {%- if col.is_float() or col.is_numeric() -%}
                {# Cast is required due to postgres not having round for a double precision number #}
                {%- do columns_list.append('round(cast(' ~ col.name ~ ' as ' ~ dbt.type_numeric() ~ '),' ~ precision ~ ') as ' ~ col.name) -%}
            {%- else -%}
                {%- do columns_list.append(col.name) -%}
            {%- endif -%}
        {% endif %}
    {%- endfor -%}

    {% set compare_cols_csv = columns_list | join(', ') %}
{% endif %}

with a as (

    select * from {{ model }}

),

b as (

    select * from {{ compare_model }}

),

a_minus_b as (

    select {{compare_cols_csv}} from a
    {{ dbt.except() }}
    select {{compare_cols_csv}} from b

),

b_minus_a as (

    select {{compare_cols_csv}} from b
    {{ dbt.except() }}
    select {{compare_cols_csv}} from a

),

unioned as (

    select 'a_minus_b' as which_diff, a_minus_b.* from a_minus_b
    union all
    select 'b_minus_a' as which_diff, b_minus_a.* from b_minus_a

)

select * from unioned

{% endmacro %}
```

### LINKED ISSUE #785: `equality` test passes when first table has a subset of second tables columns
### Describe the bug
When the `equality` test is used to compare two models and `compare_columns` is not provided the macro currently gets the list of columns from the first model only. If there are additional or different columns in this model when compared with `compare_model` the sql will fail to run and return an error, however when this first model contains a subset of the columns from the compare column no error occurs. 

### Steps to reproduce
1. Create a project, add dbt-utils as as dependency, then create two seeds (for simplicity, the same applies to models), example:
    ```
    # data1.csv
    col1,col2
    a,b
    c,d

    # data2.csv
    col1,col2,col3
    a,b,y
    c,d,z
    ```
2. Add a test on the first seed with `compare_model` set to the second seed, and the inverse
      ```
    version: 2
  
    seeds:
      - name: data1
        tests:
          - dbt_utils.equality:
              compare_model: ref('data2')
  
      - name: data2
        tests:
          - dbt_utils.equality:
              compare_model: ref('data1')
    
    ```
5. run `dbt deps`, `dbt seed`, `dbt test`. The first test should pass, but the second will fail.

### Expected results
Both tests would fail in the same way, or a custom alert would warn me the columns are different.

### Actual results
Test 1 passes, test 2 fails.

```
11:17:41  Running with dbt=1.4.5
11:17:42  Found 2 models, 2 tests, 0 snapshots, 0 analyses, 403 macros, 0 operations, 2 seed files, 0 sources, 0 exposures, 0 metrics
11:17:42  
11:17:42  Concurrency: 1 threads (target='postgres')
11:17:42  
11:17:42  1 of 2 START test dbt_utils_equality_data1_ref_data2_ .......................... [RUN]
11:17:42  1 of 2 PASS dbt_utils_equality_data1_ref_data2_ ................................ [PASS in 0.06s]
11:17:42  2 of 2 START test dbt_utils_equality_data2_ref_data1_ .......................... [RUN]
11:17:42  2 of 2 ERROR dbt_utils_equality_data2_ref_data1_ ............................... [ERROR in 0.04s]
11:17:42  
11:17:42  Finished running 2 tests in 0 hours 0 minutes and 0.23 seconds (0.23s).
11:17:42  
11:17:42  Completed with 1 error and 0 warnings:
11:17:42  
11:17:42  Database Error in test dbt_utils_equality_data2_ref_data1_ (models/test.yml)
11:17:42    column "col3" does not exist
11:17:42    LINE 48:     select "col1", "col2", "col3" from b
11:17:42                                        ^
11:17:42    HINT:  There is a column named "col3" in table "*SELECT* 1", but it cannot be referenced from this part of the query.
11:17:42    compiled Code at target/run/dbt_demo/models/test.yml/dbt_utils_equality_data2_ref_data1_.sql
11:17:42  
11:17:42  Done. PASS=1 WARN=0 ERROR=1 SKIP=0 TOTAL=2
```

### System information
- [x] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake

```
Core:
  - installed: 1.4.5
  - latest:    1.4.6 - Update available!

Plugins:
  - databricks: 1.4.2 - Update available!
  - bigquery:   1.4.3 - Up to date!
  - snowflake:  1.4.2 - Up to date!
  - redshift:   1.4.0 - Up to date!
  - postgres:   1.4.5 - Update available!
  - spark:      1.4.1 - Up to date!

```

### Additional context
https://github.com/dbt-labs/dbt-utils/blob/9507d6bcd6c04b73e174a9aaee11865f9387e096/macros/generic_tests/equality.sql#L30-L33

Sure, I think it should just need to pull the columns from both models and compare them, I could add it as part of #765 to save touching it twice if that's easier.

rlh1994:
resolves #757, resolves #734, resolves #785, replaces #737

This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation
This adds an argument to the `equality` test that allows for floating point columns (column `is_numeric` or `is_float`) to be more easily compared and reducing false errors raised. It is fully backwards compatible (default arguments retain existing behaviour and code), and has been tested on and supports BQ/Postgres/Databricks/Snowflake/Redshift. 

It also includes a few new tests both positive and negative assertions.

It also includes the work done in #737 to save rebasing if one of these was merged independently.

Finally, it adds a fix for #785 that ensures when no compare/ignore columns list is provided that both tables have the same columns.

## Checklist
    - [x] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake
    - [x] Databricks


adammarples:
@rlh1994 this PR would be very useful to me. Any idea why the tests are failing? It looks like errors in seeding on redshift but it's not clear why.

joellabes (commented):
You and @brunocostalopes have done some great stuff here @rlh1994! 

I particularly love the robust set of tests. 

I will want to come back and do a deeper read of the specific jinja in the equality.sql file itself, but it makes sense to do that in one hit alongside any changes necessary with #785, as I imagine that'll lead to some changes too. 

The good news is that the tests are now up and running again on Redshift, so we should be able to move much faster and more confidence.

rlh1994:
@joellabes I've added it in, but a few notes on this final change:

1. I've made it a separate block, which means that currently it hits the warehouse twice for the model to get the columns - I tried to do this as part of the normal `get_columns...` part but it seems like `set()` alters the object it is acting on (at least they way I was using it for the generator that's returned) which cause a load of new failures when it shouldn't have. I'm not aware if there is the equivalent of a `deep_copy` in dbt/jinja so I've done it this way just to be sure.
2. It feels like this should actually raise a compiler error rather than fail the test itself, but I wasn't sure how to add tests for such a case, let me know your preference on this and I can swap it out to raise an error instead.
3. I think this would not qualify as a breaking change, as previously passing tests would now fail. A workaround would be to add a new argument to the macro to enable/disable this check and set the default (until a breaking change version is released) to be `false`.
4. I had to make a change to one of the other tests as this highlighted that the tables had different columns (which means it works!)

nachimehta:
also looking forward to this, at the moment the equality silently passes if your comparison model has more columns than the tested model, which seems like unexpected behavior. I believe your fix rectifies this issue.

joellabes:
> * I'm not aware if there is the equivalent of a `deep_copy` in dbt/jinja so I've done it this way just to be sure.
Me either, I think this is OK

> * It feels like this should actually raise a compiler error rather than fail the test itself, but I wasn't sure how to add tests for such a case, let me know your preference on this and I can swap it out to raise an error instead.

Yeah it should raise a compiler error. I think that means we can't have a test running for it 😬 but better to do the right thing on the end user's side. 

> * I think this would now qualify as a breaking change, as previously passing tests would now fail. A workaround would be to add a new argument to the macro to enable/disable this check and set the default (until a breaking change version is released) to be `false`.

They’ll fail because the data is invalid though, right? So I am OK with that - it's a bug that we're fixing as opposed to deviating from documented behaviour 

> 4\. I had to make a change to one of the other tests as this highlighted that the tables had different columns (which means it works!)

Ha! Great news 🎉 

Will go over this properly soon

rlh1994:
Hey @joellabes just wondering if you know when you'll have a chance to look at this please?

macros/generic_tests/equality.sql
```diff
@@ -19,20 +23,108 @@
...
+        {%- else -%}
+            {%- set compare_columns = compare_columns | map(attribute='quoted') %}
+        {%- endif -%}
+    {%- endif -%}
+
+    {% set compare_cols_csv = compare_columns | join(', ') %}
+
+{% else %}
+    {#-
+        If rounding is required, we need to get the types, so it cannot be ephermeral
```
seub, line 104: typo: ephermeral

brunocostalopes:
@rlh1994 I noticed you've updated the `equality` test and renamed the arguments from ignore to exclude, just a note that when discussing with @joellabes he was of the opinion that we should name them "ignore" rather than "exclude". See here: https://github.com/dbt-labs/dbt-utils/issues/734#issuecomment-1340264441

rlh1994:
> @rlh1994 I noticed you've updated the `equality` test and renamed the arguments from ignore to exclude, just a note that when discussing with @joellabes he was of the opinion that we should name them "ignore" rather than "exclude". See here: [#734 (comment)](https://github.com/dbt-labs/dbt-utils/issues/734#issuecomment-1340264441)

Ah okay, I went with exclude based on this comment here: https://github.com/dbt-labs/dbt-utils/pull/829#issuecomment-1777254295

To be honest I have no strong feelings either way and can easily change it once this gets reviewed 😅

brunocostalopes:
Oh, I missed that PR. So we now have the same feature implemented in #737 (my original PR) #765 (which includes those changes) and #829 (new PR). I wonder if any of these will ever get merged :(

seub:
@brunocostalopes Sorry I had missed your PR as well! It appears that there's a real need for these ignore/exclude columns! ;)

It would indeed help avoid a stack of redundant PRs if some of them finally got reviewed/approved @joellabes

macros/generic_tests/equality.sql
```diff
@@ -19,20 +23,108 @@
...
 
-{% set compare_cols_csv = compare_columns | join(', ') %}
+
+{% endif %}
+
+{%- if not precision -%}
+    {#-
+        If the compare_cols arg is provided, we can run this test without querying the
+        information schema — this allows the model to be an ephemeral model
+    -#}
```
gwenwindflower, line 76: I think the wording and placement of this comment and the one below it are a little confusing. It makes sense if you really walk through and think about it, and I agree it's helpful to have some comments here to clarify the conditions, but you might consider a second pass at these. Not a blocking problem in my mind but wanted to flag it as they threw me at first.

gwenwindflower (approved):
Looks good! Left one comment for consideration but otherwise this seems good to me and everything is passing now in CI. Thanks again for your patience on this!

macros/generic_tests/equality.sql
rlh1994: Thanks, have tried to provide some better comments!

gwenwindflower (approved):

gwenwindflower (approved):

gwenwindflower:
Love where you landed with the new comments @rlh1994 ! Thanks again for the patience, this is a fantastic addition to the package, really appreciate the work on this one. 💜

jomccr:
This has already helped me a ton, thanks for this contribution!

PR #771, opened 2023-03-16
Update ci command

joellabes:
trying to fix up the weird CI errors we're seeing


joellabes:
An update on my dbt_utils CI woes: 
- some of them were caused by race conditions in the `dbt build` world, because the test was running before the relations get_relations_by_pattern goes looking for existed. I fixed them by adding explicit `depends_on`s. 
- Redshift no longer cares for the `%` sign when getting a modulus. The good news is that it now accepts `MOD` which is the default; I've deleted the redshift-specific implementation
- I updated the config to use a large runner 
- I deleted the unnecessary seed configs in dbt_project.yml for now-removed cross-db macros. 
- The get_column_values_use_default test passes when I run it locally, but fails in CI. I have disabled it for the moment so that everything else passes and I can merge it; if anyone has any ideas as to why it's failing then I would love to hear them.

run_test.sh
```diff
@@ -14,15 +14,5 @@ export DBT_PROFILES_DIR=.
...
-if [[ ! -z $3 ]]; then _seeds="--select $3 --full-refresh"; fi
-
 dbt deps --target $1 || exit 1
-dbt seed --target $1 $_seeds || exit 1
-if [ $1 == 'redshift' ]; then
-    dbt run -x -m test_insert_by_period --full-refresh --target redshift || exit 1
-fi
-dbt run -x --target $1 $_models || exit 1
-dbt test -x --target $1 $_models || exit 1
+dbt build --target $1 --full-refresh || exit 1
```
dbeatty10, line 18: Suggestion below to add newline to remove the red circle icon at the bottom of the file:

image

```suggestion
dbt build --target $1 --full-refresh || exit 1

```

dbeatty10 (changes requested):
Overall, this looks good and glad CI can get to green!

Due to the change to `dbt build`, you'll need to update two other places to keep everything aligned (since `$(models)` and `$(seeds)` parameters to `run_test.sh` are no longer functional):
https://github.com/dbt-labs/dbt-utils/blob/9507d6bcd6c04b73e174a9aaee11865f9387e096/Makefile#L5

https://github.com/dbt-labs/dbt-utils/blob/9507d6bcd6c04b73e174a9aaee11865f9387e096/integration_tests/README.md?plain=1#L75-L90

Also, would be good to create an issue to come back and enable that flakey test (if you haven't already).

joellabes:
Changes made and issue opened: https://github.com/dbt-labs/dbt-utils/issues/788

integration_tests/tests/sql/test_get_column_values_use_default.sql
```diff
@@ -1,4 +1,7 @@
 
+{# This keeps succeeding locally and failing in CI. Disabling it to get everything else out, but it should still be tested. #}
```
dbeatty10, line 2: Let's add a link to the issue in GitHub so that anyone reading will know that there's an issue to address this item of tech debt.

```suggestion
{# This keeps succeeding locally and failing in CI. Disabling it to get everything else out, but it should still be tested.
https://github.com/dbt-labs/dbt-utils/issues/788 #}
```

dbeatty10 (approved):
One more code suggestion to accept as-is or format as desired.

Approving so you can immediately merge after that is applied.

PR #776, opened 2023-03-29
Improve the performance of the at_least_one test

### LINKED ISSUE #775: Improve the performance of the at_least_one test by pruning early.
### Describe the feature
I would like the `at_least_one` test to return values faster. In some applications it is very slow.

Not using the test.

### Additional context
I believe that the new feature can be added to the SQL code in a way that is compatible with any standard SQL database.

### Who will this benefit?
Any user with large tables or external (slow) data sources or tables.

Yes, I've already done the update and will file a PR shortly.

JoshuaHuntley:
resolves #775 

[This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
](https://github.com/JoshuaHuntley)

### DIFF
--- a/macros/generic_tests/at_least_one.sql
+++ b/macros/generic_tests/at_least_one.sql
@@ -4,9 +4,17 @@
 
 {% macro default__test_at_least_one(model, column_name, group_by_columns) %}
 
-{% if group_by_columns|length() > 0 %}
-  {% set select_gb_cols = group_by_columns|join(' ,') + ', ' %}
-  {% set groupby_gb_cols = 'group by ' + group_by_columns|join(',') %}
+{#- Ensure that column_name is not in the group by list or the database will throw an ambiguous column error #}
+{%- set group_by_cols_filtered = [] %}
+{% for column in group_by_columns %}
+  {%- if column != column_name %}
+      {%- do group_by_cols_filtered.append(column) %}
+  {% endif -%}
+{% endfor %}
+
+{% if group_by_cols_filtered|length() > 0 %}
+  {% set select_gb_cols = group_by_cols_filtered|join(' ,') + ', ' %}
+  {% set groupby_gb_cols = 'group by ' + group_by_cols_filtered|join(',') %}
 {% endif %}
 
 select *
--- a/macros/generic_tests/at_least_one.sql
+++ b/macros/generic_tests/at_least_one.sql
@@ -17,7 +17,14 @@ from (
       {{select_gb_cols}}
       count({{ column_name }}) as filler_column
 
-    from {{ model }}
+    from (
+          select
+            {{select_gb_cols}}
+            {{ column_name }}
+          from {{ model }}
+          where {{ column_name }} is not null
+          limit 1
+          )
 
     {{groupby_gb_cols}}
 
--- a/macros/generic_tests/at_least_one.sql
+++ b/macros/generic_tests/at_least_one.sql
@@ -4,17 +4,27 @@
 
 {% macro default__test_at_least_one(model, column_name, group_by_columns) %}
 
+{% set pruned_cols = [column_name] %}
+
 {% if group_by_columns|length() > 0 %}
+
   {% set select_gb_cols = group_by_columns|join(' ,') + ', ' %}
   {% set groupby_gb_cols = 'group by ' + group_by_columns|join(',') %}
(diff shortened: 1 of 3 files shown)


JoshuaHuntley:
The failing Redshift test is unrelated to these changes. The at_least_one tests in Redshift succeed.

macros/generic_tests/at_least_one.sql
```diff
@@ -4,9 +4,17 @@
...
-{% if group_by_columns|length() > 0 %}
-  {% set select_gb_cols = group_by_columns|join(' ,') + ', ' %}
-  {% set groupby_gb_cols = 'group by ' + group_by_columns|join(',') %}
+{#- Ensure that column_name is not in the group by list or the database will throw an ambiguous column error #}
+{%- set group_by_cols_filtered = [] %}
+{% for column in group_by_columns %}
+  {%- if column|upper != column_name|upper %}
+      {%- do group_by_cols_filtered.append(column) %}
+  {% endif -%}
+{% endfor %}
```
joellabes, line 13: I don't think this is necessary - including the name of the column you're testing in your group by columns as well is a misconfiguration that we shouldn't allow. 

Arguably, it's nice that we fix it for you instead of penalising you, but if we do that then we should make the change across all of the macros that have this behaviour and it should be in a different PR.

macros/generic_tests/at_least_one.sql
```diff
@@ -17,7 +25,14 @@ from (
...
 
-    from {{ model }}
+    from (
+          select
+            {{select_gb_cols}}
+            {{ column_name }}
+          from {{ model }}
+          where {{ column_name }} is not null
+          limit 1
+          ) pruned_rows
```
joellabes, line 35: A third layer of subquery feels pretty ick to me as an adherent to [team CTE](https://discourse.getdbt.com/t/why-the-fishtown-sql-style-guide-uses-so-many-ctes/1091) - could this be pulled out to a `pruned_rows` CTE without impacting performance? If so then that’s my preference.

joellabes (changes requested):
This makes a ton of sense! Thanks @JoshuaHuntley 🎉 

A couple of suggested changes - one that is a deferral to a different changeset (if at all... I'm not totally sold yet) and one change from a subquery to a CTE if possible.

JoshuaHuntley:
@joellabes I made the requested changes. FYI, I'm firmly in favor of CTEs but stuck to the original approach as much as possible on the first go-round. I've flattened the code out a bit so there is now only one layer of nesting. I see your position about the all-or-nothing approach to error handling. Pushing it down to the user, especially considering our intended user base, is fair.

joellabes:
Thanks @JoshuaHuntley! From the CI logs, it looks like you might have to get rid of the test that checks for the tested column being in the group_by columns as well.

JoshuaHuntley:
@joellabes I removed the test. There are Redshift adapter tests that are not related to this PR. Should I remove them or can we ignore them?

dbeatty10:
@JoshuaHuntley I'm @joellabes's colleague and I was asked to lend a hand on this review.

I pushed a couple commits to do the following:
- restore the group by test and fix the underlying issue that was causing it to fail
- update the changelog

The tests are all passing now ✅, so I think we'll be ready to merge soon.

### Details

#### Case-sensitive group by test
The group by test was failing when the column to test also was included in the list of columns to group by. So I added some **case-sensitive** logic that checks whether it is in that list or not.

Note that other macros like `union`, `unpivot`, etc take additional steps to make this comparison **case-insensitive**.

So users might still get an error like `column reference "field" is ambiguous` if they do something like this:

```yaml
  - name: data_test_at_least_one
    columns:
      - name: field
        tests:
          - dbt_utils.at_least_one
          - dbt_utils.at_least_one:
              group_by_columns: ['FIELD']
```

Currently, I'm more comfortable with that possibility than attempting to handle all the details and complications of a case-insensitive check. We can always update the implementation later if case-insensitive comparisons are needed.

#### `limit` in subqueries

I seem to recall that not all databases support `limit` within subqueries. Maybe MariaDB and MySQL fit in this category? For any databases in which this is the case, they will need to override this implementation in order to work.

dbeatty10 (approved):

Issue #713, opened 2022-10-25
`dbt_utils.deduplicate` macro returns zero rows when deduplicating a CTE with NULL column (Redshift)


yauhen-sobaleu:
### Describe the bug

`dbt_utils.deduplicate` returns zero rows when it deduplicates a CTE with a column that consists entirely of NULL values.

### Steps to reproduce
Given model code as:

```sql
with data as (

    select 
        '2021-04-03 23:00:26'::timestamp as ts,
        1 AS id,
        NULL AS null_field
    UNION ALL
    select 
        '2021-04-03 23:00:26'::timestamp as ts,
        1 AS id,
        NULL AS null_field
),

data_deduped AS (
    {{ dbt_utils.deduplicate(
        relation='data',
        partition_by='id',
        order_by='ts desc',
    ) }}
)

select * from data_deduped
```

it is compiled by dbt on Redshift adapter into the following SQL:

```sql
with data as (

    select 
        '2021-04-03 23:00:26'::timestamp as ts,
        1 AS id,
        NULL AS null_field
    UNION ALL
    select 
        '2021-04-03 23:00:26'::timestamp as ts,
        1 AS id,
        NULL AS null_field
),

data_deduped AS (
    with row_numbered as (
        select
            _inner.*,
            row_number() over (
                partition by id
                order by ts desc
            ) as rn
        from data as _inner
    )

    select
        distinct data.*
    from data as data
    
    natural join row_numbered
    where row_numbered.rn = 1
)

select * from data_deduped
```

### Expected results
```
         ts          | id | null_field 
---------------------+----+------------
 2021-04-03 23:00:26 |  1 |          
(1 row)
```

### Actual results

```
 ts | id | null_field 
----+----+------------
(0 rows)
```

### System information
**The contents of your `packages.yml` file:**
```yaml
packages:
  - package: mjirv/dbt_datamocktool
    version: 0.2.0
```

**Which database are you using dbt with?**
- [ ] postgres
- [x] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
Core:
  - installed: 1.2.1
  - latest:    1.3.0 - Update available!

  Your version of dbt-core is out of date!
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation

Plugins:
  - redshift: 1.2.1 - Update available!
  - postgres: 1.2.1 - Update available!

[...]

Issue #621, opened 2022-07-16
Default implementation of `deduplicate` is not null-safe


dbeatty10:
### Describe the bug
@belasobral93 discovered that `deduplicate` doesn't work for Spark when any of the `order_by` columns are `null`.

#### Root cause
The root cause is that [Spark defaults to `NULLS FIRST` for the null_sort_order](https://spark.apache.org/docs/latest/sql-ref-syntax-qry-select-orderby.html#parameters) for the `ORDER BY` clause.

#### Why `dbt_utils` rather than `spark_utils`?

<details>
  <summary>Explanation</summary>

> But `dbt_utils` doesn't officially support Spark, so why is this being reported here (instead of `spark_utils`)?

`dbt_utils` is providing [five implementations for `deduplicate`](https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/deduplicate.sql):
1. default
2. redshift
3. postgres
4. snowflake
5. bigquery

Since `dbt_utils` only tests against the four databases listed above and each of those has an override, it means there is **no testing** for the `default` implementation which the other dbt adapters inherit (like Spark).

</details>

### Steps to reproduce
<details>
  <summary>SQL example</summary>

The current implementation essentially acts like the following example:
```sql
with relation as (

    select 1 as partition_by, 'a' as order_by
    union all
    select 1 as partition_by, 'a' as order_by
    union all
    select 2 as partition_by, 'a' as order_by
    union all
    select 2 as partition_by, 'b' as order_by
    union all
    select 2 as partition_by, NULL as order_by

),

row_numbered as (

    select
        _inner.*,
        row_number() over (
            partition by partition_by
            order by order_by
        ) as rn
    from relation as _inner

)

select
    distinct relation.*
from relation
natural join row_numbered
where row_numbered.rn = 1
;
```

</details>

### Expected results
| partition_by | order_by |
|--------------|----------|
| 1            | a        |
| 2            | a        |

### Actual results
| partition_by | order_by |
|--------------|----------|
| 1            | a        |

### Screenshots and log output
Not provided.

### System information
Not provided.

**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [x] other: spark

**The output of `dbt --version`:**
Not provided.

### Additional context
To make the fix, it might be as simple as:
- adding `nulls last` after `order_by` (when an `order by` clause exists, of course)

[...]

PR #769, opened 2023-02-15
Add Databricks handler for get_table_types_sql.sql

Harmuth94:
resolves [#768](https://github.com/dbt-labs/dbt-utils/issues/768)

This is a:
- [ ] documentation update
- [x] bug fix with no breaking changes
- [ ] new functionality
- [ ] a breaking change

All pull requests from community contributors should target the `main` branch (default).

## Description & motivation
`get_relations_by_pattern` throws an error when using the databricks adapter on schemas in a Unity catalog. The error comes from the table having the `table_type` `MANAGED` in the `information_schema.tables` table.
Adding a handler for databricks allows the macro to as expected with data in a Unity catalog.

## Checklist
- [ ] This code is associated with an Issue which has been triaged and [accepted for development](https://docs.getdbt.com/docs/contributing/oss-expectations#pull-requests). 
- [x] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [x] Databricks
- [x] I followed guidelines to ensure that my changes will work on "non-core" adapters by:
    - [x] dispatching any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/star.sql))
    - [ ] using the `limit_zero()` macro in place of the literal string: `limit 0`
    - [ ] using `dbt.type_*` macros instead of explicit datatypes (e.g. `dbt.type_timestamp()` instead of `TIMESTAMP`
- [ ] I have updated the README.md (if applicable)
- [ ] I have added tests & descriptions to my models (and macros if applicable)
- [x] I have added an entry to CHANGELOG.md

PR #786, opened 2023-04-26
`deduplicate` macro for Databricks now uses the `QUALIFY` clause, which fixes `NULL` columns issues from the default natural join logic

graciegoheen:
Related to issues described in #713 and #621 

It looks like we don't have circle-ci running integration tests for Databricks - should we add that? I can alternatively add this to `spark_utils`. Additionally, I'd like to add some additional integration tests to confirm this macro _will_ work for null values, but I think that we would need to update the default version of this macro and remove the natural join entirely.

This is a:
- [ ] documentation update
- [ ] new functionality
- [ ] a breaking change

## Description & motivation

In databricks, natural join was causing issues for a customer - we can use the new QUALIFY in databricks to update the deduplicate macro.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,6 +83,20 @@ jobs:
     # The resource_class feature allows configuring CPU and RAM resources for each job. Different resource classes are available for different executors. https://circleci.com/docs/2.0/configuration-reference/#resourceclass    
     resource_class: large
 
+  integration-databricks:
+    docker:
+      - image: cimg/python:3.9
+    steps:
+      - checkout
+      - run: pip install --pre dbt-databricks -r dev-requirements.txt
+      - run:
+          name: "Run OG Tests - Databricks"
+          command: ./run_test.sh databricks
+      - store_artifacts:
+          path: integration_tests/logs
+      - store_artifacts:
+          path: integration_tests/target
+
 workflows:
   version: 2
   test-all:
@@ -98,5 +112,9 @@ workflows:
             - integration-postgres
       - integration-bigquery:
           context: profile-bigquery
+          requires:
+            - integration-postgres
+      - integration-databricks:
+          context: profile-databricks
           requires:
             - integration-postgres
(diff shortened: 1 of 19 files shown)


NortySpock:
@graciegoheen  Thank you very much for creating this pull request. Dropping this macro into my project's macro folder saved me at least a half-day of "fix it" work and frantic testing after I stumbled into https://github.com/dbt-labs/dbt-utils/issues/713 

To the maintainers: This worked for me as a one-off solution in a Databricks-powered project at work, and while I have not exhaustively tested it, it definitely fixed the _"I am missing most of the data for some strange reason"_ data problems that I was seeing post-deduplication from  https://github.com/dbt-labs/dbt-utils/issues/713

dbeatty10 (approved):
@graciegoheen I'm going to merge this as-is. Rationale below.

---

> It looks like we don't have circle-ci running integration tests for Databricks - should we add that?

I made an initial attempt at adding Databricks to CI, but it didn't work. The cause appears related to the integration tests using pre-releases of dbt-core (1.6.0bx) that are somehow incompatible with the latest available version of dbt-databricks (1.5.x).

So I'm going to defer that decision to a later date.

> I can alternatively add this to spark_utils.

Since we already have `databricks__get_table_types_sql` (https://github.com/dbt-labs/dbt-utils/pull/769), it seems reasonable to add `databricks__deduplicate` here also (rather than putting it in spark_utils).

> Additionally, I'd like to add some additional integration tests to confirm this macro will work for null values, but I think that we would need to update the default version of this macro and remove the natural join entirely.

There's a simple test case in https://github.com/dbt-labs/dbt-utils/issues/713 that would work great. But if we add it to CI without changing the default implementation, then Redshift will start failing CI.

Since removing the natural join will take more time and thought, I'm going to defer adding new integration tests to a later date as well.

PR #512, opened 2022-03-06
Add deduplication macro

judahrand:
This is a:
- [ ] bug fix PR with no breaking changes — please ensure the base branch is `main`
- [x] new functionality — please ensure the base branch is the latest `dev/` branch
- [ ] a breaking change — please ensure the base branch is the latest `dev/` branch

## Description & motivation

This PR aims to fix #335.

## Checklist
- [x] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [x] BigQuery
    - [x] Postgres
    - [x] Redshift
    - [x] Snowflake
- [x] I followed guidelines to ensure that my changes will work on "non-core" adapters by:
    - [x] dispatching any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/star.sql))
    - [x] using the `limit_zero()` macro in place of the literal string: `limit 0`
    - [x] using `dbt_utils.type_*` macros instead of explicit datatypes (e.g. `dbt_utils.type_timestamp()` instead of `TIMESTAMP`
- [x] I have updated the README.md (if applicable)
- [x] I have added tests & descriptions to my models (and macros if applicable)
- [x] I have added an entry to CHANGELOG.md

Issue #335, opened 2021-02-12
Deduplication macro


blake-enyart:
### Describe the feature
It would be great to have deduplication macro which utilizes something such as the surrogate_key macro to deduplicate on. The basis for this would be something such as: https://github.com/fishtown-analytics/dbt-utils/pull/29

### Describe alternatives you've considered
At this point, the only alternative is writing this in full SQL query and this feels like a pretty common repetitive type of query that is prime for integrating into a macro

### Additional context
Is this feature database-specific? Which database(s) is/are relevant? Please include any other relevant context here.
I work mainly in Snowflake, but this could be generalized. Perhaps a first draft in Snowflake and expands outward?

### Who will this benefit?
It is very common practice to deduplicate data on ingestion in addition to type casting which will greatly expedite new data source ingestion

### Are you interested in contributing this feature?

I would need a hand to get started on this. I can leverage this PR as a reference, https://github.com/fishtown-analytics/dbt-utils/pull/29, but would greatly appreciate any support and perspective provided here.

rwatts3:
Here is a macro I use to dedupe using BigQuery, not sure if its relevant or could help. 

```
{%- macro deduped_dataset(database, table, key="id") -%}
  select deduped.*
    from (
      select
        _inner.*,
        row_number() over (partition by {{ key }} order by updated_at desc ) as rn
      from {{ source(database, table) }} as _inner
    ) as deduped
  where deduped.rn = 1
{%- endmacro -%}
```

switzer:
We've had resource issues using `row_number()` vs. using `array_agg`.  Our dedupe macro is as follows:

```
{% macro dedupe_source(source, group_by, order_by = None) %}
  select
    event.*
  from (
    select
      array_agg (
        {{ source }}
        {% if order_by != None %}
        order by {{ order_by }}
        {% endif %}
        limit 1
      )[offset(0)] event
    from {{ source }}
    group by {{ group_by }}
  )
{% endmacro %}
```

It is more performant (can handle more rows, not necessarily faster).  You may want to consider going down this path.

blake-enyart:
Hi @switzer and @rwatts3 !

Thank you both for sharing out your dedupe methodologies. I'm working on standing up some client projects right now and hoping to utilize these shortly :)

TinKlo:
Hi @blake-enyart , @switzer and @rwatts3 !

I was trying to apply this method but to many columns (4). but I got some errors. 

[...]

PR #815, opened 2023-07-26
Methods to achieve null safety for `deduplicate`

### LINKED ISSUE #621: Default implementation of `deduplicate` is not null-safe

### LINKED ISSUE #814: deduplicate bug in Spark in case of a null column.
### Describe the bug
Same bug as https://github.com/dbt-labs/dbt-utils/issues/713 however for Spark
When there is a null column(Unrelated to the partition by or order by columns), Spark doesn't return the expected rows in the deduplicate default function.

### Steps to reproduce
*** This is python code where the deduplication code was copied to in order to reproduce the bug.
```
  sql = """
  with relation as (
  select * from ( values (1, 3, null), (1, 2, null), (1, 1, null)) 
  ), row_numbered as (
      select
          _inner.*,
          row_number() over (
              partition by col1
              order by col2
          ) as rn
      from relation as _inner
  )

  select
      distinct data.*
  from relation as data
  natural join row_numbered
  where row_numbered.rn = 1
  """
  df = session.sql(sql)
  df.show()
```
### Expected results
+----+----+----+
|col1|col2|col3|
+----+----+----+
|       1|       1| null |
+----+----+----+

### Actual results
+----+----+----+
|col1|col2|col3|
+----+----+----+
+----+----+----+

### System information

- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [x] other (specify: Spark)

```
Core:
  - installed: 1.6.0-b8
  - latest:    1.5.3    - Ahead of latest version!

Plugins:
```

dbeatty10:
resolves #814
resolves #621

This is a bug fix with no breaking changes.

It also adds two new features:
- optional `row_alias` keyword argument (type: string, default: `none`)
- optional `columns` keyword argument (type: list, default: `none`)

## Description & motivation
This PR is still in draft status, and more description will be added at a later date.

In the meantime, see #814 (and everything it links to, in particular #713) for background motivation and discussion to-date.

As a summary, this PR gives the user multiple options to achieve null safety for `deduplicate`:

1. The user passes a relation that has been materialized in the database (i.e., it isn't an ephemeral model)
    - https://github.com/dbt-labs/dbt-utils/pull/815/commits/fe03f43cc5f57d86006addfa1770ab78230cd725 -- as long as the `relation` is not a CTE, it's columns can be fetched via the `get_filtered_columns_in_relation` macro
1. The user passes the `row_alias` keyword argument
    -  https://github.com/dbt-labs/dbt-utils/pull/815/commits/3eced4d100f8e7fe91827678bd321c242718ed94 -- when the `row_alias` keyword argument is set, then we can deduplicate via the `row_number()` window function (at the cost of the `row_alias` being an extra column that wasn't in the original data set)
1. The user passes the `columns` keyword argument
    -  https://github.com/dbt-labs/dbt-utils/pull/815/commits/d46676e98bd3e0c5f3d4e5056ff83c765720a3a0 -- when `columns` keyword argument is set, then we can deduplicate via the `row_number()` window function and only return the requested columns

Outside of those options, the deduplication will not be null-safe.

### Option 1

`models/my_model_1.sql`
```
{{ config(materialized="table") }}

select 1 as user_id, cast(null as date) as created_at, 1 as version union all
select 1 as user_id, cast(null as date) as created_at, 2 as version union all
select 1 as user_id, cast(null as date) as created_at, 2 as version
```

`models/deduped_1.sql`

```sql
    {{
        dbt_utils.deduplicate(
            ref('my_model_1'),
            partition_by='user_id, created_at',
            order_by='version desc'
        ) | indent
    }}
```

```shell
dbt build -s +deduped_1
dbt show -s deduped_1
```

| user_id | created_at | version |
| ------- | ---------- | ------- |
|       1 |            |       2 |

### Option 2

`models/my_model_2.sql`
```
{{ config(materialized="ephemeral") }}

select 1 as user_id, cast(null as date) as created_at, 1 as version union all
select 1 as user_id, cast(null as date) as created_at, 2 as version union all
select 1 as user_id, cast(null as date) as created_at, 2 as version
```

`models/deduped_2.sql`

```sql
    {{
        dbt_utils.deduplicate(
            ref('my_model_2'),
            partition_by='user_id, created_at',
            order_by='version desc',
            row_alias='rn'
        ) | indent
    }}
```

```shell
dbt build -s +deduped_2
dbt show -s deduped_2
```

| user_id | created_at | version | rn |
| ------- | ---------- | ------- | -- |
|       1 |            |       2 |  1 |

### Option 3

`models/my_model_3.sql`
```
{{ config(materialized="ephemeral") }}

select 1 as user_id, cast(null as date) as created_at, 1 as version union all
select 1 as user_id, cast(null as date) as created_at, 2 as version union all
select 1 as user_id, cast(null as date) as created_at, 2 as version
```

`models/deduped_3.sql`

```sql
    {{
        dbt_utils.deduplicate(
            ref('my_model_3'),
            partition_by='user_id, created_at',
            order_by='version desc',
            columns=['user_id', 'created_at', 'version']
        ) | indent
    }}
```

```shell
dbt build -s +deduped_3
dbt show -s deduped_3
```

| user_id | created_at | version |
| ------- | ---------- | ------- |
|       1 |            |       2 |

### Option 4

`models/my_model_4.sql`
```
{{ config(materialized="ephemeral") }}

select 1 as user_id, cast(null as date) as created_at, 1 as version union all
select 1 as user_id, cast(null as date) as created_at, 2 as version union all
select 1 as user_id, cast(null as date) as created_at, 2 as version
```

> [!WARNING]  
> This is the one not guaranteed to be null-safe (depending on the adapter).

`models/deduped_4.sql`

```sql
    {{
        dbt_utils.deduplicate(
            ref('my_model_4'),
            partition_by='user_id, created_at',
            order_by='version desc',
        ) | indent
    }}
```

```shell
dbt build -s +deduped_4
dbt show -s deduped_4
```

Here's the warning that will be logged:

```
Warning: the implementation of the `deduplicate` macro for the `postgres` adapter is not null safe. 
Set `row_alias` within calls to `deduplicate` to achieve null safety (which will also add it as an extra column to the output).

e.g.,
    {
        dbt_utils.deduplicate(
            'my_cte',
            partition_by='user_id',
            order_by='version desc',
            row_alias='rn'
        ) | indent
    }

Warning triggered by model: my_project.deduped_4
dbt project / package: my_project
path: models/deduped_4.sql
```

| user_id | created_at | version |
| ------- | ---------- | ------- |

### Key history of `deduplicate` macro
- https://github.com/dbt-labs/dbt-utils/issues/335
- https://github.com/dbt-labs/dbt-utils/pull/512
- https://github.com/dbt-labs/dbt-utils/pull/526
- https://github.com/dbt-labs/dbt-utils/pull/548

## Checklist
- [x] I followed guidelines to ensure that my changes will work on "non-core" adapters
- [ ] 👈 I have updated the README.md (if applicable)
- [ ] 👈 I have added tests & descriptions to my models (and macros if applicable)
- [ ] 👈 I have added an entry to CHANGELOG.md

### DIFF
--- a/macros/sql/deduplicate.sql
+++ b/macros/sql/deduplicate.sql
@@ -1,8 +1,40 @@
 {%- macro deduplicate(relation, partition_by, order_by) -%}
-    {{ return(adapter.dispatch('deduplicate', 'dbt_utils')(relation, partition_by, order_by)) }}
+    {{ return(adapter.dispatch('deduplicate', 'dbt_utils')(relation, partition_by, order_by, **kwargs)) }}
 {% endmacro %}
 
-{%- macro default__deduplicate(relation, partition_by, order_by) -%}
+{#
+-- ⚠️ This macro drops rows that contain NULL values ⚠️
+
+-- The implementation below uses a natural join which avoids returning an
+-- extra column at the cost of not being null safe.
+
+-- dbt_utils._safe_deduplicate is an alternative that avoids dropping rows
+-- that contain NULL values at the cost of adding an extra column.
+#}
+{%- macro _unsafe_deduplicate(relation, partition_by, order_by) -%}
+
+{%- set error_message = "
+Warning: the implementation of the `deduplicate` macro for the `{}` adapter is not null safe. \
+
+Set `row_alias` within calls to `deduplicate` to achieve null safety (which will also add it \
+as an extra column to the output).
+
+e.g.,
+    {{
+        dbt_utils.deduplicate(
+            'my_cte',
+            partition_by='user_id',
+            order_by='version desc',
+            row_alias='rn'
+        ) | indent
+    }}
+
+Warning triggered by model: {}.{}
+dbt project / package: {}
+path: {}
+".format(target.type, model.package_name, model.name, model.package_name, model.original_file_path) -%}
+
+{%- do exceptions.warn(error_message) -%}
 
     with row_numbered as (
         select
@@ -26,6 +58,44 @@
     #}
     natural join row_numbered
     where row_numbered.rn = 1
+
+{%- endmacro -%}
+
+{#
+-- For data platforms that don't support QUALIFY or an equivalent, the
+-- best we can do to ensure null safety is to use a window function +
+-- filter (which returns an extra column):
+-- https://modern-sql.com/caniuse/qualify
+#}
+{%- macro _safe_deduplicate(relation, partition_by, order_by, row_alias="rn") -%}
+
+    with row_numbered as (
+        select
+            _inner.*,
+            row_number() over (
+                partition by {{ partition_by }}
+                order by {{ order_by }}
+            ) as {{ row_alias }}
+        from {{ relation }} as _inner
+    )
+
+    select *
+    from row_numbered
+    where {{ row_alias }} = 1
+
+{%- endmacro -%}
+
+{#
+-- ⚠️ This macro drops rows that contain NULL values unless the `row_alias` parameter is included ⚠️
+#}
+{%- macro default__deduplicate(relation, partition_by, order_by) -%}
+    {% set row_alias = kwargs.get('row_alias') %}
+
+    {% if row_alias != None %}
+        {{ dbt_utils._safe_deduplicate(relation, partition_by, order_by, row_alias=row_alias) }}
+    {% else %}
+        {{ dbt_utils._unsafe_deduplicate(relation, partition_by, order_by) }}
+    {% endif %}
 
 {%- endmacro -%}
 
--- a/macros/sql/deduplicate.sql
+++ b/macros/sql/deduplicate.sql
@@ -67,11 +67,23 @@ path: {}
 -- filter (which returns an extra column):
 -- https://modern-sql.com/caniuse/qualify
 #}
-{%- macro _safe_deduplicate(relation, partition_by, order_by, row_alias="rn") -%}
+{%- macro _safe_deduplicate(relation, partition_by, order_by, row_alias="rn", columns=none) -%}
+
+    {% if not row_alias %}
+        {% set row_alias = "rn" %}
+    {% endif %}
 
     with row_numbered as (
         select
+
+        {% if columns != None %}
+            {% for column in columns %}
+            {{ column }},
+            {% endfor %}
+        {% else %}
             _inner.*,
+        {% endif %}
+
             row_number() over (
                 partition by {{ partition_by }}
                 order by {{ order_by }}
@@ -90,9 +102,10 @@ path: {}
 #}
 {%- macro default__deduplicate(relation, partition_by, order_by) -%}
     {% set row_alias = kwargs.get('row_alias') %}
+    {% set columns = kwargs.get('columns') %}
 
-    {% if row_alias != None %}
-        {{ dbt_utils._safe_deduplicate(relation, partition_by, order_by, row_alias=row_alias) }}
+    {% if row_alias != None or columns != None %}
+        {{ dbt_utils._safe_deduplicate(relation, partition_by, order_by, row_alias=row_alias, columns=columns) }}
     {% else %}
         {{ dbt_utils._unsafe_deduplicate(relation, partition_by, order_by) }}
     {% endif %}
--- a/macros/sql/deduplicate.sql
+++ b/macros/sql/deduplicate.sql
@@ -98,7 +98,10 @@ path: {}
 {%- endmacro -%}
 
 {#
--- ⚠️ This macro drops rows that contain NULL values unless the `row_alias` parameter is included ⚠️
+-- ⚠️ This macro drops rows that contain NULL values unless one of the following is true:
+--   - `relation` parameter is a non-CTE dbt Relation
+--   - `row_alias` parameter is included
+--   - `columns` parameter is included
 #}
 {%- macro default__deduplicate(relation, partition_by, order_by) -%}
     {% set row_alias = kwargs.get('row_alias') %}
--- a/macros/sql/deduplicate.sql
+++ b/macros/sql/deduplicate.sql
@@ -104,7 +104,10 @@ path: {}
     {% set row_alias = kwargs.get('row_alias') %}
     {% set columns = kwargs.get('columns') %}
 
-    {% if row_alias != None or columns != None %}
+    {% if relation.is_cte is defined and not relation.is_cte %}
+        {% set columns = dbt_utils.get_filtered_columns_in_relation(relation) %}
+        {{ dbt_utils._safe_deduplicate(relation, partition_by, order_by, columns=columns) }}
+    {% elif row_alias != None or columns != None %}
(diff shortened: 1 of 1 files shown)


macros/sql/deduplicate.sql
```diff
@@ -29,6 +61,44 @@
...
+        select
+            _inner.*,
+            row_number() over (
+                partition by {{ partition_by }}
+                order by {{ order_by }}
+            ) as {{ row_alias }}
+        from {{ relation }} as _inner
+    )
+
+    select *
```
graciegoheen, line 82: What databases allow for `minus` or `except` syntax? I know [snowflake](https://docs.snowflake.com/en/sql-reference/operators-query#minus-except) does - that could be an option for removing the extra column. Though maybe in that case you'd just use `qualify`

dbeatty10: How would [minus or except](https://docs.snowflake.com/en/sql-reference/operators-query#minus-except) work to remove extra column(s)? Do you mean `select * exclude ( <col_name>, <col_name>, ... )`?

This would be the perfect solution if we could rely on it! 💡 

But it is not in the [SQL standard](https://en.wikipedia.org/wiki/ISO/IEC_9075), and the databases that don't have `qualify` are probably missing `select * exclude (...)` as well. So I don't think we'll be able to reliably use it as part of the default implementation 😢.

---

#### `select * exclude (...)`

[Snowflake](https://docs.snowflake.com/en/sql-reference/sql/select#label-select-cmd-examples-select-all-in-table-but-two) has `select * exclude`:

image

And so does [DuckDB](https://duckdb.org/docs/sql/query_syntax/select.html):

image

#### `select * except (...)`

And because it's not in the standard, other databases use `except` instead of `exclude`.

[BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_except) uses `except`:

image

As does [Databricks](https://docs.databricks.com/sql/language-manual/sql-ref-syntax-qry-select.html#syntax):

image

graciegoheen: Sorry, yes I meant `exclude`. What about using the [star](https://github.com/dbt-labs/dbt-utils#star-source) macro with the `except` argument?

dbeatty10: The initial implementation in [#512](https://github.com/dbt-labs/dbt-utils/pull/512/files) used the `star` macro but it was removed in https://github.com/dbt-labs/dbt-utils/pull/548.

I haven't considered the details of how we might be able to bring it back or what those implications would be.

I think we'd still need to handle the case where the `relation` is a CTE name instead of a [Relation](https://docs.getdbt.com/reference/dbt-classes#relation). That's the case that this draft PR is covering with the `row_alias` parameter. An alternative way to cover it would be a `columns` parameter like suggested [here](https://github.com/dbt-labs/dbt-utils/issues/713#issuecomment-1341066314). Allowing the end user to choose between either `row_alias` or `columns` would provide the most optionality.

macros/sql/deduplicate.sql
```diff
@@ -104,7 +104,10 @@ path: {}
     {% set row_alias = kwargs.get('row_alias') %}
     {% set columns = kwargs.get('columns') %}
 
-    {% if row_alias != None or columns != None %}
+    {% if relation.is_cte is defined and not relation.is_cte %}
```
dbeatty10, line 110: This is a simplified alternative to [this](https://github.com/dbt-labs/dbt-utils/blob/a17d66f332a782a7e81a59f8d249237f3c29612c/macros/jinja_helpers/_is_relation.sql#L2) and [this](https://github.com/dbt-labs/dbt-utils/blob/a17d66f332a782a7e81a59f8d249237f3c29612c/macros/jinja_helpers/_is_ephemeral.sql#L2) to determine if a Relation is backed by a real database object whose columns can be fetched via the `star` macro.

i.e., the columns for tables and views can be retrieved via `information_schema.columns` (or an equivalent), but can't for CTE.

macros/sql/deduplicate.sql
dbeatty10: @graciegoheen your idea about using the [star](https://github.com/dbt-labs/dbt-utils#star-source) macro inspired https://github.com/dbt-labs/dbt-utils/pull/815/commits/fe03f43cc5f57d86006addfa1770ab78230cd725.

It retrieves columns similarly to `dbt_utils.star` IFF:
- relation is a Relation
- relation is not an ephemeral CTE

Otherwise, a user can pass a list of `columns` manually (d46676e98bd3e0c5f3d4e5056ff83c765720a3a0). Or they can specify a `row_alias` that is acceptable to them.

PR #829, opened 2023-08-30
Equality exclude columns

### LINKED ISSUE #828: Add optional `exclude_columns` to the `equality` test
### Describe the feature

It'd be nice to have an optional `exclude_columns` in the `equality` test, similar to `compare_columns`:

```yaml
    tests:
      - dbt_utils.equality:
          compare_model: ref('other_table_name')
          exclude_columns:
            - third_column
            - fourth_column
```

Such columns would be ignored when comparing the two tables.

1. Rewrite the test in our projects
2. Somehow leverage `compare_relations` from [dbt-audit-helper](https://github.com/dbt-labs/dbt-audit-helper/blob/main/macros/compare_relations.sql)

I don't think either solution is great compared to adding the feature here.

### Additional context / Who will this benefit?

As an example, here's why my team would like this. We leverage the `equality` test from dbt-utils to implement unit tests. However, we need to exclude the "processing timestamp" column, as it cannot be predicted for the "expected output" model.

Yep! In fact, I wrote the PR ([#829 ](https://github.com/dbt-labs/dbt-utils/pull/829)) before the issue, because I'd forgotten that all PRs must be associated with an issue, oops.

seub:
resolves #828 

This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation

This PR adds an optional `exclude_columns` to the `equality` test (similar to `compare_columns`).

## Checklist
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [x] Snowflake

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 --->
 
 # Unreleased
+
+## Enhancements
+- Added an optional `exclude_columns` to the `equality` test. ([#828](https://github.com/dbt-labs/dbt-utils/issues/828))
+
 ## Fixes
 - deduplicate macro for Databricks now uses the QUALIFY clause, which fixes NULL columns issues from the default natural join logic
 - deduplicate macro for Redshift now uses the QUALIFY clause, which fixes NULL columns issues from the default natural join logic
--- a/README.md
+++ b/README.md
@@ -146,6 +146,8 @@ With `exclude_columns`:
             - fourth_column
 ```
 
+**Note:** The exclude columns are case-insensitive.
+
 
 ### expression_is_true ([source](macros/generic_tests/expression_is_true.sql))
 
--- a/README.md
+++ b/README.md
@@ -114,15 +114,20 @@ This test supports the `group_by_columns` parameter; see [Grouping in tests](#gr
 
 ### equality ([source](macros/generic_tests/equality.sql))
 
-Asserts the equality of two relations. Optionally specify a subset of columns to compare.
+Asserts the equality of two relations. Optionally specify a subset of columns to compare or to exclude.
 
 **Usage:**
 
 ```yaml
-version: 2
-
 models:
   - name: model_name
+    tests:
+      - dbt_utils.equality:
+          compare_model: ref('other_table_name')
+```
+
+With `compare_columns`:
+```yaml
     tests:
       - dbt_utils.equality:
           compare_model: ref('other_table_name')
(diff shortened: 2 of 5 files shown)


seub:
Update: I figured out what the deal was with Snowflake and quoted columns. I fixed it with a new version that is a bit uglier but seems to work. However it's now failing for BigQuery apparently. Some help would be great 🙏

seub:
Update: I think I finally figured out how to handle quoting in a satisfactory way (tricky! But I learned something).

I think the PR is good to go! Thanks for reviewing 🙏

joellabes (changes requested):
At a glance this looks good! One note is that only one of `compare_columns` and `exclude_columns` should be accepted (see [union_relations](https://github.com/dbt-labs/dbt-utils#union_relations-source) for an example)

I'll try and find someone for a more robust review and getting this merged 🎉

seub:
Thank you @joellabes !

About forbidding to have both `compare_columns` and `exclude_columns`: I considered that; I also saw an example of that in the macro `get_merge_update_columns` in dbt-core [here](https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/include/global_project/macros/materializations/models/incremental/column_helpers.sql).

We could easily forbid it explicitly, but imho there's no need. The final set of columns that will be compared is the subset of `compare_columns` that are not in `exclude_columns`, that seems pretty natural.

rlh1994:
I think this is a duplicate of https://github.com/dbt-labs/dbt-utils/pull/765 (although that uses `ignore_columns`, no particular reason)?

seub:
@rlh1994 Sorry for the slow response!

Indeed it is duplicate, my bad for missing it. Do you know what's the hold up with your PR? Looks like it's been open for more than 6 months.

If I compare what I wrote to your PR, I only see 2 tiny differences:
- Terminology: I went for `exclude_columns` instead of your `ignore_columns`. Either is perfectly reasonable, I went for `exclude` to mimic the `merge_exclude_columns` in dbt-core.
- My PR allows to have both `compare_columns` and `exclude_columns`. I don't think it hurts to allow both, but it also makes sense to not allow it.

I tried to be very careful about quoting (depending on whether `compare_columns` is provided) and I think my PR works. I was pleased to see that your logic does exactly the same as mine. I use `get_quoted_csv` in one case and you don't but the result is the same.

I'm looking forward to your PR being merged and I can close this one. Unless you want to split yours to only add `precision`? I don't think it's worth it, unless something about that is holding up your PR from being approved?

rlh1994:
@seub I think `exclude_columns` makes more sense, I'll swap mine over to that. Mine also additionally checks if one of the tables is missing one of the columns (which was previously missed in one direction, I can't remember which).

I don't know of any reasons the PR isn't progressing tbh, I think it's ready to go but just awaiting a review

dbeatty10:
Thank you @seub and @rlh1994 !

Closing this in favor of https://github.com/dbt-labs/dbt-utils/pull/765

PR #849, opened 2023-10-18
Fix for #848 (not_constant fails within groups with 1 element)

### LINKED ISSUE #848: not_constant may generate false positive if used with group_by_column
### Describe the bug
not_constant generic test would generate a false positive if used with group_by_column and there is only one row within a group

### Steps to reproduce
Test model:

```sql
-- file test.sql
select 1 as a, 10 as b
union all
select 1 as a, 11 as b
union all
select 2 as a, 21 as b
```

Test schema.yml:
```yaml
version: 2
models:
  - name: test
    columns:
      - name: b
        tests:
          - dbt_utils.not_constant:
              group_by_columns: ['a']
```

Run: `dbt test -m test`

### Expected results
Test passes, because if there is only one value in the group, it does not break the constancy rule.

### Actual results
Test fails.

### Context
To explain what I'm doing and where this might be useful to others. I am flattening a nested array, and setting a flag if a given record is the first one. I wanted to add a test to make sure this logic always works, because I'm relying on this flag to be true exactly once for each id of the original record. In some cases there is only one element in the array, which gets the flag set to true correctly, but the test thinks it's wrong.

Not sure what else would help.

### System information
```yaml
packages:
  - package: dbt-labs/dbt_utils
    version: 1.1.1
```

Dremio

```
Core:
  - installed: 1.5.8
  - latest:    1.6.6 - Update available!

Plugins:
  - dremio: 1.5.0 - Up to date!
```

PR provided #849

igor-lobanov-maersk:
resolves #848

This is a:
- [ ] documentation update
- [ ] new functionality
- [ ] a breaking change

## Description & motivation
As described in the bug report for #848, not_constant generic test fails when grouping is used and one of the groups has only one element. The implementation considers the value of this one element constant and fails the test. Whilst such interpretation is technically correct, the context where such test is used a different interpretation arguably would be preferable. I provided additional details in the bug report.

## Checklist
- [ ] This code is associated with an Issue which has been triaged and [accepted for development](https://docs.getdbt.com/docs/contributing/oss-expectations#pull-requests). **Not really, I just filed a bug report, but found it's trivial to fix**
- [ ] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered) **I only have access to Dremio environment, and I tested it there**
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake
    - [ ] dispatching any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/star.sql)) **N/A**
    - [ ] using the `limit_zero()` macro in place of the literal string: `limit 0` **N/A**
    - [ ] using `dbt.type_*` macros instead of explicit datatypes (e.g. `dbt.type_timestamp()` instead of `TIMESTAMP` **N/A**
- [ ] I have updated the README.md (if applicable) **N/A**
- [ ] I have added tests & descriptions to my models (and macros if applicable) **No, see below**

I did not add a test, because it's not clear to me how such tests are written. There is no prior test for not_constant macro to extend. That said, I'm happy to write one with some guidance from the core team.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 * Improve the performance of the `at_least_one` test by pruning early. This is especially helpful when running against external tables. By @joshuahuntley in https://github.com/dbt-labs/dbt-utils/pull/775
 ## Fixes
 * Fix legacy links in README by @dbeatty10 in https://github.com/dbt-labs/dbt-utils/pull/796
+* Fixed false positive in not_constant test when using groupings and a grouping has only one row by @igor-lobanov-maersk https://github.com/dbt-labs/dbt-utils/issues/848
 
 # dbt utils v1.1.0
 ## What's Changed
--- a/macros/generic_tests/not_constant.sql
+++ b/macros/generic_tests/not_constant.sql
@@ -22,6 +22,6 @@ from {{ model }}
   {{groupby_gb_cols}}
 
 having count(distinct {{ column_name }}) = 1
-
+  and count(1) > 1
 
 {% endmacro %}


dbeatty10 (changes requested):
@igor-lobanov-maersk many apologies for the delayed review here!

Could you add a test case for the scenario that your PR is fixing? That way we can confirm it fails without your fix but succeeds afterwards.

Here's an example:
- https://github.com/dbt-labs/dbt-utils/blob/9237ba9ab060870d23f4f0d14a8df2ab28d0b387/integration_tests/data/schema_tests/data_test_not_constant.csv?plain=1#L1-L4
- https://github.com/dbt-labs/dbt-utils/blob/9237ba9ab060870d23f4f0d14a8df2ab28d0b387/integration_tests/models/generic_tests/schema.yml#L4-L10

Issue #861, opened 2024-01-05
Support streaming tables in macro databricks get_table_types_sql()


w0ut0:
### Describe the bug

There is a 'new' table type in Databricks, `streaming tables`. Current get_table_types_sql() Databricks macro does not support this yet. The new table type is already implemented in the dbt-databricks adapter.

### Steps to reproduce

na

### Expected results

na

### Actual results

na

### Screenshots and log output

na

### System information
**The contents of your `packages.yml` file:**

**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [ x] other (specify: databricks)

**The output of `dbt --version`:**
```
<output goes here>
```

### Are you interested in contributing the fix?

fix incoming

w0ut0:
Keep open

w0ut0:
Bump

PR #862, opened 2024-01-05
include streaming_table and fix materialized_view in databricks__get_table_types_sql()

w0ut0:
solves #861 
solves #881 

This is a:
- [ ] documentation update
- [ ] new functionality
- [ ] a breaking change

## Description & motivation
databricks__get_table_types_sql() will now handle streaming_tables

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake


Dynosphere:
This is the same file that has a bug with MATERIALIZED VIEW.
If you could roll in the change from `MATERIALIZED VIEW` -> `MATERIALIZED_VIEW` we can also close: https://github.com/dbt-labs/dbt-utils/issues/881

w0ut0:
Hi @Dynosphere, I added your suggestion, thanks!
Is there something that still needs to be done in order to get this merged? I see it wasn't included in the 1.3.0 release.
Maybe @dbeatty10?

w0ut0:
Keep open

macros/sql/get_table_types_sql.sql
```diff
@@ -26,7 +26,8 @@
             case table_type
                 when 'MANAGED' then 'table'
                 when 'BASE TABLE' then 'table'
-                when 'MATERIALIZED VIEW' then 'materializedview'
+                when 'MATERIALIZED_VIEW' then 'materializedview'
```
davehowell, line 29: I don't think this is quite right. The input is correct, but the return should be 'materialized_view'

See the RelationType enum here -> https://github.com/dbt-labs/dbt-adapters/blob/e966f0ddea91ae531d92e703696bb80144e5b28e/dbt-adapters/src/dbt/adapters/contracts/relation.py#L20

PR #864, opened 2024-01-23
make column name to be fully qualified to prevent BQ confusion

### LINKED ISSUE #173: union should explicitly specify table name in casts
Bigquery gets confused by tables that have a column with the same name in CASTs, and will try to cast the entire table instead of the column.

```
-- causes error, as foo is a struct
select cast(`foo` as STRING) as `foo` 
from `data-playground-1234`.`myschema`.`foo`

-- works when table name is specified
select cast(`foo`.`foo` as STRING) as `foo` 
from `data-playground-1234`.`myschema`.`foo`
```

I *think* this is a simple as adding `table.name` [here](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/union.sql#L61), like:

```                    {%- set col_name = adapter.quote(table.name) + '.' + adapter.quote(col_name) if col_name in table_columns[table] else 'null' %}```

adrian3ka:
resolves #173 

This is a:
- [ ] documentation update
- [ ] new functionality
- [ ] a breaking change

## Description & motivation
For better support for an edge case in Bigquery, which gets confused by casting a column with the same name as its table name. 

## Checklist
- [x] This code is associated with an Issue that has been triaged and [accepted for development](https://docs.getdbt.com/docs/contributing/oss-expectations#pull-requests). 
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/sql/union.sql
+++ b/macros/sql/union.sql
@@ -107,7 +107,8 @@
 
                     {%- set col = column_superset[col_name] %}
                     {%- set col_type = column_override.get(col.column, col.data_type) %}
-                    {%- set col_name = adapter.quote(col_name) if col_name in relation_columns[relation] else 'null' %}
+                    {%- set column_exists = col_name in relation_columns[relation] %}
+                    {%- set col_name = adapter.quote(table.name) ~ '.' ~ adapter.quote(col_name) if column_exists else 'null' %}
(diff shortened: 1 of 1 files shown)


adrian3ka:
I am not so sure why the CI is failing and giving irrelevant error(s).

PR #892, opened 2024-05-20
feat: add callable where clause in union_relations

tnightengale:
resolves #

This is a:
- [ ] documentation update
- [x] new functionality
- [ ] a breaking change

## Description & motivation
I noticed that the `where` parameter in the `union_relations` macro does not allow one to create unique where clauses for each relation being unioned. In the case of an incremental model, it would be very helpful to be able to configure an incremental where clause for each model separately. 

In the simplest form, a project could implement a macro like the following:
```
{% macro where_clause(relation) %}

{% if execute %}
{% set where_clause %}
{% if is_incremental() %}
occurred_at >= coalesce((select max(occurred_at) from {{ relation }}), '1900-01-01')
{% else %}
    true
{% endif %}
{% endset %}
{% endif %}

{{ where_clause }}

{% endmacro %}
```

And the pass it to the `union_relations macro like so:
```
{{
    config(
        materialized="incremental"
    )
}}


{{ dbt_utils.union_relations(
    relations=[
        ref("model_a"),
        ref("model_b")
    ],
    where=where_clause
) }}
```

## Checklist
    - [x] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/sql/union.sql
+++ b/macros/sql/union.sql
@@ -114,7 +114,10 @@
 
             from {{ relation }}
 
-            {% if where -%}
+            {% if where is callable -%}
+            {# Allows possibility of customizing where clause by relation, eg. in incremental runs. #}
+            {{ where(relation) }} 
+            {% elif where -%}
             where {{ where }}
             {%- endif %}
         )


tnightengale:
I would welcome some discussion on this feature before I endeavor to add tests and the relevant changes to the README and CHANGELOG files.

Cheers!

thisismath:
I recently deployed similar functionality to my local repository. Rather than pull another macro into the `union_relations`, its possible to add the functionality for unique where clauses for each relation while maintaining the current functionality of same where deployed to all, by updating `macro.sql`'s [logic for the where clause](https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/union.sql#L117) like this instead:

proposed change:
```
{% if where -%}
  {% if where is string -%}
    where {{ where }}
  {% else %}
    where {{ where[loop.index0] }}
  {%- endif %}    
{%- endif %}
```

doing so allows for lists to be passed to the where param when a unique clause is necessary for each relation and maintains the use of passing a string for using the same where clause should be used for all - maintaining code integrity for all current deployments of the macro

tnightengale:
@dbeatty10 Would you mind reviewing this draft? And if you agree, I can finish off the PR with tests and update the docs?

PR #893, opened 2024-05-22
update tests: -> data_tests:

### LINKED ISSUE #894: Updated syntax of data tests within YAML
### Describe the feature

With the new dbt-core v1.8, the syntax for data tests have updated from `tests:` to `data_tests:`.

While using `tests:` continues to work for now, it is deprecated and emits a warning message.

Since our integration tests uses the latest version of dbt (currently v1.8), we should update our syntax to get rid of the warning message.

graciegoheen:
resolves #894

This is a:
- [x] just an update to the integration_tests (non-breaking change)

## Description & motivation

With the new dbt-core `v1.8`, the syntax for data tests have updated from `tests:` to `data_tests:`. This PR updates that syntax to get rid of the warning message.

## Checklist
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ version: 2
 
 models:
   - name: model_name
-    tests:
+    data_tests:
       - dbt_utils.equal_rowcount:
           compare_model: ref('other_table_name')
 
@@ -105,7 +105,7 @@ version: 2
 
 models:
   - name: model_name
-    tests:
+    data_tests:
       - dbt_utils.fewer_rows_than:
           compare_model: ref('other_table_name')
 ```
@@ -124,13 +124,13 @@ version: 2
 models:
   # compare the entire table 
   - name: model_name
-    tests:
+    data_tests:
       - dbt_utils.equality:
           compare_model: ref('other_table_name')
 
   # only compare some of the columns
   - name: model_name_compare_columns
-    tests:
+    data_tests:
       - dbt_utils.equality:
           compare_model: ref('other_table_name')
           compare_columns:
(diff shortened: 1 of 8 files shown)


dbt_project.yml
```diff
@@ -1,7 +1,7 @@
 name: 'dbt_utils'
 version: '0.1.0'
 
-require-dbt-version: [">=1.3.0", "<2.0.0"]
+require-dbt-version: [">=1.8.0", "<2.0.0"]
```
dbeatty10, line 4: Updating `tests` to `data_tests` looks like a breaking change to me, and would probably necessitate bumping the semantic version of `dbt_utils` from 1.x.x to 2.x.x.

On a related note, we've been trying to keep the required dbt versions as wide as possible.

One of the reasons is that narrowing the range of versions has downstream effects on other package maintainers that include `dbt_utils` within their dependencies.

Are there any alternatives to changing `tests` -> `data_tests`?

graciegoheen: This is only a change to the integration tests, so it's not a breaking change. I've removed this required version bump

README.md
```diff
@@ -86,7 +86,7 @@ version: 2
 
 models:
   - name: model_name
-    tests:
+    data_tests:
```
dbeatty10, line 89: Let's leave the README as-is right now so that the examples work across the broadest range of dbt-core.

But we can definitely revisit making this update in a separate PR, especially when there aren't any supported versions of dbt-core that allow `tests:` within `dbt_project.yml`.

PR #899, opened 2024-05-28
Fix unquoted database name

skion:
This fixes `get_relations_by_prefix()` when the database name contains e.g. a hyphen on Postgres.

This is a:
- [ ] documentation update
- [v] bug fix with no breaking changes
- [ ] new functionality
- [ ] a breaking change

## Checklist
- [v] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [v] Postgres
    - [ ] Redshift
    - [ ] Snowflake

### DIFF
--- a/macros/sql/get_tables_by_pattern_sql.sql
+++ b/macros/sql/get_tables_by_pattern_sql.sql
@@ -9,7 +9,7 @@
             table_schema as {{ adapter.quote('table_schema') }},
             table_name as {{ adapter.quote('table_name') }},
             {{ dbt_utils.get_table_types_sql() }}
-        from {{ database }}.information_schema.tables
+        from {{ adapter.quote(database) }}.information_schema.tables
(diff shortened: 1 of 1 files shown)


dbeatty10:
Thanks for putting this together @skion !

Taking an initial look, your proposal is simple and aligns with all the other macros. Ideally, we'd add a test case that includes a `"mIxEdCaSe"` database name so that we can validate that all the data warehouse we test work in this scenario (and also to prevent regressions in the future). But that might prove to be a bit tricky / impractical. So we might need to fall back on some manual testing for this. I'm doing to do some manual testing with Postgres, Snowflake, and Databricks both with and without the changes in your PR. I'll report back later with those results.

skion:
Any luck on this @dbeatty10 ?

skion:
Looks like still an issue in `main`...

PR #922, opened 2024-06-25
Revert pruning in `at_least_one` test

### LINKED ISSUE #838: bug in at_least_one test code logic
### Describe the bug
From v1.1.0 to v1.1.1 there was a change to `at_least_one` test´s code.
I think this change actually broke the way the test should work. The bug is in the logic for `pruned_rows` CTE.

### Steps to reproduce
Let´s look at a simple table, say `demo_table`:

```
| col_1 | col_2 |
+-------+-------+
| A     | 10    |
+-------+-------+
| B     | null  |
+-------+-------+
```

If we were to configure the `at_least_one` test for this table, using `group_by_columns` parameter,  we would have the following:

`select_gb_cols = 'col_1'`
`groupby_gb_cols  = 'group by col_1'`
`select_pruned_col = 'col_1, col_2'`

Then, in `pruned_rows` CTE:

```
select
  col_1, col_2
from demo_table
where col_2 is not null
limit 1
```

The output of this CTE would be:
```
| col_1 | col_2 |
+-------+-------+
| A     | 10    |
+-------+-------+
```
Continuing following the test's code, the following piece of logic would now be applied:
```
select
   col_1,
   count(col_2) as filler_column
from pruned_rows
group by col_1
having count(col_2) = 0
```
The output would be empty, as our only record in `pruned_rows` does have a non-null value for `col_2`.

### Expected results
I would expect this test to return the record with `col_1=B`, since for this value there are no non-null values in `col_2`.

### Actual results
Explained above.

### System information
dbt-labs/dbt_utils version 1.1.1
dbt version 1.6.3
bigquery plugin version 1.6.4

- [ ] postgres
- [ ] redshift
- [x] bigquery
- [ ] snowflake

Can definitely provide more context if needed, but not interested in developing the code.

katieclaiborne-duet:
resolves #838 

### Problem

The pruning logic introduced in #776 disrupted the grouping behavior in the `at_least_one` test.

Specifically, the initial `pruned_rows` CTE includes a `limit 1` statement, which selects a single record where the target column is not null.

As a result, the test passes inappropriately when the `group_by_columns` parameter is configured, as long as the target column has at least one not-null value.

Essentially, the `group_by_columns` parameter currently has no effect on the test. It evaluates target column values as a single group, regardless of whether or not the parameter is present.

### Solution

Restore the `at_least_one` test to its prior state.

Since the goal of #776 was to improve the performance of the test, I also considered a set-logic approach to evaluating a column at the group level:
- identify all group-by column values
- identify group-by column values with at least one not-null target column value
- return group-by column values without at least one not-null target column value as errors

In BigQuery, however, this version of the test took twice as long as the simple reverted version.

## Checklist

### DIFF
--- a/macros/generic_tests/at_least_one.sql
+++ b/macros/generic_tests/at_least_one.sql
@@ -4,20 +4,39 @@
 
 {% macro default__test_at_least_one(model, column_name, group_by_columns) %}
 
+{% set pruned_cols = [column_name] %}
+
 {% if group_by_columns|length() > 0 %}
   {% set select_gb_cols = group_by_columns|join(' ,') + ', ' %}
   {% set groupby_gb_cols = 'group by ' + group_by_columns|join(',') %}
+  {% set pruned_cols = group_by_columns %}
+
+  {% if column_name not in pruned_cols %}
+    {% do pruned_cols.append(column_name) %}
+  {% endif %}
+
 {% endif %}
+
+{% set select_pruned_cols = pruned_cols|join(' ,') %}
 
 select *
 from (
+    with pruned_rows as (
+      select
+        {{ select_pruned_cols }}
+      from {{ model }}
+      {% if group_by_columns|length() == 0 %}
+        where {{ column_name }} is not null
+        limit 1
+      {% endif %}
+    )
     select
         {# In TSQL, subquery aggregate columns need aliases #}
         {# thus: a filler col name, 'filler_column' #}
       {{select_gb_cols}}
       count({{ column_name }}) as filler_column
 
-    from {{ model }}
+    from pruned_rows
 
     {{groupby_gb_cols}}
 
--- a/macros/generic_tests/at_least_one.sql
+++ b/macros/generic_tests/at_least_one.sql
@@ -4,38 +4,20 @@
 
 {% macro default__test_at_least_one(model, column_name, group_by_columns) %}
 
-{% set pruned_cols = [column_name] %}
-
 {% if group_by_columns|length() > 0 %}
-
   {% set select_gb_cols = group_by_columns|join(' ,') + ', ' %}
   {% set groupby_gb_cols = 'group by ' + group_by_columns|join(',') %}
-  {% set pruned_cols = group_by_columns %}
-
-  {% if column_name not in pruned_cols %}
-    {% do pruned_cols.append(column_name) %}
-  {% endif %}
-
 {% endif %}
-
-{% set select_pruned_cols = pruned_cols|join(' ,') %}
 
 select *
 from (
-    with pruned_rows as (
-      select
-        {{ select_pruned_cols }}
-      from {{ model }}
-      where {{ column_name }} is not null
-      limit 1
-    )
     select
         {# In TSQL, subquery aggregate columns need aliases #}
         {# thus: a filler col name, 'filler_column' #}
       {{select_gb_cols}}
       count({{ column_name }}) as filler_column
 
-    from pruned_rows
+    from {{ model }}
 
     {{groupby_gb_cols}}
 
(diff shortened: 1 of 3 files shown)


macros/generic_tests/at_least_one.sql
```diff
@@ -4,38 +4,20 @@
...
-      limit 1
-    )
     select
         {# In TSQL, subquery aggregate columns need aliases #}
         {# thus: a filler col name, 'filler_column' #}
       {{select_gb_cols}}
       count({{ column_name }}) as filler_column
 
-    from pruned_rows
+    from {{ model }}
```
katieclaiborne-duet, line 20: Is there a way to add integration tests to validate that the test fails appropriately?

katieclaiborne-duet:
@dbeatty10, would you mind to review when you have a chance?

macros/generic_tests/at_least_one.sql
dbeatty10: Yes, adding a test would be the key thing here 🔑

[Here](https://github.com/dbt-labs/dbt-utils/blob/4da44cb930df07921e9c4335dba2e30f9e0802d4/integration_tests/models/generic_tests/schema.yml#L151-L152) is an example of using  `error_if` in a clever way.

### Explanation

By default, dbt treats:
- `0` row as `PASS`
- `>0` rows as `FAIL`

Overriding the config to `error_if: "<1"` will invert that. To the best of my knowledge, this feature was discovered rather than designed, and it also requires adding `warn_if: "<0"` to work properly.

dbeatty10 (commented):
A crucial step before merging this would be the addition of test(s) as described in https://github.com/dbt-labs/dbt-utils/pull/922#discussion_r1662536020

macros/generic_tests/at_least_one.sql
katieclaiborne-duet: Sweet! Thanks, Doug. I've given this a go in https://github.com/dbt-labs/dbt-utils/pull/922/commits/1fe5fcfc28475f1addb1b038db8a529b71cfe30f.

dbeatty10 (commented):
Thanks for raising this PR @katieclaiborne-duet 🤩 

I used https://github.com/dbt-labs/dbt-utils/pull/934 to quickly check that the new test you added correctly surfaces the error case discovered in https://github.com/dbt-labs/dbt-utils/issues/838. Looks good ✅ 

It would be nice to keep the optimization and also fix the bug 🤞 

When `group_by_columns` is an empty list, is there a way to keep the optimizations in https://github.com/dbt-labs/dbt-utils/issues/776 but avoid the bug reported in https://github.com/dbt-labs/dbt-utils/issues/838?

dbeatty10:
### Background

The `at_least_one` data test was originally added in https://github.com/dbt-labs/dbt-utils/pull/32. Then https://github.com/dbt-labs/dbt-utils/pull/633 / https://github.com/dbt-labs/dbt-utils/commit/78afeddc6bde120f776fbafeee6f04f854b302a3 added the `group_by_columns` argument, but no test was added that would catch a future regression. Then https://github.com/dbt-labs/dbt-utils/pull/776 accidentally added a regression, and this was reported in https://github.com/dbt-labs/dbt-utils/issues/838.

katieclaiborne-duet:
> When group_by_columns is an empty list, is there a way to keep the optimizations in https://github.com/dbt-labs/dbt-utils/pull/776 but avoid the bug reported in https://github.com/dbt-labs/dbt-utils/issues/838?

Good idea! Let me know what you think of this latest pass.

dbeatty10 (approved):
This looks awesome @katieclaiborne-duet 🤩

Thank you for creating a relevant test for this bug and implementing a targeted fix 🧠

tffragoso:
Thank you so much for the fix, @katieclaiborne-duet!  🙌

Issue #928, opened 2024-07-01
BigQuery `deduplicate` macro doesn't support downstream partition pruning


austinclooney:
### Describe the feature
The [Bigquery deduplicate macro](https://github.com/dbt-labs/dbt-utils/blob/4da44cb930df07921e9c4335dba2e30f9e0802d4/macros/sql/deduplicate.sql#L96) uses `array_agg` to deduplicate. The way it is currently set up, any query using the macro will not be able to partition prune downstream of the macro due to the way `array_agg` interacts with partition pruning because the partition column is not explicitly selected separate from the array agg. This can be avoided in a table by doing the partition filtering in a CTE prior to using the deduplicate macro, but it can't be avoided in a view. 

### Describe alternatives you've considered
For models materialized as tables, we can add the partition filtering above the deduplication. For views, the alternative is to write the deduplication step ourselves to manual select the partition column outside of the `array_agg`.

Basically it is written like this:

```
    select unique.*
    from (
        select
            array_agg (
                original
                order by {{ order_by }}
                limit 1
            )[offset(0)] unique
        from {{ relation }} original
        group by {{ partition_by }}
    )
```

and instead it needs to look like this:
```
    select <partition pruning column>, unique.* except(<partition pruning column>)
    from (
        select
            <partition pruning column>,
            array_agg (
                original
                order by {{ order_by }}
                limit 1
            )[offset(0)] unique
        from {{ relation }} original
        group by <partition pruning column>, {{ partition_by }}
    )
```

### Additional context
This is specific to the BQ deuplicate macro.

### Who will this benefit?
BQ users who want to deduplicate in a view. 

### Are you interested in contributing this feature?
I can create a PR that implements the fix we used.

austinclooney:
I've created a PR with an example solution, but will be happy to work with someone if it doesn't meet standards.

PR #929, opened 2024-07-01
Update BQ deduplicate macro to support partition pruning downstream

austinclooney:
resolves https://github.com/dbt-labs/dbt-utils/issues/928

### Problem

The current BQ deduplicate macro will not work for a view as it doesn't allow partition pruning downstream of the macro. 

### Solution

Add an an optional parameter that explicitly selects the `partition by` columns outside of the `array_agg`

## Checklist
- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-utils/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have run this code in development and it appears to resolve the stated issue

### DIFF
--- a/macros/sql/deduplicate.sql
+++ b/macros/sql/deduplicate.sql
@@ -93,12 +93,19 @@
 --  clause in BigQuery:
 --  https://github.com/dbt-labs/dbt-utils/issues/335#issuecomment-788157572
 #}
-{%- macro bigquery__deduplicate(relation, partition_by, order_by) -%}
+{%- macro bigquery__deduplicate(relation, partition_by, order_by, partition_by_pass_thru=false) -%}
 
     select unique.*
+    {%- if partition_by_pass_thru %}
+    except({{ partition_by }}),
+    {{ partition_by }}
+    {%- endif %}
     from (
         select
-            array_agg (
+            {%- if partition_by_pass_thru %}
+            {{ partition_by }},
+            {%- endif %}
+            array_agg(
                 original
                 order by {{ order_by }}
                 limit 1


macros/sql/deduplicate.sql
```diff
@@ -93,12 +93,19 @@
 --  clause in BigQuery:
 --  https://github.com/dbt-labs/dbt-utils/issues/335#issuecomment-788157572
 #}
-{%- macro bigquery__deduplicate(relation, partition_by, order_by) -%}
+{%- macro bigquery__deduplicate(relation, partition_by, order_by, partition_by_pass_thru=false) -%}
```
dbeatty10, line 96: Thanks for raising this PR @austinclooney !

I'd prefer not to introduce a new parameter if we can avoid it.

BigQuery supports the [`qualify`](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#qualify_clause) clause. Is there any reason we shouldn't / couldn't just use `qualify` in the same way as  [Snowflake and Databricks](https://github.com/dbt-labs/dbt-utils/blob/4da44cb930df07921e9c4335dba2e30f9e0802d4/macros/sql/deduplicate.sql#L59-L89)?

The original implementation for BigQuery was for [performance reasons](https://github.com/dbt-labs/dbt-utils/issues/335#issuecomment-788157572). But I'm not sure if those performance reasons still exist or if working for views outweighs those considerations.

austinclooney: To my knowledge the memory issue still exists, yes. Deduplicating with `array_agg` uses fewer resources than with `qualify` which allows it to be done on larger datasets before hitting memory issues. The reason I added a parameter here is because explicitly selecting the `partition_by` by default will change the column order in the output so I think it would technically be a breaking change?

dbeatty10: There are several different scenarios to consider for the relation being deduplicated:
- table
- view
- materialized view
- CTE

What would be the consequences if we did **not** add a new parameter named `partition_by_pass_thru` and just treated it as always being `true` instead?

austinclooney: I **_believe_** the only difference between the current method and the changes added in this PR is this PR would move the `partition_by` columns to the end of the output (or the beginning if preferred) and the current output retains the original column order. 

Since we're only explicitly selecting the partitioning keys it won't affect the actual deduplication as those are the groups we're deduplicating within.

dbeatty10: So the net consequence would be the following?
- partition pruning allowing faster queries for less cost
- output relation has the same columns in the same order as before

And they would equally apply to all of tables, views, materialized views, and CTE?

austinclooney: Yes

dbeatty10: Can you think of any way to do the following?
- output relation with the same columns in the same order as before

If so, that would allow us to ship these performance improvements without any breaking changes or new parameters.

austinclooney: I've thought about this for a while but I don't believe it's possible without running another query since the input to the macro can be a CTE. 

Something like this would work:

```sql
{%- macro bigquery__deduplicate(relation, partition_by, order_by) -%}
    {%- set columns_query -%}
        select * from {{ relation }} limit 1
    {%- endset -%}

    {%- set columns_result = run_query(columns_query) -%}
    {%- set columns = columns_result.column_names -%}

    select
        {%- for column in columns %}
            unique.{{ column }}{%- if not loop.last %},{% endif %}
        {%- endfor %}
    from (
        select
            {{ partition_by }},
            array_agg(
                original
                order by {{ order_by }}
                limit 1
            )[offset(0)] unique
        from {{ relation }} original
        group by {{ partition_by }}
    )
{%- endmacro -%}
```

But I don't think we'd want to run an external query to get the macro to work.

PR #945, opened 2024-09-02
update tests: -> data_tests: in README.md

syou6162:
resolves https://github.com/dbt-labs/dbt-utils/issues/894

### Problem
- The README was not fixed at https://github.com/dbt-labs/dbt-utils/pull/893.

### Solution
- I updated tests: -> data_tests: in README.md

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ version: 2
 
 models:
   - name: model_name
-    tests:
+    data-tests:
       - dbt_utils.equal_rowcount:
           compare_model: ref('other_table_name')
 
@@ -105,7 +105,7 @@ version: 2
 
 models:
   - name: model_name
-    tests:
+    data-tests:
       - dbt_utils.fewer_rows_than:
           compare_model: ref('other_table_name')
 ```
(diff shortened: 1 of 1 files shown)


syou6162:
@dbeatty10 Could you review this pull request?

dbeatty10 (commented):
It looks like https://github.com/dbt-labs/dbt-utils/pull/893 _intentionally_ only changed the `integration_tests` folder in order for it to be a non-breaking change.

So I think we're going to leave the documentation as-is.

If we did choose to merge this change, it would need to be `data_tests:` (underscore!) rather than `data-tests:`.

syou6162:
@dbeatty10 Thank you for your response. I have corrected `data-tests:` to `data_tests:`.

According to the [official documentation](https://docs.getdbt.com/docs/build/data-tests), `tests:` will continue to be supported, but dbt 1.8 and later will output a warning by default. I think that the number of users of dbt 1.8 and later will continue to increase, and [versions prior to dbt 1.8 will reach the end of life support level on Nov 1, 2024](https://docs.getdbt.com/docs/dbt-versions/core). I think that dbt-labs would like to recommend `data_tests:`, and I think that `data_tests:` is appropriate as the documentation provided by dbt-labs.

dbeatty10:
> @dbeatty10 Thank you for your response. I have corrected `data-tests:` to `data_tests:`.
> 
> According to the [official documentation](https://docs.getdbt.com/docs/build/data-tests), `tests:` will continue to be supported, but dbt 1.8 and later will output a warning by default.

Is avoiding the warning your primary concern here? Or alignment with the official documentation? It's okay if your main concern is both or something completely else too!

If it's the warning, we do have an open issue to cease raising a warning here: https://github.com/dbt-labs/dbt-core/issues/10564. We'd probably update the official documentation at the same time.

PR #964, opened 2024-11-13
feat(nulltest): add not_null option to unique_combination_of_columns.sql

### LINKED ISSUE #963: feat(nulltest): add not_null option to unique_combination_of_columns
### Problem

We perform unique and not null tests on all primary key columns. Previously, not_null data tests had to be specified repeatedly for the same columns that are also listed in the unique_combination_of_columns test.

### Solution

The proposed solution adds an optional for loop that performs not null tests on each column in the unique_combination_of_columns test. Gives neater code. The default parameter is false, so it should not be a breaking change.

### Describe the feature
A neat way to enforce primary keys is to have an option to not only use the "unique_combination_of_columns" test for unique tests of composite keys, but also for enforcing the not_null constraint for each of the underlying columns. This way developers have the option of not having to individually repeat single not_null tests for each of the columns listed in the "unique_combination_of_columns" test. This results in better readability, and hence better code quality.

Have considered writing a separate macro which combines these two types of tests, but I think it seemed more useful to have the not_null option inside the unique_combination_of_columns macro.

### Additional context
Primary keys are not enforced on Snowflake, but unique_combination_of_columns and not_null can be used as a way around to enforce it.

### Who will this benefit?
Enforcing primary keys in all databases.

Yes, already have a pull request ready (simple for loop).

halfos:
Add option to enforce uniqueness and non-null constraints on composite primary key columns in one test.

This introduces an option to ensure that composite primary key columns are both unique and do not contain null values. This is beneficial as it reduces the need for duplicate tests by allowing these constraints to be specified directly in a single test.

resolves #963

### Problem
I perform unique and not null tests on all primary key columns. Previously, not_null data tests had to be specified repeatedly for the same columns that are also listed in the unique_combination_of_columns test.

### Solution
The proposed solution adds an optional for loop that performs not null tests on each column in the unique_combination_of_columns test. Gives neater code. The default parameter is false, so it should not be a breaking change.

One problem can be that putting them together in one test may make it difficult to see which part of the test failed.

## Checklist
- [ ] This code is associated with an [issue](https://github.com/dbt-labs/dbt-utils/issues) which has been triaged and [accepted for development](https://docs.getdbt.com/docs/contributing/oss-expectations#pull-requests).

### DIFF
--- a/README.md
+++ b/README.md
@@ -565,6 +565,17 @@ An optional `quote_columns` argument (`default=false`) can also be used if a col
 
 ```
 
+An optional `not_null` argument (`default=false`) can also be used to test for null values in each of the provided columns.
+```yaml
+- name: revenue_by_product_by_month
+  tests:
+    - dbt_utils.unique_combination_of_columns:
+        combination_of_columns:
+          - month
+          - group
+        not_null: true
+```
+
 ### accepted_range ([source](macros/generic_tests/accepted_range.sql))
 
 Asserts that a column's values fall inside an expected range. Any combination of `min_value` and `max_value` is allowed, and the range can be inclusive or exclusive. Provide [a `where` argument](https://docs.getdbt.com/reference/resource-configs/where) to filter to specific records only.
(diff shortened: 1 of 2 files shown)


halfos:
A major flaw with this suggestion as it is coded right now is that when the test fails, it does not say whether it is the unique combination or not_null part that fails.

PR #1001, opened 2025-03-25
Add model conditions to equality test

### LINKED ISSUE #704: equality test with where config
### Describe the bug
I would like to limit the rows compared in dbt_utils.equality test with a where clause.  

### Steps to reproduce

Create two models with the same set of columns. Configure equality test with where configuration - in this case a date limitation. 

`
version: 2

models:
  - name: model_v2
    tests:
      - dbt_utils.equality:
          compare_model: ref('model_v1')
          where: "dt_date > DATE_ADD(current_date('Europe/Stockholm'), INTERVAL -1 DAY)"
`

### Expected results

The equality test executed with the where clause added to both "model" and "compare_model".   

### Actual results
Compilation error 

Compilation Error in test dbt_utils_equality_model_v2_ref_model_v1_ (models/model_v2.yml)
11:48:56    Macro test_equality expected a Relation but received the value: (select * from `project`.`dataset`.`model_v2` where dt_date > DATE_ADD(current_date('Europe/Stockholm'), INTERVAL -1 DAY)) dbt_subquery
11:48:56    
11:48:56    > in macro _is_relation (macros/cross_db_utils/_is_relation.sql)
11:48:56    > called by macro default__test_equality (macros/generic_tests/equality.sql)
11:48:56    > called by macro test_equality (macros/generic_tests/equality.sql)

### System information
packages:
  - package: dbt-labs/dbt_utils
    version: 0.8.6
  - package: mjirv/dbt_datamocktool
    version: 0.2.0

- [ ] postgres
- [ ] redshift
- [X] bigquery
- [ ] snowflake

```
Core:
  - installed: 1.1.1
  - latest:    1.3.0 - Update available!

Plugins:
  - bigquery: 1.1.1 - Update available!

  At least one plugin is out of date or incompatible with dbt-core.
```

### Additional context

It seems the check not working is https://github.com/dbt-labs/dbt-utils/blob/main/macros/generic_tests/equality.sql#L23

My understanding of the check is to make sure of the later use of adapter.get_columns_in_relation will work. 

My guess is that the most important part that the where config is added is in the CTEs named a and b.  

Sure - would be great with some hints on where to start

james-johnston-thumbtack:
resolves #704

### Problem

In #704, it's expressed that it would be useful to limit the rows compared in the `equality` test.  The issue was closed due to inactivity, but the discussion seems to indicate it was a reasonable request, and the only reason the issue was closed was because nobody got around to writing a PR:

> I guess we could also add a compare_model_condition parameter to this test, to accompany the where param. I'd be happy to review a PR in that vein if you wanted to open one.

> Although we are closing this issue as stale, it's not gone forever. Issues can be reopened if there is renewed community interest. Just add a comment to notify the maintainers.

### Solution

This adds the following parameters to `generic_tests/equality.sql`:

- `model_condition`
- `compare_model_condition`

It allows one to filter out rows from either model before comparing the remaining rows in the models as usual.

The rationale and usage is very similar to the `relationships_where` test.  For example, you can exclude recently-modified rows which might not be fully updated in the other table due to e.g. ELT limitations, etc.

Note that `model_condition` is introduced, because the normal 'where' clause can't be used because it means the given "model" is no longer a relation, and so the dbt_utils._is_relation check will fail.  This was discussed at https://github.com/dbt-labs/dbt-utils/issues/704#issuecomment-1344387690

Also note that by using two parameters, the user has full control over the filtering, and can choose different filters for each table.

## Checklist

NOTE: i tested by writing integration tests, and testing using the postgres target.


joellabes:
@james-johnston-thumbtack a quick message to say I see all the stuff you've done here! We will check this out but probably a couple of weeks away sorry 😬

PR #587, opened 2022-05-12
Fix: Make `union_relations` `include/exclude` case insensitive

jeremyyeo:
This is a:
- [ ] documentation update
- [x] bug fix with no breaking changes
- [ ] new functionality
- [ ] a breaking change

All pull requests from community contributors should target the `main` branch (default).

## Description & motivation
Fixes #578.

## Checklist
- [x] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [ ] Postgres
    - [ ] Redshift
    - [x] Snowflake
- [x] I followed guidelines to ensure that my changes will work on "non-core" adapters by:
    - [ ] dispatching any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/fishtown-analytics/dbt-utils/blob/master/macros/sql/star.sql))
    - [ ] using the `limit_zero()` macro in place of the literal string: `limit 0`
    - [ ] using `dbt_utils.type_*` macros instead of explicit datatypes (e.g. `dbt_utils.type_timestamp()` instead of `TIMESTAMP`
- [ ] I have updated the README.md (if applicable)
- [x] I have added tests & descriptions to my models (and macros if applicable)
- [x] I have added an entry to CHANGELOG.md

PR #1018, opened 2025-04-29
Make union relations case agnostic for include + column_override

### LINKED ISSUE #1017: union_relations: passing `include` parameter makes column_override fields case sensitive in data types quietly
### Describe the bug
When using the `union_relations` macro, when passing an `include` parameter, `column_override` requires the passed field names to be uppercase. Otherwise, at least for some data types/situations in Snowflake, the data types are not picked up for casting and the default behavior of using the first relation's datatypes are used.

Usability note/alternative: It's not clear to me if/why one should use both `include` and `column_override`, would be nice to just provide a dictionary with data types like `column_override` and know I also get the `include` behavior.

### Steps to reproduce
- Make a model using dbt_utils.union_relations to union two relations where not all upstream datatypes are varchar
- Add an `include` list of all fields or other set
- Add a column_override list that would cast a datatype in lowercase SQL, like  {"id": "number(38,0)", ...}
- Run DBT build on the model
- Observe the datatypes in the output

### Expected results

column_override datatypes are used for casting regardless of `include` list

The interaction between `include` and `union_relations` is clear and documented. At a minimum, providing an `include` list should not make `union_relations` case sensitive.

### Actual results
`column_override` is case sensitive and silently ignored when `include` list is passed.

### System information

- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [X] snowflake

```
1.9.4
```

Yes!

ttusing:
resolves #1017

### Problem

See https://github.com/dbt-labs/dbt-utils/issues/1017 - passing both `include` and `column_override` causes `column_override` to be case sensitive and silently fail/default to using first relation's datatypes.

AI summary from Claude:
```
Looking at the code, I can explain why the case sensitivity issue occurs with the include parameter. Here's what's happening in the original code:

1. When columns are processed from each relation, there's a check for inclusion that uses lowercase comparison:
{%- elif include and col.column | lower not in all_includes -%}

2. The column names that pass this check are stored in column_superset using their original case:
{%- do column_superset.update({col.column: col}) -%}

3. Later, when applying the column_override, the lookup is done using the original case of the column name from column_superset, but the column_override dictionary is used as-is without any case conversion.

This creates a mismatch because:

* The include filter converts column names to lowercase for comparison
* But the resulting column_superset maintains the original case of the columns
* Then the column_override lookup fails because it's trying to match the original case against whatever case was provided in the column_override dictionary
* This is why when using include, the column_override needs to match the original case of the columns in the source tables - because the case preservation in column_superset creates a case-sensitive context for the column_override lookup.
```

### Solution

Continuing AI generated output:
```
Our fix addresses this by converting both the column names and types in column_override to lowercase at the beginning:

{%- set lowercase_column_override = {} -%}
{%- if column_override is not none -%}
    {%- for col, type in column_override.items() -%}
        {%- do lowercase_column_override.update({col | lower: type | lower}) -%}
    {%- endfor -%}
{%- endif -%}

And then using lowercase comparison when looking up the override:

{%- set col_type = lowercase_column_override.get(col.column | lower, col.data_type) %}

This makes the behavior consistent regardless of whether include is used or not. The bug report should emphasize that this is a consistency fix that makes the column_override behavior case-insensitive in all scenarios, matching the case-insensitive behavior that already exists for the include and exclude parameters.

```

I tested this behavior locally and it works.

## Checklist
- [] This PR includes tests, or tests are not required/relevant for this PR
- [] I have updated the README.md (if applicable)

### DIFF
--- a/macros/sql/union.sql
+++ b/macros/sql/union.sql
@@ -14,6 +14,14 @@
     {% endif -%}
 
     {%- set column_override = column_override if column_override is not none else {} -%}
+    
+    {# Convert column_override keys to lowercase #}
+    {%- set lowercase_column_override = {} -%}
+    {%- if column_override is not none -%}
+        {%- for col, type in column_override.items() -%}
+            {%- do lowercase_column_override.update({col | lower: type | lower}) -%}
+        {%- endfor -%}
+    {%- endif -%}
 
     {%- set relation_columns = {} -%}
     {%- set column_superset = {} -%}
@@ -106,7 +114,7 @@
                 {% for col_name in ordered_column_names -%}
 
                     {%- set col = column_superset[col_name] %}
-                    {%- set col_type = column_override.get(col.column, col.data_type) %}
+                    {%- set col_type = lowercase_column_override.get(col.column | lower, col.data_type) %}
                     {%- set col_name = adapter.quote(col_name) if col_name in relation_columns[relation] else 'null' %}
                     cast({{ col_name }} as {{ col_type }}) as {{ col.quoted }} {% if not loop.last %},{% endif -%}
 


ttusing:
Reminds me of https://github.com/dbt-labs/dbt-utils/pull/587/files

ryramshaka:
Can someone approve I need this over 'id' and 'Id' from source tables 😭

PR #1051, opened 2025-09-24
Recency test to fail when no recent records are found

### LINKED ISSUE #1041: Issue
### Describe the bug

When using dbt_utils.recency test in combination with the config: where parameter to limit the scanned data the test will pass if no record is found to be less than the threshold - so the test is passing if no recent data are found! 

### Steps to reproduce

example seed (example.csv)
```
ingestion_date
2024-10-01
2025-01-01
```

test configuration on the seed (ie _seeds.yml)

```
seeds:
  - name: example
     tests:
       - dbt_utils.recency:
            arguments:
              datepart: day
              field: ingestion_date
              interval: 1
              config: 
                where: "ingestion_date > current_date - 2"
```

### Expected results

I would expect this to fail the tests (as it happens when there is no filters)

### Actual results

The test is actually passing! 

### System information
```
packages:
  - package: dbt-labs/dbt_utils
    version: 1.3.0
```

- [ ] postgres
- [ ] redshift
- [X] bigquery
- [X] snowflake

```
Latest of dbt 
```

### Additional context

The reason for the issue is we have this where clause here https://github.com/dbt-labs/dbt-utils/blob/31382da621330401e5bac528ac31a4dd2dd43cdc/macros/generic_tests/recency.sql#L39 
we should have something like this:
```
where most_recent < {{ threshold }} or most_recent is null
```

Yes happy to contribute to the fix, will try to open a PR for this :)

marzaccaro:
resolves #1041

  ### Problem

  

  The `dbt_utils.recency` test incorrectly passes when a `where` clause filters
  out all records.

  **Current behavior**: When a `where` clause results in no matching records,
  `most_recent` becomes `NULL`. The test only checks `where most_recent < {{
  threshold }}`, but since `NULL < threshold` evaluates to `NULL` (not true), no
  rows are returned and the test incorrectly passes.

  **Expected behavior**: The test should fail when no recent data exists (even
  when filtered by a where clause), indicating missing recent data.

  ### Solution

  

  Modified the `recency` macro in `macros/generic_tests/recency.sql` to include
  `or most_recent is null` in the WHERE clause. This ensures that when a where
  clause filters out all records (causing `most_recent` to be null), the test
  correctly fails by returning the null row.

  **Changes made**:
  - Updated the WHERE condition from `where most_recent < {{ threshold }}` to
  `where most_recent < {{ threshold }} or most_recent is null`
  - Added integration test `recency_with_where_filter` that reproduces the
  original bug and verifies the fix
  - Test uses data with old timestamps (10-15 days) and a where clause that
  filters to non-existent records

  **Testing approach**:
  - Before fix: Test incorrectly passes (returns 0 rows)
  - After fix: Test correctly fails (returns 1 row with null `most_recent`)
  - Verified existing recency tests continue to work as expected

  ## Checklist
  - [x] This code is associated with an
  [issue](https://github.com/dbt-labs/dbt-utils/issues/1041) which has been
  triaged and [accepted for development](https://docs.getdbt.com/docs/contributin
  g/oss-expectations#pull-requests).
  - [x] I have read [the contributing 
  guide](https://github.com/dbt-labs/dbt-utils/blob/main/CONTRIBUTING.md) and
  understand what's expected of me
  - [x] I have run this code in development and it appears to resolve the stated
  issue

### DIFF
--- a/integration_tests/models/generic_tests/recency_with_where_filter.sql
+++ b/integration_tests/models/generic_tests/recency_with_where_filter.sql
@@ -1,0 +1,12 @@
+-- Test model with data that's intentionally old
+-- This simulates a scenario where we have data but it's all outside the recency window
+-- this should fail with a where clause but currently it would pass
+select
+    1 as id,
+    {{ dbt.dateadd('day', -10, dbt.current_timestamp()) }} as created_at
+
+union all
+
+select
+    2 as id,
+    {{ dbt.dateadd('day', -15, dbt.current_timestamp()) }} as created_at
--- a/integration_tests/models/generic_tests/schema.yml
+++ b/integration_tests/models/generic_tests/schema.yml
@@ -261,3 +261,19 @@ models:
           compare_model: ref('data_test_equality_a')
           exclude_columns:
             - col_c
+
+  - name: recency_with_where_filter
+    data_tests:
+      # This test should FAIL and return 0 rows (test failure) because when we filter to id = 3,
+      # no records match the where condition, so most_recent will be null.
+      # Before the fix: test would incorrectly PASS (no rows returned)
+      # After the fix: test should correctly FAIL (returns the null row)
+      - dbt_utils.recency:
+          datepart: day
+          field: created_at
+          ignore_time_component: true
+          interval: 5
+          config:
+            where: "id = 3"  # This filters to no records since only id 1,2 exist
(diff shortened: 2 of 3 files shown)


marzaccaro:
closing as this is resolved by https://github.com/dbt-labs/dbt-utils/pull/1065

Issue #1059, opened 2025-11-25
Union Relations pulls full SQL to source relation when using microbatch


moseleyi:
### Describe the bug
When using incremental strategy **Microbatch** the relation gets updated with a WHERE clause and timeframe.

Two things happen:

1. The code fails
2. The relation name is full SQL

### Steps to reproduce
Create two base models
Create third model that unions the other two using `union_relations` but make it an incremental model with strategy `microbatch`

### Expected results
`_dbt_source_relation` should still say `db.schema.model_name`

### Actual results
Error
```
cast('(select * from "db"."sch"."em_bounces" where partition_date >= '2025-11-10 00:00:00+00:00' and partition_date < '2025-11-11 00:00:00+00:00')' as TEXT) as _dbt_source_relation,
 ```

### System information
**The contents of your `packages.yml` file:**
```
- name: dbt_utils
    package: dbt-labs/dbt_utils
    version: 1.3.1
```
**Which database are you using dbt with?**
- [ ] postgres
- [x] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
Core: 1.10.13
Redshift: 1.9.5
```

moseleyi:
Still quite a bug

PR #1061, opened 2025-12-05
fix: Render relations to avoid including time filters in source_column_name

harshvardhan-j:
resolves #
https://github.com/dbt-labs/dbt-utils/issues/1059

### Problem

If the model has event_time filters or micro-batch implemented the source_column_name returns the complete query for the relation like so
```
cast('(select * from "db"."sch"."em_bounces" where partition_date >= '2025-11-10 00:00:00+00:00' and partition_date < '2025-11-11 00:00:00+00:00')' as TEXT) as _dbt_source_relation,
```
This causes the macro to break and prevents setting up sampling and other related features.

This occurs because of the function in dbt-core [here](https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-adapters/src/dbt/adapters/base/relation.py#L294) which is used in the [__str__](https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-adapters/src/dbt/adapters/base/relation.py#L428) method that is used by `dbt.string_literal`.

### Solution

Use the 'render' method on the relation to obtain the relevant information as a string literal rather than the __str__ method of the relation.

It returns the expected SQL
```
WITH test_cte AS (



        (
            select
                cast('a.b.c' as TEXT) as source_system
```

## Checklist

### DIFF
--- a/macros/sql/union.sql
+++ b/macros/sql/union.sql
@@ -100,7 +100,7 @@
             select
 
                 {%- if source_column_name is not none %}
-                cast({{ dbt.string_literal(relation) }} as {{ dbt.type_string() }}) as {{ source_column_name }},
+                cast({{ dbt.string_literal(relation.render()) }} as {{ dbt.type_string() }}) as {{ source_column_name }},
                 {%- endif %}
 
                 {% for col_name in ordered_column_names -%}


joellabes (approved):
nice - thank you!

joellabes:
@harshvardhan-j it's out as 1.3.3!

PR #1063, opened 2025-12-18
patch union_relations macro to avoid including 'collation' in cast

colin-rogers-dbt:
resolves #

### Problem

We recently updated dbt-snowflake to return the collation on a column as part of the data_type as it will soon be required when altering a column's data type (if it's a string). Most of the time this is correct, unless you want to use that datatype in a `cast`. 

### Solution

Add a macro with a snowflake specific override to scrub `collation` from the datatype if it's present.

## Checklist
- [ ] This code is associated with an [issue](https://github.com/dbt-labs/dbt-utils/issues) which has been triaged and [accepted for development](https://docs.getdbt.com/docs/contributing/oss-expectations#pull-requests).
- [ ] I have read [the contributing guide](https://github.com/dbt-labs/dbt-utils/blob/main/CONTRIBUTING.md) and understand what's expected of me
- [ ] I have run this code in development and it appears to resolve the stated issue

### DIFF
--- a/macros/sql/union.sql
+++ b/macros/sql/union.sql
@@ -2,6 +2,27 @@
     {{ return(adapter.dispatch('union_relations', 'dbt_utils')(relations, column_override, include, exclude, source_column_name, where)) }}
 {% endmacro %}
 
+{%- macro union_cast_type(data_type) -%}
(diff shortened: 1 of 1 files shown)


dbeatty10:
Solved in https://github.com/dbt-labs/dbt-adapters/pull/1510 instead

PR #1070, opened 2026-02-14
Migrate top-level test 'values' to 'arguments' (dbt0102)

sushil6666:
Migrate top-level test `values` to `arguments` (dbt0102)

Summary:
Wrap top-level `values:` under the `arguments:` key for `accepted_values` tests in the integration tests schema to address dbt deprecation `dbt0102` (deprecated top-level test arguments). This keeps test behaviour identical while conforming to the newer schema.

Changes:
- Move `values:` into `arguments:` for all `accepted_values` usages in `integration_tests/models/sql/schema.yml`.

Why:
- Prevents deprecation warnings and future breakage in dbt-fusion/dbt versions that enforce the new schema.

Testing:
- This is a static YAML-only change; no runtime logic modified. Tests remain the same and will continue to pass in CI.

Notes:
- Created by automated migration from the local repo that detected dbt0102 warnings. See PR for patch details.

### DIFF
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,10 @@
+Migrate top-level test `values` to `arguments` (dbt0102)
+
+This repo contains a small YAML-only change to the integration tests schema to
+wrap `values:` under `arguments:` for `accepted_values` tests.
+
+Rationale:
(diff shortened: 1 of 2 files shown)


sushil6666:
done

Issue #881, opened 2024-04-16
Resolve support for Materialized views on databricks in dbt_utils.get_table_types_sql 


SLe-Corre:
### Describe the bug
Databricks have recently updated their information schema standards to standardize table types for materialized views and streaming tables.
This means the macro "Get_table_types_sql" and specifically the "databricks__get_table_types_sql()" part needs to reflect the move to `MATERIALIZED_VIEW` as opposed to `MATERIALIZED VIEW` as current.

Running the macro currently, or calling something like generate_sources from dbt_codegen which depends on this will return the error:
Runtime Error
  Field "type" of type Optional[DatabricksRelationType] in DatabricksRelation has invalid value 'materialized_view'

when trying to read from a schema containing a materialized table.

### Steps to reproduce
On databricks, have a schema containing a materialized view.
call the codegen macro  `codegen.generate_source()` or directly call the dbt utils get table types against this schema.

### Expected results

Code runs without error.

### Actual results

Code returns:
Runtime Error
  Field "type" of type Optional[DatabricksRelationType] in DatabricksRelation has invalid value 'materialized_view'

### Screenshots and log output

This is the code that gets generated, and successfully run against the warehouse. I've highlighted in yellow how materialized_view is falling under the else condition rather than the catchment case above it.

<img width="500" alt="image" src="https://github.com/dbt-labs/dbt-utils/assets/32954229/1c4fbdb7-bfde-470b-b19b-b9ce445b7c46">

### System information

packages:
  - package: dbt-labs/codegen
    version: 0.12.1 ## update to latest version here

  - package: dbt-labs/dbt_utils
    version: 1.1.1

**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [x] other (specify: Databricks)

**The output of `dbt --version`:**
```
Core:
  - installed: 1.7.11
  - latest:    1.7.11 - Up to date!

Plugins:
  - spark:      1.7.1  - Up to date!
  - databricks: 1.7.13 - Up to date!
```

### Additional context

This can be fixed by changing the line:
https://github.com/dbt-labs/dbt-utils/blob/951d087821fb4b07255cd74891ffe5b21e19a3ed/macros/sql/get_table_types_sql.sql#L29C1-L29C65

from `MATERIALIZED VIEW` to `MATERIALIZED_VIEW`

### Are you interested in contributing the fix?

[...]

PR #1071, opened 2026-02-16
Fix get_table_types for bigquery and databricks materialized view and streaming table

davehowell:
resolves #
https://github.com/dbt-labs/dbt-utils/issues/881
https://github.com/dbt-labs/dbt-utils/issues/861

Includes (corrected) fix from https://github.com/dbt-labs/dbt-utils/pull/862 

### Problem

I reproduce this bug using the dbt-codegen package, which in turn calls this dbt-utils package. Specifically I was trying to run `dbt run-operation generate_source --args '{"table_pattern": "%_raw"}'`

This ends up calling `dbt_utils.get_relations_by_pattern` which calls `dbt_utils.get_tables_by_pattern_sql` to fetch metadata from information_schema and then passes it to `api.Relation.create()` which then throws the error

    Invalid relation type: materializedview

### Solution

The dbt-databricks adapter uses `DatabricksRelationType`, see: https://github.com/databricks/dbt-databricks/blob/25caa2a14ed0535f08f6fd92e29b39df1f453e4d/dbt/adapters/databricks/relation.py#L40

The dbt-bigquery adapter uses the base `RelationType`, see: https://github.com/dbt-labs/dbt-adapters/blob/e966f0ddea91ae531d92e703696bb80144e5b28e/dbt-bigquery/src/dbt/adapters/bigquery/relation.py#L13

and where it is used:  https://github.com/dbt-labs/dbt-adapters/blob/e966f0ddea91ae531d92e703696bb80144e5b28e/dbt-adapters/src/dbt/adapters/contracts/relation.py#L20 

Also see the databricks table_types documented here https://docs.databricks.com/aws/en/sql/language-manual/information-schema/tables#table-types

This PR is a very small fix, and I have verified it works on my own dbt projects by monkey-patching it with a file `./macros/get_table_types_sql.sql` with the contents as per this PR

## Checklist

There are pre-existing tests for this macro, this PR is for a bug, not new functionality.

Not applicable, because functionality doesn't vary from the documented intention

### DIFF
--- a/macros/sql/get_table_types_sql.sql
+++ b/macros/sql/get_table_types_sql.sql
@@ -31,13 +31,3 @@
                 else lower(table_type)
             end as {{ adapter.quote('table_type') }}
 {% endmacro %}
-
-
-{% macro bigquery__get_table_types_sql() %}
-            case table_type
-                when 'BASE TABLE' then 'table'
-                when 'EXTERNAL TABLE' then 'external'
-                when 'MATERIALIZED VIEW' then 'materialized_view'
-                else lower(table_type)
-            end as {{ adapter.quote('table_type') }}
-{% endmacro %}
--- a/macros/sql/get_table_types_sql.sql
+++ b/macros/sql/get_table_types_sql.sql
@@ -30,3 +30,13 @@
                 else lower(table_type)
             end as {{ adapter.quote('table_type') }}
 {% endmacro %}
+
+
+{% macro bigquery__get_table_types_sql() %}
+            case table_type
+                when 'BASE TABLE' then 'table'
+                when 'EXTERNAL TABLE' then 'external'
+                when 'MATERIALIZED VIEW' then 'materialized_view'
+                else lower(table_type)
+            end as {{ adapter.quote('table_type') }}
+{% endmacro %}
--- a/macros/sql/get_table_types_sql.sql
+++ b/macros/sql/get_table_types_sql.sql
@@ -6,7 +6,7 @@
             case table_type
                 when 'BASE TABLE' then 'table'
                 when 'EXTERNAL TABLE' then 'external'
-                when 'MATERIALIZED VIEW' then 'materializedview'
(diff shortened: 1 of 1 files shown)


liaodaniel (approved):

b-per:
Thanks for the PR! The fix for Databricks looks correct — `MATERIALIZED_VIEW` (underscore) is indeed what Databricks returns in `information_schema.tables`, and `materialized_view` matches `DatabricksRelationType`.

However, I think the new `bigquery__get_table_types_sql` dispatch isn't needed. The simpler fix is to update the **default** dispatch (and the Postgres one) to output `materialized_view` instead of `materializedview`. The default already handles BigQuery's `'MATERIALIZED VIEW'` (with space) input correctly — it just had the wrong output value.

So the suggestion would be:
- Keep the Databricks changes (input match fix + `STREAMING_TABLE`)
- Drop the new `bigquery__get_table_types_sql` macro
- Update `default__get_table_types_sql` and `postgres__get_table_types_sql` to output `materialized_view` instead of `materializedview`

For context: `materializedview` was never a valid `RelationType` value. The macro was created in Feb 2022 (dbt-utils 0.8.1) before the `RelationType.MaterializedView` enum existed. When the enum was added to dbt-core in June 2023 (PR #7239), it used `materialized_view` (with underscore), but dbt-utils was never updated to match. There was actually a commit (`fbd295e`) that fixed this, but it landed on a `next/patch` branch that was never merged to `main`.

davehowell:
@b-per  Thank you for the review and suggestions. I agree your suggestions are better! I have pushed those changes so would appreciate the re-review.

I did start to change the default dispatch but I wasn't sure about the impact of modifying that would have on a bunch of other targets so I went with the more cautious BigQuery-only approach. I also wasn't sure about Postgres but have made that change too as you suggested.

davehowell:
@b-per  is there anything else I can help with here? I think I addressed your concerns, happy to do more if needed

PR #1082, opened 2026-04-24
Refactor not_empty_string macro to fix alias bugs and catch invisible…

shivasingh4945-tech:
**What changed & why:**
* **No more syntax crashes:** The old code tried to alias the column (`as {{ column_name }}`), which broke dbt if we tested a nested column (like `user.email`) or a SQL function. I moved the logic straight into the `WHERE` clause so it works on everything.
* **Catches sneaky characters:** Standard `trim()` was missing invisible characters like tabs and newlines. This now catches those so completely blank strings actually fail the test like they are supposed to.
* **Faster queries:** Instead of doing 4 separate string scans with nested `replace()` functions, it now uses `translate()` to swap out all hidden characters in one fast pass. 

Tested and verified on normal text, empty strings, spaces, tabs, and `NULL`s!

### DIFF
--- a/macros/generic_tests/not_empty_string.sql
+++ b/macros/generic_tests/not_empty_string.sql
@@ -1,37 +1,24 @@
 {% test not_empty_string(model, column_name, trim_whitespace=true) %}
-
   {{ return(adapter.dispatch('test_not_empty_string', 'dbt_utils')(model, column_name, trim_whitespace)) }}
-
 {% endtest %}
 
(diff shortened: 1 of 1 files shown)


shivasingh4945-tech:
Hi @remigabillet @mjumbewu  @haf @zqureshi 
I have tested this fix on PostgreSQL and it works correctly.

Would love your feedback on this approach. 
Happy to make any changes needed!

PR #1084, opened 2026-05-18
Reduce INFORMATION_SCHEMA calls in BigQuery get_tables_by_pattern_sql

emilmar:
### Problem

On BigQuery, `dbt_utils.get_relations_by_pattern` (via `get_tables_by_pattern_sql`) runs two kinds of `INFORMATION_SCHEMA` queries whenever the `schema_pattern` contains `%`:

1. An `INFORMATION_SCHEMA.SCHEMATA` lookup to enumerate matching datasets.
2. A `UNION ALL` of one dataset-scoped `INFORMATION_SCHEMA.TABLES` query per match.

This is more work than necessary. The region-qualified `` `project`.`region-X`.INFORMATION_SCHEMA.TABLES `` view already exposes `table_schema`, so the schema match can happen inside the same query as the table match.

### Solution

When `target.location` is set, query the region-qualified `INFORMATION_SCHEMA.TABLES` view once and filter `table_schema` directly. The `SCHEMATA` call and the `UNION ALL` go away. 

The existing `SCHEMATA` + iterate path is kept as a fallback for profiles without a configured location, so behaviour is unchanged for those users. The single-schema (no `%`) path is also unchanged. Since BigQuery is highly dependent on region, I assume most users set the target.location value. The previous version did not allow region override so it is still an improvement over that. I think we might even remove the fallback, and allow it to error instead.

### Note on CI coverage

`integration_tests/profiles.yml` doesn't set `location`, so the existing `test_get_relations_by_pattern` model exercises the fallback path, not the new one. Happy to add `location: US` to the integration profile if maintainers want the fast path covered in CI.

## Checklist

### DIFF
--- a/macros/sql/get_tables_by_pattern_sql.sql
+++ b/macros/sql/get_tables_by_pattern_sql.sql
@@ -45,6 +45,25 @@
 
 {% macro bigquery__get_tables_by_pattern_sql(schema_pattern, table_pattern, exclude='', database=target.database) %}
 
+    {# When `target.location` is set, the region-qualified INFORMATION_SCHEMA.TABLES
+       exposes `table_schema` directly. That collapses a SCHEMATA lookup plus a
+       per-dataset UNION ALL into one query. Falls back to the original
+       SCHEMATA + iterate path when no location is configured. #}
+    {% if '%' in schema_pattern and target.location %}
+        {% set sql %}
+            select distinct
+                table_schema,
+                table_name,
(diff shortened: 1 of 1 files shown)


emilmar:
This PR was created after a [discussion](https://getdbt.slack.com/archives/C088YCAB6GH/p1760612531832169) on undocumented rate limits in BigQuery that were hit due to dbt Fusion effectiveness.

PR #1085, opened 2026-05-18
Add BigQuery-native union_relations using FULL OUTER UNION ALL BY NAME

emilmar:
### Problem

On BigQuery, `dbt_utils.union_relations` calls `adapter.get_columns_in_relation` on every input relation, which on the BigQuery adapter is backed by an `INFORMATION_SCHEMA.COLUMNS` query per relation. It then emits a fully-enumerated `SELECT` per relation: every column from the cross-relation superset listed with explicit `cast(... as type)` plus `cast(null as type)` padding for columns missing from that relation. For wide tables or many relations this produces hundreds of lines of generated SQL.

BigQuery natively supports both halves of what the macro is doing by hand:

- `FULL OUTER UNION ALL BY NAME` matches columns by name across queries and fills missing columns with NULL.
- `SELECT * EXCEPT (...)` and `SELECT * REPLACE (cast(col as type) AS col)` cover `exclude` and `column_override` without enumerating columns.

_But most importantly_: it cuts out the need to call INFORMATION_SCHEMA, which means one less roundtrip to BigQuery.

### Solution

Add a `bigquery__union_relations` dispatch that uses these features directly. The new path **issues zero `INFORMATION_SCHEMA` queries** — no per-relation column lookup is needed at all, since `BY NAME` resolves the schema match at query time. Generated SQL collapses from N enumerated `SELECT` lists to one `*`-projection per relation joined by `FULL OUTER UNION ALL BY NAME`, with `EXCEPT`/`REPLACE` wrapped around the result for `exclude` and `column_override`.

Example output for `union_relations([a, b, c])`:

```sql
select * from (
    select cast('a' as string) as _dbt_source_relation, * from a
    full outer union all by name
    select cast('b' as string) as _dbt_source_relation, * from b
    full outer union all by name
    select cast('c' as string) as _dbt_source_relation, * from c
)
```

The `include` path falls back to the default implementation, since BigQuery has no native syntax for projecting a named subset of columns — column enumeration (and the `INFORMATION_SCHEMA` calls) are unavoidable there.

### Behaviour differences worth flagging

- **`column_override` referencing a column that exists in no relation now errors.** The default macro silently drops such overrides; BigQuery's `REPLACE` requires the column to exist in the projection. This seems more useful — a missing-column override is almost always a typo — but it is a tightening.
- **Cross-relation type coercion uses BigQuery's supertype rules** instead of "first relation wins". E.g. `INT64` in one relation and `FLOAT64` in another widen to `FLOAT64` rather than truncating to whichever was seen first. Strictly more correct, but technically different behaviour.
- **"No columns found" compiler error** is not raised in the fast path, since columns are no longer enumerated upfront. If all inputs are empty the union just returns no rows.

## Checklist

### DIFF
--- a/integration_tests/data/sql/data_union_exclude_expected.csv
+++ b/integration_tests/data/sql/data_union_exclude_expected.csv
@@ -2,5 +2,5 @@ id,favorite_color,favorite_number
 1,,pi
 2,,e
 3,,4
-1,"green",7
-2,"pink",13
+1,"green",seven
+2,"pink",thirteen
--- a/integration_tests/data/sql/data_union_expected.csv
+++ b/integration_tests/data/sql/data_union_expected.csv
@@ -2,5 +2,5 @@ id,name,favorite_color,favorite_number
 1,"drew",,pi
 2,"bob",,e
 3,"alice",,4
-1,,"green",7
-2,,"pink",13
+1,,"green",seven
+2,,"pink",thirteen
--- a/integration_tests/data/sql/data_union_table_2.csv
+++ b/integration_tests/data/sql/data_union_table_2.csv
@@ -1,3 +1,3 @@
 id,favorite_color,favorite_number
-1,green,7
-2,pink,13
+1,green,seven
+2,pink,thirteen
--- a/macros/sql/union.sql
+++ b/macros/sql/union.sql
@@ -18,6 +18,11 @@
 
     {%- set column_override = column_override if column_override is not none else {} -%}
 
+    {%- for relation in relations -%}
+        {%- do dbt_utils._is_relation(relation, 'union_relations') -%}
+        {%- do dbt_utils._is_ephemeral(relation, 'union_relations') -%}
+    {%- endfor -%}
+
     select *
         {%- if exclude %} except ({{ exclude | join(', ') }}){% endif -%}
         {%- if column_override %} replace (
@@ -27,9 +32,6 @@
         ){% endif %}
     from (
         {%- for relation in relations %}
-
-            {%- do dbt_utils._is_relation(relation, 'union_relations') -%}
-            {%- do dbt_utils._is_ephemeral(relation, 'union_relations') -%}
 
             select
                 {%- if source_column_name is not none %}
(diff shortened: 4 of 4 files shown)


emilmar:
This PR was created after a [discussion](https://getdbt.slack.com/archives/C088YCAB6GH/p1760612531832169) on undocumented rate limits in BigQuery that were hit due to dbt Fusion effectiveness.

emilmar:
I discovered undocumented behaviour in the default macro: it silently picks the first relation's column type and casts all others to it. The existing seed tests only pass because of that ordering — `data_union_table_1.favorite_number` is STRING and `data_union_table_2.favorite_number` is INT64, and the cast quietly stringifies the INT64. If the seeds had randomly been ordered the other way around, the test would have failed — I don't see this as a deliberate feature.

The BigQuery-native `FULL OUTER UNION ALL BY NAME` doesn't do that silent stringification — it asks for a real common supertype, finds none, and surfaces the mismatch. Arguably more honest, but it means the existing seeds need to actually agree on type.

Although this comment has the typical em-dash, I have to say I wrote some of this by hand!!! I trust this finding ;)

integration_tests/data/sql/data_union_exclude_expected.csv
```diff
@@ -2,5 +2,5 @@ id,favorite_color,favorite_number
 1,,pi
 2,,e
 3,,4
-1,"green",7
-2,"pink",13
+1,"green",seven
```
serramatutu, line 5: This would actually change a behavior that people (other dbt packages, people's dbt projects) implicitly depend on, so it's a breaking change (even if it's more "BigQuery native").

Ideally this would preserve the current behavior as we don't wan't people's project to start breaking in the middle of the night :)

serramatutu: I do like that your macro does not need to enumerate columns for every relation, relying only on `* excluding` syntax. But as a "black box" it should behave like the previous version did.

In other words, we should be able to swap the implementation under the hood without any noticeable changes to the return values

emilmar: Yep, you are correct and that is your call to make. Using all I know in BigQuery, the only way we could remove that SCHEMATA call (which can cause rate limits due to overload of introspective queries) was to remove the implicit call to string (god it really bugs me now that I know about it!).

I think the main takeaway on that feature is that half of users trying union with differing column types, will end up with an error, because
> If the seeds had randomly been ordered the other way around, the test would have failed

But yes, we can improve the column enumeration any ways!

Issue #822, opened 2023-08-07
Missing column specification when using `expression_is_true` test (snowflake)


Nic3Guy:
### Describe the bug
In Snowflake when I try to use the test `dbt_utils.expression_is_true` I get Database Error: SQL compilation error of `Missing column specification`

### Steps to reproduce
Set up the test under the column you would like and execute `dbt test -s <model_name>

### Expected results
The test should compile and run

### Actual results
The test is being compiled but is not running due to the compilation error in the database (in this case Snowflake)

### Screenshots and log output
![Screenshot 2023-08-07 154513](https://github.com/dbt-labs/dbt-utils/assets/17640648/cc7a3b39-8dbd-4786-ae2f-668c9fd45301)
![Screenshot 2023-08-07 154347](https://github.com/dbt-labs/dbt-utils/assets/17640648/1abd4e56-30b8-46e3-a469-8e897a422fc7)
the compiled test:
![Screenshot 2023-08-07 155343](https://github.com/dbt-labs/dbt-utils/assets/17640648/bbcb3c97-7b94-4387-b924-ba6947bd65eb)

### System information
**The contents of your `packages.yml` file:**
```
packages:
  - package: calogica/dbt_expectations
    version: [">=0.9.0", "<0.10.0"]
  - package: dbt-labs/codegen
    version: 0.10.0
  - package: elementary-data/elementary
    version: 0.9.0
```

**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [ x ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
```
Core:
  - installed: 1.5.0
  - latest:    1.6.0 - Update available!

  Your version of dbt-core is out of date!
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation

Plugins:
  - snowflake: 1.5.0 - Update available
```

### Additional context
When I created the same test from the source code but added a column alias like this it works as expected:
![image](https://github.com/dbt-labs/dbt-utils/assets/17640648/40db7997-4281-4b0b-ad3c-e9ed9751b89d)
```
{% test expression_is_true(model, expression, column_name=None) %}
	select
		1 as col
	from {{ model }}
	{% if column_name is none %}
		where not({{ expression }})
	{%- else %}
		where not({{ column_name }} {{ expression }})
	{%- endif %}
{% endtest %}
```

### Are you interested in contributing the fix?
Maybe why not 👍🏼

PR #700, opened 2022-10-04
Remove obsolete condition argument from expression_is_true

miles170:
Closes #695 

This is a:
- [ ] documentation update
- [ ] bug fix with no breaking changes
- [ ] new functionality
- [x] a breaking change

All pull requests from community contributors should target the `main` branch (default).

## Description & motivation

Remove `condition` argument since it can be replaced by `where`.

## Checklist
- [x] This code is associated with an Issue which has been triaged and [accepted for development](https://docs.getdbt.com/docs/contributing/oss-expectations#pull-requests). 
- [x] I have verified that these changes work locally on the following warehouses (Note: it's okay if you do not have access to all warehouses, this helps us understand what has been covered)
    - [ ] BigQuery
    - [x] Postgres
    - [ ] Redshift
    - [ ] Snowflake
- [ ] I followed guidelines to ensure that my changes will work on "non-core" adapters by:
    - [ ] dispatching any new macro(s) so non-core adapters can also use them (e.g. [the `star()` source](https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/star.sql))
    - [ ] using the `limit_zero()` macro in place of the literal string: `limit 0`
    - [ ] using `dbt.type_*` macros instead of explicit datatypes (e.g. `dbt.type_timestamp()` instead of `TIMESTAMP`
- [ ] I have updated the README.md (if applicable)
- [ ] I have added tests & descriptions to my models (and macros if applicable)
- [x] I have added an entry to CHANGELOG.md

PR #1086, opened 2026-05-19
fix: alias filler column in expression_is_true for strict-alias dialects

sdebruyn:
## Problem

`default__test_expression_is_true` emits an unnamed literal `1` as its placeholder column when `should_store_failures()` is false:

https://github.com/dbt-labs/dbt-utils/blob/main/macros/generic_tests/expression_is_true.sql#L7

```jinja
{% set column_list = '*' if should_store_failures() else "1" %}
```

This compiles fine standalone, but as soon as anything wraps the rendered SQL in a subquery, strict-alias dialects reject it because the inner column has no name. Concrete failure modes:

- **T-SQL (Microsoft SQL Server / Fabric Data Warehouse):** `Msg 8155, Level 16, State 2: No column name was specified for column 1 of 'sub'.`
- **Snowflake (historical):** `SQL compilation error: Missing column specification` — reported in #822, closed as Stale without a fix.
- **Some Spark configurations** that enforce strict aliasing in nested SELECTs.

Subquery wrapping is not exotic — it happens whenever a downstream macro composes generic tests, when stored-failures alternates with non-stored-failures, or when users manually wrap the test (for example to feed it into a custom monitoring model).

## Fix

One-line change: name the placeholder column.

```diff
-{% set column_list = '*' if should_store_failures() else "1" %}
+{% set column_list = '*' if should_store_failures() else "1 as filler_column" %}
```

The name `filler_column` matches the convention already used in `at_least_one`, which carries an explicit comment about this same T-SQL constraint:

https://github.com/dbt-labs/dbt-utils/blob/main/macros/generic_tests/at_least_one.sql#L36-L38

```jinja
{# In TSQL, subquery aggregate columns need aliases #}
{# thus: a filler col name, 'filler_column' #}
count({{ column_name }}) as filler_column
```

So the fix is consistent with how the project already solves the same class of problem in a sibling test.

## Impact

- **Helps:** every strict-alias dialect (T-SQL / Fabric, Snowflake under the configurations from #822, some Spark setups, anyone wrapping the test in a subquery).
- **Breaks:** nothing. `select 1 as filler_column from x` is valid SQL on PostgreSQL, Snowflake, BigQuery, Redshift, Databricks, Spark, T-SQL, and DuckDB. The test still returns the same row count to dbt, so pass/fail semantics are unchanged.

## Out of scope

This PR does **not** reintroduce the `condition` argument removed in #700. That decision stands; this is purely a portability fix for the existing macro.

## Evidence

- Stale upstream bug report: dbt-labs/dbt-utils#822
- Existing downstream workaround in `dbt-fabric`: https://github.com/sdebruyn/dbt-fabric/blob/main/src/dbt/include/fabric/macros/dbt_package_support/dbt_utils/generic_tests/expression_is_true.sql — patches exactly the same line to `"1 as col"`. With this PR, that override can be deleted.

## Test plan

- [x] Added `integration_tests/tests/generic/expression_is_true_subquery_alias.sql`, which compiles the macro and wraps its output in an outer SELECT. Without the fix this would have failed to compile on strict-alias dialects.
- [ ] Existing `data_test_expression_is_true` schema tests continue to pass on the supported adapter matrix.

### DIFF
--- a/integration_tests/tests/generic/expression_is_true_subquery_alias.sql
+++ b/integration_tests/tests/generic/expression_is_true_subquery_alias.sql
@@ -1,0 +1,33 @@
+{#
+    Validates that test_expression_is_true emits a column alias for its
+    placeholder column so that the rendered SQL is valid inside a subquery
+    on strict-alias dialects (e.g. T-SQL / Microsoft Fabric Data Warehouse,
+    some Spark configurations). Previously the macro emitted `select 1
+    from ...`, which compiles standalone but fails with errors like
+    "Missing column specification" (Snowflake, historical) or "no column
+    name was specified for column 1 of 'sub'" (T-SQL) once any caller wraps
+    the result in another SELECT. See dbt-labs/dbt-utils#822.
+
+    This test compiles the macro for a model with column_name unset and
+    wraps the rendered SQL in an outer SELECT. If the inner SELECT did not
+    alias its placeholder column, the wrapping outer SELECT would fail to
+    compile on strict-alias dialects.
+#}
+
+-- depends_on: {{ ref('data_test_expression_is_true') }}
+
+{% set inner_sql = dbt_utils.test_expression_is_true(
+    model=ref('data_test_expression_is_true'),
+    expression='col_a + col_b = 1',
+    column_name=none
+) %}
+
+with wrapped as (
+    select * from (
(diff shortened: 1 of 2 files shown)


sdebruyn:

Issue #986, opened 2025-02-19
equal_rowcount pass the test when comparing 2 different tables of the same groupby columns


light-le:
### Describe the bug

Using `equal_rowcount` between 2 different tables with the same groupby columns results in a false positive test

### Steps to reproduce
1. Create 2 different tables, with the same group by column. Let's call this `gb`. `gb` can have different values between the 2 tables (but same type). For example: 1 table could be (select 'x' as gb union all (select 'y' as gb)) and the other could be (select 'z' as gb)
2. Apply `equal_rowcount` between the 2 table using `gb` as a `group_by_columns`
3. Run the dbt data test

### Expected results

This test should fail and users should be alerted

### Actual results

This test passes without any problem

### Screenshots and log output

Result of `dbt_internal_test` if using above example
![Image](https://github.com/user-attachments/assets/e4ac0f75-4a1b-4553-a0db-236a48619b4d)

Because we have coalesce functions in the test evaluation, all null `diff_count` result as 0 failure, while it should be something else

```
select
      sum(coalesce(diff_count, 0)) as failures,
      sum(coalesce(diff_count, 0)) != 0 as should_warn,
      sum(coalesce(diff_count, 0)) != 0 as should_error
    from dbt_internal_test
```

### System information
**The contents of your `packages.yml` file:**
packages:
  - package: dbt-labs/dbt_utils
    version: 1.3.0

**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x] snowflake
- [ ] other (specify: ____________)

**The output of `dbt --version`:**
I'm using dbt cloud so I suppose it's versionless

### Additional context

I think it's line 63 of [equal_rowcount.sql](https://github.com/dbt-labs/dbt-utils/blob/main/macros/generic_tests/equal_rowcount.sql
). Instead of `abs(count_a - count_b) as diff_count` it should be `abs(coalesce(count_a, 0) - coalesce(count_b, 0))`

If this is applied, `diff_count` would no longer be null and coalesce() is no longer needed inside the sum() functions above. It could be moved to outside the sum() function like #974 which resolves #973 

### Are you interested in contributing the fix?

Yes I can contribute

PR #1090, opened 2026-05-19
fix: NULL-safe diff calculation in equal_rowcount and fewer_rows_than

sdebruyn:
## Problem

When `dbt_utils.equal_rowcount` and `dbt_utils.fewer_rows_than` are invoked with `group_by_columns`, the macros build per-group counts in CTEs `a` and `b` and then `full join` them on the group key. If a group exists in only one of the two models, the side without the group produces `NULL` for its count after the join. Subsequent arithmetic and comparisons then collapse to `NULL`, which silently hides the failure.

### `equal_rowcount`

```sql
abs(count_a - count_b) as diff_count
```

For an unmatched group (`count_a = 5`, `count_b = NULL`):
- `count_a - count_b` -> `NULL`
- `abs(NULL)` -> `NULL`
- `diff_count` for that row is `NULL`

The configured `fail_calc` is `sum(coalesce(diff_count, 0))`, so the row contributes `0` to the failure count. A genuinely unequal pair of models passes the test.

### `fewer_rows_than`

```sql
case
    when count_our_model > count_comparison_model then (count_our_model - count_comparison_model)
    when count_our_model = count_comparison_model then 1
    else 0
end as row_count_delta
```

For an unmatched group (`count_our_model = 5`, `count_comparison_model = NULL`):
- `5 > NULL` -> `NULL`
- `5 = NULL` -> `NULL`
- The `case` falls through to `else 0`, so `row_count_delta` for the row is `0`

The configured `fail_calc` is `sum(coalesce(row_count_delta, 0))`, so the row again contributes `0` and the test silently passes.

### Concrete repro

Model A (our_model):

| group_col | rows |
|-----------|------|
| x         | 3    |
| y         | 5    |

Model B (compare_model):

| group_col | rows |
|-----------|------|
| x         | 3    |
| z         | 2    |

With `group_by_columns: ['group_col']` on either test, current `main` reports zero failures because the rows for `y` and `z` have `NULL` count on one side.

This matches the user-reported bug in #986 (false positive on Snowflake). The bug is in the ANSI SQL logic, not adapter-specific, so it affects every warehouse.

## Fix

Wrap the count fields in `coalesce(..., 0)` before subtraction and comparison, so a missing group is treated as a count of zero:

- `equal_rowcount.sql`:
  ```sql
  coalesce(count_a, 0) as count_a,
  coalesce(count_b, 0) as count_b,
  abs(coalesce(count_a, 0) - coalesce(count_b, 0)) as diff_count
  ```
- `fewer_rows_than.sql`: wrap both sides of each `>` and `=` comparison and the subtraction in the `then` branch.

The fix uses only ANSI SQL and does not introduce a new dialect-specific construct.

## Integration test

Adds two new seeds (`data_test_equal_rowcount_groups_a.csv`, `data_test_equal_rowcount_groups_b.csv`) and two new test models that wire up the regression scenario above:

- `test_equal_rowcount_groups` - runs `dbt_utils.equal_rowcount` with `group_by_columns: ['group_col']`. Uses `error_if: "<1"` so the test fails on `main` (because the macro reports zero failures despite the genuine mismatch) and passes post-fix (because the failure count is now `5 + 2 = 7`).
- `test_fewer_rows_than_groups` - same data, runs `dbt_utils.fewer_rows_than` with `group_by_columns: ['group_col']`. Same `error_if: "<1"` mechanism.

## Related

- Reported upstream: #986 (false positive in `equal_rowcount` with `group_by_columns` on Snowflake)
- This was discovered while maintaining a Fabric adapter override of these two macros that already applied the `coalesce(..., 0)` fix locally. The override is no longer necessary once this lands upstream.

## Checklist

- [x] CLA on file (signed for prior contributions)
- [x] No `CHANGELOG.md` edit (per `CONTRIBUTING.md`)
- [x] Adapter-agnostic - uses only ANSI SQL `coalesce`
- [x] Integration test covers the regression

### DIFF
--- a/integration_tests/data/schema_tests/data_test_equal_rowcount_groups_a.csv
+++ b/integration_tests/data/schema_tests/data_test_equal_rowcount_groups_a.csv
@@ -1,0 +1,9 @@
+group_col,field
+x,1
+x,2
+x,3
+y,1
+y,2
+y,3
+y,4
+y,5
--- a/integration_tests/data/schema_tests/data_test_equal_rowcount_groups_b.csv
+++ b/integration_tests/data/schema_tests/data_test_equal_rowcount_groups_b.csv
@@ -1,0 +1,6 @@
+group_col,field
+x,1
+x,2
+x,3
+z,1
+z,2
--- a/integration_tests/models/generic_tests/schema.yml
+++ b/integration_tests/models/generic_tests/schema.yml
@@ -238,6 +238,18 @@ models:
           compare_model: ref('test_equal_rowcount')
           group_by_columns: ['field']
 
+  - name: test_equal_rowcount_groups
+    data_tests:
+      # Regression test for NULL handling with group_by_columns when groups are
+      # present in one model but not the other. After the full join, count_a or
+      # count_b is NULL for unmatched groups; without coalesce, diff_count
+      # collapses to NULL and the test silently passes despite a true mismatch.
+      - dbt_utils.equal_rowcount:
+          compare_model: ref('data_test_equal_rowcount_groups_b')
+          group_by_columns: ['group_col']
+          error_if: "<1" #sneaky way to ensure that the test is returning failing rows
+          warn_if: "<0"
+
   - name: test_equal_column_subset
     data_tests:
       - dbt_utils.equality:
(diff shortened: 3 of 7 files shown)


sdebruyn:

PR #1094, opened 2026-06-16
feature/star-unquote-alias-rename

### LINKED ISSUE #1093: dbt_utils.star(): Add unquote_aliases parameter to decouple identifier quoting
### Describe the feature
Add a `unquote_aliases` parameter to dbt_utils.star() that controls quoting behavior for the alias independently of the source identifier. When `quote_identifiers=true` and `unquote_aliases=true`, the macro would render `"field" as field` instead of `"field"`. This allows preserving case-sensitive source reads while producing unquoted (case-insensitive) aliases in the output. Default would be `false` to preserve existing functionality.

- Using except + manually listing columns, which is brittle and defeats the purpose of star()
- Post-processing column names with a custom macro, which adds complexity

### Additional context
This is particularly relevant for Snowflake when querying Polaris/Iceberg catalogs, where column names are case-sensitive and must be quoted in the SELECT clause, but downstream models should follow Snowflake's default uppercase-folding behavior.

One known limitation of this feature is columns whose names are reserved words cannot safely use unquoted aliases (eg. `comment`). Those columns could be handled manually via except and an explicit select. However, that defeats the purpose of a seamless dbt_utils.star() use. A complementary `rename` parameter on the star macro would make this pattern feasible.

Potentially macro use
```sql
select
    {{ dbt_utils.star(
        source('my_source', 'my_table'),
        quote_identifiers=true,
        unquote_aliases=true,
        rename={'comment': 'ticket_comment'}
    ) }}
from {{ source('my_source', 'my_table') }} 
```
Sample output
```sql
select
    "id" as id,
    "description" as description,
    "created_at" as created_at,
    "comment" as ticket_comment
....
```

### Who will this benefit?
Teams using case-sensitive data sources such as Polaris catalogs or external tables who want to read from those sources correctly without propagating quoted identifiers throughout their dbt project. Without this, you either get query errors from unquoted case-sensitive fields, or you lock your entire lineage into quoted identifiers.

Yes, I'm interested in contributing this feature!

fivetran-joemarkiewicz:
resolves #1093 

### Problem

Particularly solves cases when using the `dbt_utils.star()` macro and want to retain case sensitivity in the identifier values (already covered by `quote_identifiers=True`) but don't want to retain case sensitivity for downstream queries. Particularly relevant for Snowflake when querying Polaris/Iceberg catalogs.

### Solution

This PR introduces compatibility with the above desired functionality while not breaking the existing behavior for existing users. Users are able to leverage the new `unquote_aliases` parameter (default: False) to introduce aliases to the star macro results which are not quoted to ensure case insensitivity. 

Additionally, introduces a new parameter `rename` (default = `{}`) which allows users to specific certain fields known in the star query which could be reserved words in the database and rename them directly from the macro. This is particularly useful when wanting to retain case sensitivity in the identifier, but ensure the unquote_aliases don't fail with a reserved word failure.

See implementation examples below:
```sql
-- Model file
select
    {{ dbt_utils.star(source('zendesk','ticket_comment'), quote_identifiers=true, unquote_aliases=true, rename={'from':'from_user'}) }}
from {{ source('zendesk','ticket_comment') }}
```
Results
```sql
select
  "id" as id,
  "ticket_id" as ticket_id,
  "started_at" as started_at,
  "from" as from_user,
  etc...
from DATABASE."schema".ticket_comment
```

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1020,6 +1020,8 @@ This macro also has an optional `quote_identifiers` argument that will encase th
 - `prefix` (optional, default=`''`): will prefix the output `field_name` (`field_name as prefix_field_name`).
 - `suffix` (optional, default=`''`): will suffix the output `field_name` (`field_name as field_name_suffix`).
 - `quote_identifiers` (optional, default=`True`): will encase selected columns and aliases in double quotes (`"field_name" as "field_name"`).
+- `unquote_aliases` (optional, default=`False`): when `quote_identifiers=True`, renders aliases without quotes (`"field_name" as field_name`). Useful when reading from case-sensitive sources (e.g. Polaris/Iceberg catalogs) while producing unquoted aliases for downstream models that follow default case-folding rules.
+- `rename` (optional, default=`{}`): a dictionary mapping column names to custom alias names. The alias is always rendered unquoted. Takes precedence over `unquote_aliases` and `prefix`/`suffix` for the matched column. Useful for renaming reserved-word columns that cannot safely use an unquoted alias (e.g. `rename={"comment": "ticket_comment"}`).
 
 **Usage:**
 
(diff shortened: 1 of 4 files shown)


b-per (approved):
lgtm

PR #1095, opened 2026-06-23
Add fail_on_null option to expression_is_true generic test

### LINKED ISSUE #1026: expression_is_true tests for not-false: should test for true
`expression_is_true` tests whether `not(expression)`. It should instead test whether `not(expression) or (expression) is null`

nnqtruong:
resolves #1026

### Problem

The `expression_is_true` generic test compiles to `WHERE NOT(expression)`. In SQL three-valued logic, `NOT NULL` evaluates to `NULL`, and rows where the `WHERE` clause is `NULL` are filtered out. As a result, rows where the expression evaluates to `NULL` silently pass the test instead of failing.

This means the test checks "not false" rather than "is true". For example, a row where `col_a + col_b = 1` evaluates to `NULL` (due to NULL inputs) will pass the test even though the expression is not explicitly true.

Maintainer feedback on the issue confirmed that changing this default would be breaking for existing users, and that an opt-in approach is preferred: https://github.com/dbt-labs/dbt-utils/issues/1026#issuecomment-4645503157

### Solution

Add an optional `fail_on_null` argument (default `false`) to `expression_is_true`. When set to `true`, the test appends `OR (expression) IS NULL` to the `WHERE` clause so that NULL expression results are returned as failing rows.

The fix applies to both code paths in the macro:
- Model-level tests: `where not({{ expression }})`
- Column-level tests: `where not({{ column_name }} {{ expression }})`

Default behavior is unchanged, preserving backward compatibility.

**Example usage:**

```yaml
- dbt_utils.expression_is_true:
    expression: col_a + col_b = total
    fail_on_null: true
```

## Checklist

## Test plan

- [x] Added seed `data_test_expression_is_true_null` with a row where `col_a + col_b = 1` evaluates to NULL
- [x] Default test (no `fail_on_null`) still passes — backward compatibility preserved
- [x] `fail_on_null: true` test returns the NULL row as a failure (verified via `error_if: "<1"` pattern)
- [x] Column-level test with `fail_on_null: true` also catches NULL rows
- [ ] Full Postgres integration suite (`make test target=postgres`) — pending CI

### Before / after

**Before** (`fail_on_null` not set, NULL row present):
```
dbt test --select data_test_expression_is_true_null
→ PASS (0 failing rows returned; NULL silently passes)
```

**After** (`fail_on_null: true`, same data):
```
dbt test --select data_test_expression_is_true_null
→ Returns 1 failing row (the NULL row)
```

Made with [Cursor](https://cursor.com)


nnqtruong:
Quick CI note for reviewers: the only red check here is the **Snowflake** job, and it fails on a single unrelated test — `assert_equal_test_star_unquote_aliases_actual__expected` (part of the `star` macro, not `expression_is_true`). Postgres, Redshift, and BigQuery are all green, and all four `expression_is_true` tests pass on Snowflake too.

That failure is pre-existing on `main`: it was introduced in #1094 and the post-merge run on `main` shows the identical `FAIL 2`. Because PR CI runs against the branch merged into `main`, this PR inherits it. Root cause is a case-sensitive `rename` lookup that misses Snowflake's upper-cased identifiers.

I've opened #1096 and a fix in #1097 for that, so it's tracked separately and shouldn't block review of this PR.

PR #1097, opened 2026-06-23
fix(star): match `rename` keys to column names case-insensitively

### LINKED ISSUE #1096: star(): `rename` argument is ignored on case-folding warehouses (e.g. Snowflake)
### Describe the bug

The `rename` argument added to `star()` in #1094 matches rename keys to column names **case-sensitively**:

```jinja
{%- if col in rename %} as {{ rename[col] }}
```

On warehouses that fold unquoted identifiers to a fixed case — e.g. Snowflake stores `column_one` as `COLUMN_ONE` — `get_filtered_columns_in_relation()` returns the folded name (`COLUMN_ONE`), which never equals a lowercase key the user supplies (`{"column_one": ...}`). The rename is silently dropped and the column falls through to `unquote_aliases` / `prefix`/`suffix` / quoting as if no rename was given.

This is inconsistent with `except`, which is already normalized case-insensitively in `get_filtered_columns_in_relation`:

```jinja
```

### Steps to reproduce

On Snowflake (or any upper-case-folding adapter), against a relation with an unquoted column `column_one`:

```sql
select {{ dbt_utils.star(from=ref('my_model'), rename={"column_one": "renamed_col"}) }}
```

The compiled SQL emits `"COLUMN_ONE"` with no `as renamed_col` alias.

### Expected results

`rename` applies regardless of how the warehouse folds identifier case, consistent with `except`.

### Actual results

The rename is ignored on case-folding warehouses. This also makes the new `test_star_unquote_aliases` integration test fail on Snowflake (`FAIL 2`, "Got 2 results"). That test is currently **red on `main`** (the post-#1094 run failed on the Snowflake job).

### Root cause / proposed fix

Lowercase the `rename` keys and compare `col | lower` against them, mirroring how `except` is handled. I have a fix and a cross-adapter regression test ready (verified on Postgres) and will open a PR referencing this issue.

### System information

Affects all `star()` callers using `rename` on warehouses that fold unquoted identifiers to upper case (e.g. Snowflake). Introduced in #1094.

nnqtruong:
resolves #1096

### Problem

The `rename` argument added to `star()` in #1094 matches rename keys to column names **case-sensitively** (`{% if col in rename %}`). On warehouses that fold unquoted identifiers to a fixed case — e.g. Snowflake stores `column_one` as `COLUMN_ONE` — `get_filtered_columns_in_relation()` returns the folded name, which never equals a lowercase key the user supplies (`{"column_one": ...}`). The rename is silently dropped and the column keeps its original name.

This is inconsistent with `except`, which `get_filtered_columns_in_relation` already normalizes case-insensitively (`except | map("lower")`). It also makes the new `test_star_unquote_aliases` integration test fail on Snowflake (`FAIL 2`), which is currently red on `main`.

### Solution

Build a lowercased `rename_lookup` once and match `col | lower` against it, mirroring how `except` is handled in `get_filtered_columns_in_relation`. This makes `rename` adapter-agnostic. Lowercase-folding adapters (Postgres / Redshift / BigQuery) are unaffected; Snowflake now applies renames correctly.

I also added a regression row to `test_star_unquote_aliases` that supplies a rename key in a **different case** than the stored column name (`COLUMN_ONE` vs `column_one`). This exercises the case-insensitive path on every adapter — not just Snowflake — so the bug is caught on the always-on Postgres job too.

### Before / after (verified locally on Postgres)

With rename key `{"COLUMN_ONE": "renamed_col"}` against a `column_one` column:

**Before** (case-sensitive lookup):
```
FAIL 1 — compiled to "column_one" with no alias
```

**After** (this PR):
```
PASS — compiled to "column_one" as renamed_col
```

Full `star` test suite on Postgres: `PASS=5`.

## Checklist
- [x] This code is associated with an [issue](https://github.com/dbt-labs/dbt-utils/issues/1096) (regression introduced in #1094).
- [x] This PR includes tests
- [x] I have updated the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1021,7 +1021,7 @@ This macro also has an optional `quote_identifiers` argument that will encase th
 - `suffix` (optional, default=`''`): will suffix the output `field_name` (`field_name as field_name_suffix`).
 - `quote_identifiers` (optional, default=`True`): will encase selected columns and aliases in double quotes (`"field_name" as "field_name"`).
 - `unquote_aliases` (optional, default=`False`): when `quote_identifiers=True`, renders aliases without quotes (`"field_name" as field_name`). Useful when reading from case-sensitive sources (e.g. Polaris/Iceberg catalogs) while producing unquoted aliases for downstream models that follow default case-folding rules.
-- `rename` (optional, default=`{}`): a dictionary mapping column names to custom alias names. The alias is always rendered unquoted. Takes precedence over `unquote_aliases` and `prefix`/`suffix` for the matched column. Useful for renaming reserved-word columns that cannot safely use an unquoted alias (e.g. `rename={"comment": "ticket_comment"}`).
+- `rename` (optional, default=`{}`): a dictionary mapping column names to custom alias names. Keys are matched to column names case-insensitively (the same way `except` is), so a key works regardless of how the warehouse folds identifier case. The alias is always rendered unquoted. Takes precedence over `unquote_aliases` and `prefix`/`suffix` for the matched column. Useful for renaming reserved-word columns that cannot safely use an unquoted alias (e.g. `rename={"comment": "ticket_comment"}`).
 
 **Usage:**
 
--- a/integration_tests/models/sql/test_star_unquote_aliases.sql
+++ b/integration_tests/models/sql/test_star_unquote_aliases.sql
@@ -16,3 +16,12 @@ union all
 select
     {{ dbt.string_literal(quoted_col ~ " as renamed_col") | lower }} as expected,
     {{ dbt.string_literal(dbt_utils.star(from=ref('data_star_quote_identifiers'), quote_identifiers=True, rename={"column_one": "renamed_col"})) | trim | lower }} as actual
+
+union all
+
+-- `rename` keys should match column names case-insensitively, the same way `except` does.
+-- Warehouses that fold unquoted identifiers (e.g. Snowflake -> COLUMN_ONE) must still
+-- match a rename key supplied in a different case.
+select
+    {{ dbt.string_literal(quoted_col ~ " as renamed_col") | lower }} as expected,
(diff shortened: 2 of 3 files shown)


nnqtruong:
Marking this as a **draft** pending Snowflake validation.

What's confirmed:
- The fix is verified on **Postgres** — the added regression row fails against the old case-sensitive lookup (`FAIL 1`) and passes with this change; the full `star` suite is green (`PASS=5`).
- **Postgres, Redshift, and BigQuery** all pass in CI.

What's not resolved:
- The **Snowflake** job is still red on `test_star_unquote_aliases` (`FAIL 2`) — the same failure that exists on `main` since #1094. The fix *is* present in the tested code, so I can't yet explain why the result is unchanged on Snowflake specifically. I don't have a Snowflake environment to inspect the rendered SQL, so I'm likely missing either an identifier-folding nuance the test encodes or a package-install detail in the tox harness.

Keeping this as a draft until that's understood or a maintainer with Snowflake access can confirm. Pointers very welcome.

dbeatty10:
Thanks for opening an issue and this PR @nnqtruong !

> * The **Snowflake** job is still red on `test_star_unquote_aliases` (`FAIL 2`) — the same failure that exists on `main` since [feature/star-unquote-alias-rename #1094](https://github.com/dbt-labs/dbt-utils/pull/1094). The fix _is_ present in the tested code, so I can't yet explain why the result is unchanged on Snowflake specifically.

https://github.com/dbt-labs/dbt-utils/issues/1105 explains why the Snowflake job was red -- basically the CI wasn't working properly 😭

We already merged a fix in https://github.com/dbt-labs/dbt-utils/pull/1107 that was basically identical to the fix you proposed here. But your PR has a novel test case and a nice update to the docs, so I'm going to merge it also.

dbeatty10 (approved):
Thanks again for opening an issue and this PR @nnqtruong ! 🏆

PR #1099, opened 2026-06-23
fix(recency): add fail_on_empty param to restore pre-1.4.0 CI behaviour

### LINKED ISSUE #1098: recency test: `or most_recent is null` (in dbt_utils 1.4.0) breaks CI with `dbt --empty`
## Problem

dbt-utils 1.4.0 added `or most_recent is null` to the recency test ([#1065](https://github.com/dbt-labs/dbt-utils/pull/1065)) to catch empty tables. However, dbt-core's official `--empty` flag — designed specifically for CI use — intentionally creates models with 0 rows. This causes **all recency tests to always fail in CI** environments using `--empty`, even when the test logic is correct.

**Repro:**
1. Use `dbt build --select <model> --empty` in CI (standard slim-CI pattern)
2. Any recency test on that model now fails: `MAX(field) = NULL` → `or most_recent is null = TRUE` → FAIL

This was a silent breaking change — not flagged in the 1.4.0 release notes.

## Why we run recency tests in CI with `--empty`

The point of running recency tests in CI is **not** to check data freshness (there is no data in CI) — it is to validate that the test itself is written correctly. If someone writes a recency test referencing a non-existent column or a wrong field, the SQL will fail at runtime and CI will catch it:

```yaml
# Typo in field name — should be caught in CI
- dbt_utils.recency:
    arguments:
      datepart: hour
      field: sending_timestampz   # typo — runtime error in CI
      interval: 3
```

Before 1.4.0, this worked correctly: an empty table returned `NULL` from `MAX()`, `NULL < threshold` evaluated to `NULL` (not TRUE), so the WHERE returned 0 rows → test passed. A wrong column name would still cause a SQL error and fail CI.

After 1.4.0, the `or most_recent is null` condition makes **every** recency test fail on empty tables, regardless of whether the test is correctly written — making CI output noisy and meaningless for these tests.

## Expected behaviour

The recency test should distinguish between:
- Table has rows but data is stale → **FAIL** (intended)
- Table is intentionally empty (CI / `--empty`) → **PASS** (but wrong column names still cause a SQL error)

## Proposed fix

Add a `fail_on_empty` parameter (default `true` for backwards compat):

```yaml
- dbt_utils.recency:
    arguments:
      datepart: hour
      field: sending_timestamp
      interval: 3
      fail_on_empty: false  # safe for CI with --empty
```

Or guard with a row count check inside the macro:

```sql
where row_count > 0
  and (most_recent < {{ threshold }} or most_recent is null)
```

## References

- dbt-core `--empty` flag: https://docs.getdbt.com/reference/global-configs/empty
- Breaking change introduced in: https://github.com/dbt-labs/dbt-utils/pull/1065

onurtashan:
## Summary

Adds a `fail_on_empty` parameter (default `False`) to the `recency` test, restoring the pre-1.4.0 behaviour for CI environments that use `dbt --empty`.

Fixes #1098

## Problem

1.4.0 introduced `or most_recent is null` to catch genuinely empty production tables. However, `dbt --empty` — dbt-core's official flag for CI slim builds — **intentionally** creates 0-row tables for schema and SQL validation. This causes every recency test to fail in CI, regardless of whether the test logic is correct.

The primary reason to run recency tests in CI with `--empty` is to **validate the test itself** (wrong column name → SQL error → CI catches it). 1.4.0 broke this workflow silently, without a breaking-change notice.

## Solution

```yaml
# CI: restore original pass-on-empty behaviour (default)
- dbt_utils.recency:
    arguments:
      datepart: hour
      field: sending_timestamp
      interval: 3

# Production: opt-in to catching empty tables
- dbt_utils.recency:
    arguments:
      datepart: hour
      field: sending_timestamp
      interval: 3
      fail_on_empty: true
```

## Behaviour

| Scenario | `fail_on_empty=False` (default) | `fail_on_empty=True` |
|---|---|---|
| Table has rows, data is fresh | ✅ PASS | ✅ PASS |
| Table has rows, data is stale | ❌ FAIL | ❌ FAIL |
| Table is empty (e.g. CI `--empty`) | ✅ PASS | ❌ FAIL |
| Wrong column name | ❌ SQL ERROR | ❌ SQL ERROR |

## Backwards compatibility

Default is `False` — restores pre-1.4.0 behaviour. Projects that relied on the 1.4.0 empty-table detection can opt in with `fail_on_empty: true`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* fix(recency): add `fail_on_empty` parameter (default `False`) to restore pre-1.4.0 behaviour for CI environments using `dbt --empty` by @onurtashan in https://github.com/dbt-labs/dbt-utils/pull/TBD
+
 **Full Changelog**: https://github.com/dbt-labs/dbt-utils/compare/1.4.0...main
 
 # dbt utils v1.3.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-* fix(recency): add `fail_on_empty` parameter (default `False`) to restore pre-1.4.0 behaviour for CI environments using `dbt --empty` by @onurtashan in https://github.com/dbt-labs/dbt-utils/pull/TBD
+* fix(recency): add `fail_on_empty` parameter (default `False`) to restore pre-1.4.0 behaviour for CI environments using `dbt --empty` by @onurtashan in https://github.com/dbt-labs/dbt-utils/pull/1099
 
 **Full Changelog**: https://github.com/dbt-labs/dbt-utils/compare/1.4.0...main
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,3 @@
-# Unreleased
-
-* fix(recency): add `fail_on_empty` parameter (default `False`) to restore pre-1.4.0 behaviour for CI environments using `dbt --empty` by @onurtashan in https://github.com/dbt-labs/dbt-utils/pull/1099
-
-**Full Changelog**: https://github.com/dbt-labs/dbt-utils/compare/1.4.0...main
-
 # dbt utils v1.3.3
 
 * fix: Render relations to avoid including time filters in source_column_name by @harshvardhan-j in https://github.com/dbt-labs/dbt-utils/pull/1061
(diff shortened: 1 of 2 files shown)


dbeatty10:
Thanks for reporting this issue and opening this PR @onurtashan 🤩 

I incorporated your ideas from this PR into #1102. I think it will solve the problem you are experiencing without your (or other users) needing to make any changes to your projects (like `fail_on_empty: true`).

So I'm going to close this in favor of #1102.

PR #1107, opened 2026-06-25
fix/star-rename-casing

### LINKED ISSUE #1096: star(): `rename` argument is ignored on case-folding warehouses (e.g. Snowflake)

### LINKED ISSUE #1103: CI failing after #1094
### Describe the bug

After https://github.com/dbt-labs/dbt-utils/pull/1094 was merged in commit https://github.com/dbt-labs/dbt-utils/commit/8124d54bfc3af8d1aa2ea569fa4332d9d6db4be3, CI failed with this output:

```
FAIL 2 assert_equal_test_star_unquote_aliases_actual__expected ...... [FAIL 2 in 0.21s]
```

### Cause

Here is the [compiled model for Snowflake](https://app.circleci.com/pipelines/github/dbt-labs/dbt-utils/1974/workflows/5df067b5-6e90-41c2-b8dc-eff411b5241e/jobs/6541/artifacts):

```sql
select
    '"column_one" as column_one' as expected,
    '"column_one" as column_one' as actual

union all

select
    '"column_one" as renamed_col' as expected,
    '"column_one" as column_one' as actual

union all

select
    '"column_one" as renamed_col' as expected,
    '"column_one"' as actual
```

Whereas the [compiled model for BigQuery](https://app.circleci.com/pipelines/github/dbt-labs/dbt-utils/1974/workflows/5df067b5-6e90-41c2-b8dc-eff411b5241e/jobs/6539/artifacts) is this:

```sql
select
    '''`column_one` as column_one''' as expected,
    '''`column_one` as column_one''' as actual

union all

select
    '''`column_one` as renamed_col''' as expected,
    '''`column_one` as renamed_col''' as actual

union all

select
    '''`column_one` as renamed_col''' as expected,
    '''`column_one` as renamed_col''' as actual
```

fivetran-joemarkiewicz:
resolves #1096
resolves #1103

### Problem

PR #1094 introduced the `rename` parameter to `dbt_utils.star()`. This parameter does not work as expected for Snowflake due to the casing of default Snowflake fields. The previous implementation required the name of the field in `rename` to match the casing of the field in the table. As such, if the field was `"COLUMN_ONE"` then the rename parameter of `rename={'column_one':'renamed_col'}` wouldn't match. This is why the test was failing for Snowflake.

```sql
-- Using main branch
select
    {{ dbt_utils.star(source('zendesk','ticket_comment'), quote_identifiers=true, unquote_aliases=true, rename={'body':'content'}) }}
from {{ source('zendesk','ticket_comment') }}
```
results in 
```sql
select
  "CREATED" as CREATED,
  "PUBLIC" as PUBLIC,
  "BODY" as BODY, -- should be `"BODY" as content`
  "ID" as ID,
  "TICKET_ID" as TICKET_ID,
  "USER_ID" as USER_ID
from DATABASE.zendesk_schema.ticket_comment
```

### Solution

This is addressed by ensuring the dictionary lookup of the key is lowered when in both the list and the check. This way it will be able to understand that `"COLUMN_ONE"` and `"column_one"` are the same and should be renamed accordingly.

A con of this approach is this could result in duplicative matches if there truly are two or more columns with the same name, but with different casing (eg. `"Column_One"`, `"COLUMN_ONE"`, `"column_one"`). If this is a hard blocker, the existing implementation is sufficient. However, the test will need to be modified for Snowflake to ensure the key in rename is all caps and the documentation will need to be updated. My reasoning for taking the approach in this PR is this is similar to how `except` works currently. Since there's precedent it felt appropriate to take this approach.

```sql
-- Using this PR branch
select
    {{ dbt_utils.star(source('zendesk','ticket_comment'), quote_identifiers=true, unquote_aliases=true, rename={'body':'content'}) }}
from {{ source('zendesk','ticket_comment') }}
```
results in

```sql
select
  "CREATED" as CREATED,
  "PUBLIC" as PUBLIC,
  "BODY" as content, -- is now correct
  "ID" as ID,
  "TICKET_ID" as TICKET_ID,
  "USER_ID" as USER_ID
from DATABASE.zendesk_schema.ticket_comment
```

## Checklist

### DIFF
--- a/integration_tests/models/sql/schema.yml
+++ b/integration_tests/models/sql/schema.yml
@@ -194,8 +194,6 @@ models:
       - assert_equal:
           actual: actual
           expected: expected
-          config:
-            enabled: "{{ target.type != 'snowflake' }}"
 
   - name: test_star_no_columns
     columns: 
--- a/macros/sql/star.sql
+++ b/macros/sql/star.sql
@@ -13,6 +13,11 @@
 
     {% set cols = dbt_utils.get_filtered_columns_in_relation(from, except) %}
 
+    {%- set rename_lower = {} -%}
+    {%- for key, val in rename.items() -%}
+        {%- do rename_lower.update({key.lower(): val}) -%}
+    {%- endfor -%}
+
     {%- if cols|length <= 0 -%}
         {% if flags.WHICH == 'compile' %}
             {% set response %}
@@ -30,13 +35,13 @@ dbt compile, and exists to keep SQLFluff happy. */
             {%- if relation_alias %}{{ relation_alias }}.{% else %}{%- endif -%}
                 {%- if quote_identifiers -%}
                     {{ adapter.quote(col)|trim }}
-                    {%- if col in rename %} as {{ rename[col] }}
+                    {%- if col.lower() in rename_lower %} as {{ rename_lower[col.lower()] }}
                     {%- elif unquote_aliases %} as {{ (prefix ~ col ~ suffix)|trim }}
                     {%- elif prefix!='' or suffix!='' %} as {{ adapter.quote(prefix ~ col ~ suffix)|trim }}
                     {%- endif -%}
                 {%- else -%}
                     {{ col|trim }}
-                    {%- if col in rename %} as {{ rename[col] }}
+                    {%- if col.lower() in rename_lower %} as {{ rename_lower[col.lower()] }}
                     {%- elif prefix!='' or suffix!='' %} as {{ (prefix ~ col ~ suffix)|trim }}
                     {%- endif -%}
                 {%- endif -%}
@@ -44,4 +49,3 @@ dbt compile, and exists to keep SQLFluff happy. */
         {%- endfor -%}
     {% endif %}
 {%- endmacro %}
-


dbeatty10 (approved):
Case-insensitive comparisons make me uncomfortable for exactly the reasons you mentioned.

But I see what you are saying about the precedent set with the [behavior of the `except` parameter](https://github.com/dbt-labs/dbt-utils/blob/f3b225dfde76800e932db35e8cc0420fe3eedd47/macros/sql/get_filtered_columns_in_relation.sql#L16-L21). (Also, it's not the only place dbt does case-insensitive comparisons 😅 ).

So we can continue with that precedent here. If it turns out to be a problem, we can always change the behavior in a new minor version and/or added YAP ("yet another parameter") to control case-sensitive vs. non-case-sensitive comparisons (not quite a [collation](https://stackoverflow.com/questions/4538732/what-does-collation-mean) setting, but in the same neighborhood).

PR #1115, opened 2026-07-25
Fix: Quote database identifier in get_tables_by_pattern_sql for CamelCase support

### LINKED ISSUE #1060: Non-quoted database refrence causing error when CamelCase Database name is in used for Postgres
https://github.com/dbt-labs/dbt-utils/blob/09382f23693f2e8bc1dca275cc39638129e4c683/macros/sql/get_tables_by_pattern_sql.sql#L12

{{ database }} should be replaced with {{ adapter.quote(database) }}

Ahmed-Gomaa1:
resolves #1060

### Problem

The `default__get_tables_by_pattern_sql` macro does not quote the database identifier. This causes failures when using CamelCase or case-sensitive database names (e.g. in Postgres), since unquoted identifiers are automatically lowercased.

Example of generated SQL before the fix:

```sql
from test.information_schema.tables
```

This breaks when the actual database name is case-sensitive (e.g. `"TestDB"`).

### Solution

Updated the default macro to use adapter-aware quoting:

```sql
from {{ adapter.quote(database) }}.information_schema.tables
```

This ensures compatibility with case-sensitive database names across adapters.

Note: only the `default__get_tables_by_pattern_sql` macro was modified. Adapter-specific implementations (e.g. Redshift, BigQuery) were left unchanged to avoid unintended side effects.

## Checklist
- [ ] This code is associated with an [issue](https://github.com/dbt-labs/dbt-utils/issues) which has been triaged and [accepted for development](https://docs.getdbt.com/docs/contributing/oss-expectations#pull-requests)
- [ ] I have updated the README.md (not applicable)


Ahmed-Gomaa1:
It looks like the CI failure is unrelated to this PR's changes. The pipeline is failing to install a dependency:

dbt-core @ git+https://github.com/dbt-labs/dbt-core.git#subdirectory=core

This fails because the referenced subdirectory no longer contains a valid Python project (no setup.py or pyproject.toml), which suggests the dbt-core repo structure has changed upstream.

This PR only modifies SQL macro logic in default__get_tables_by_pattern_sql, so I don't believe it's related. Happy to investigate further or adjust if needed — just let me know.

Ahmed-Gomaa1:
Hi 👋

Following up on this PR. 

@b-per would appreciate your feedback when you have time.

This change is scoped to quoting the database identifier to fix issues with CamelCase names, with no intended behavior changes.

Happy to update anything if needed. 
Thanks

Ahmed-Gomaa1:
Hi @dbeatty10, quick follow-up on this PR.

It fixes case-sensitivity issues by quoting the database in "default__get_tables_by_pattern_sql" using "adapter.quote()".

Would appreciate your feedback when you have time.

